Issue #34, opened 2017-04-26
Add bundles, resources, structured_resources to swift_library.


allevato:
`swift_library` should be able to propagate resources and bundles to applications/extensions/frameworks just like `objc_library` can.

This was almost working but had some weird issues with certain dependency graphs where the aspect wasn't drilling down into other deps for unknown reasons. Gotta investigate.

allevato:
`resources` and `structured_resources` are now supported. (Things like `datamodels`, `storyboards`, `xibs`, etc. should just be placed in `resources` now.)

`bundles` will be coming soon.

sergiocampama:
Obsolete. The current plan is to use `apple_bundle_import`, `apple_resource_bundle` and `apple_resource_group`, plus direct file references, through the `data` attribute.

PR #40, opened 2017-05-02
Add "resources", "structured_resources" attrs to swift_library.

allevato:
This *only* works with the Skylark-based bundling rules; the attributes are ignored if a native bundlng rule depends on them.

The "resources" attribute takes the place of "asset_catalogs", "datamodels", "resources", "storyboards", "strings", and "xibs" that are on similar rules like "objc_library". All the specific resource types can be included in "resources" instead and they are processed according to their names.

"structured_resources" has the same meaning as before.

"bundles" are not yet supported; they will be in the near future.

Partially addresses #34.

RELNOTES: None.
PiperOrigin-RevId: 154857955

### DIFF
--- a/apple/bundling/apple_bundling_aspect.bzl
+++ b/apple/bundling/apple_bundling_aspect.bzl
@@ -190,9 +190,7 @@ def _transitive_apple_resource(target, ctx):
   """
   resource_sets = []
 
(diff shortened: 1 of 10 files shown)


keith:
Awesome thanks!!

PR #45, opened 2017-05-08
Add new Skylark-based unit and UI testing rules.

allevato:
PiperOrigin-RevId: 155093995


StevenEWright:
Horray!!!

rahul-malik:
@allevato - This is great. Is `apple_unit_test` going to replace existing `ios_test` rules?

allevato:
Yes, like the native application/extension rules, the native testing rules (`ios_test`) will be deprecated and these rules will replace them (and will be extended to more platforms as well).

We hope to get some documentation/examples up on these rules soon.

allevato:
(And credit goes to @sergiocampama; I just pushed the commits. 😄 )

sergiocampama:
apple_unit_test and apple_ui_test are not meant to be used directly, but through the platform specific ios_unit_test and ios_ui_test, which have the responsibility of building the test bundle and configuring the test binary.

And yes, now that ios_unit_test and ios_ui_test are available, we will start the deprecation process for ios_test.

rahul-malik:
@allevato @sergiocampama - Thank you and I'm looking forward to the docs! One last question - I'm assuming this means we'll be able to utilize `bazel test` for running unit tests?

sergiocampama:
For some specific use cases. The test runner made available (and the default for now) is very limited in what it can do, which is only run tests that do not depend on a test host (logic based tests).

We are currently working on an open source test runner, but I can't yet give an estimate on when it might be released. 

Keep in mind that `ios_unit_test` and `ios_ui_test` do work with Tulsi for both simulator and device tests.

PR #68, opened 2017-05-16
Add trap to clean up simulator in ios_sample_test_runner

shamanskyh:
Hi Bazel team,

I’m working on changing over our `ios_test` rules to the new `ios_unit_test` rules, and I’ve run into a few issues. The first is that I’m unable to call `bazel test` on the rule unless I disable the sandbox (using the `--strategy=TestRunner=standalone` flag). While debugging this, I found that the simulator that the default testrunner script creates isn’t cleaned up if anything fails along the way. To address this, could we add a trap that deletes the created simulator?

Also, should I open an issue around the sandboxing issue? The error I get is:
```
SandboxViolation: simctl(41704) deny file-write-create /Users/hshamansky/Library/Developer/CoreSimulator/Devices/26D97441-A139-45F7-A75F-2FA3A0C564DD/data/Library/Logs
Violation:       deny file-write-create /Users/hshamansky/Library/Developer/CoreSimulator/Devices/26D97441-A139-45F7-A75F-2FA3A0C564DD/data/Library/Logs 
Process:         simctl [41704]
Path:            /Applications/Xcode.app/Contents/Developer/usr/bin/simctl
Load Address:    0x104ed4000
Identifier:      simctl
Version:         ??? (???)
Code Type:       x86_64 (Native)
Parent Process:  bash [41689]
Responsible:     /Applications/Utilities/Terminal.app/Contents/MacOS/Terminal [33314]
User ID:         501
```

### DIFF
--- a/apple/testing/default_runner/ios_sample_test_runner.template.sh
+++ b/apple/testing/default_runner/ios_sample_test_runner.template.sh
@@ -30,13 +30,6 @@ basename_without_extension() {
 
 # Unpack the output IPA into a tmp folder
 TEST_TMP_DIR="$(mktemp -d "${TMPDIR:-/tmp}/tests.XXXXXX")"
-trap 'rm -rf "${TEST_TMP_DIR}"' ERR EXIT
-
-TEST_BUNDLE_PATH="%(test_bundle_path)s"
-TEST_BUNDLE_NAME=$(basename_without_extension "$TEST_BUNDLE_PATH")
-
-unzip -qq "$TEST_BUNDLE_PATH" 'Payload/*' -d "$TEST_TMP_DIR"
-TEST_BUNDLE="$TEST_TMP_DIR/Payload/$TEST_BUNDLE_NAME.xctest"
 
 # Create a simulator with a random name and the iOS SDK provided by the
 # Xcode currently selected with xcode-select.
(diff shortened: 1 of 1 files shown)


apple/testing/default_runner/ios_sample_test_runner.template.sh
```diff
@@ -45,6 +45,12 @@ SDK_VERSION="$(xcrun --sdk iphonesimulator --show-sdk-version)"
 
 NEW_SIM_ID=$(xcrun simctl create "$RANDOM_NAME" "iPhone 6" "$SDK_VERSION")
 
+# Clean our simulator up even if we fail along the way
+function cleanup {
```
sergiocampama: so trap doesn't stack, which means that the one in line 33 would be discarded... would you mind moving this to the top and merge it with the `rm -rf` part? this would also mean creating the simulator at the top too

sergiocampama:
@allevato for any insights on the sandbox issue...

apple/testing/default_runner/ios_sample_test_runner.template.sh
shamanskyh: Ah, totally missed the trap on line 33. Good catch - I'll submit a revision.

apple/testing/default_runner/ios_sample_test_runner.template.sh
```diff
@@ -57,8 +63,6 @@ XCTEST_PATH="$XCODE_PATH/Platforms/iPhoneSimulator.platform/Developer/Library/Xc
 xcrun simctl spawn "$NEW_SIM_ID" "$XCTEST_PATH" "$TEST_BUNDLE"
 EXIT_CODE=$?
 
-xcrun simctl delete $"$NEW_SIM_ID"
-
 # Bazel detects the exit code from this script as the status of whether the
 # tests succeeded or failed. Any exit code other than 0 means tests failed.
 exit $EXIT_CODE
```
sergiocampama: I don;t think we need the proxy any more, we can just call `xcrun simctl spawn ...` without storing the exit code... that was used so we had a chance to clean up the sim... but since that is now in trap, we can remove this...

apple/testing/default_runner/ios_sample_test_runner.template.sh
```diff
@@ -30,13 +30,6 @@ basename_without_extension() {
 
 # Unpack the output IPA into a tmp folder
```
sergiocampama: perhaps reword this comment to "Create tmp folder to contain the extracted .xctest bundle", and then add a comment where we extract the test bundle in line 52

shamanskyh:
Thanks @sergiocampama!

PR #84, opened 2017-06-20
Fix iOS simulator unit tests

jgavris:
I started working on this before @dflems fixed the sim runtime, but there were a few more issues in the runner script. The test `//test:ios_unit_test_test.simulator` now passes for me against Xcode 9.0 beta 1. 

Unfortunately, when I run this against Xcode 8.3.3, I get the following error. Does anyone maybe know more about the toolchain to know what's wrong here? 

>2017-06-20 21:46:03.261 simctl[99457:8822718] CoreSimulatorService connection became invalid.  Simulator services will no longer be available.
Unable to locate device set: Error Domain=NSPOSIXErrorDomain Code=53 "Software caused connection abort" UserInfo={NSLocalizedFailureReason=Failed to subscribe to notifications from CoreSimulatorService., NSLocalizedDescription=Failed to initialize simulator device set., NSUnderlyingError=0x7fef3660f0b0 {Error Domain=NSPOSIXErrorDomain Code=53 "Software caused connection abort" UserInfo={NSLocalizedDescription=Error returned in reply to notification request: Connection invalid}}}
Use --strategy=TestRunner=standalone to disable sandboxing for the failing actions.

`bazel version` == `0.5.1-homebrew`

### DIFF
--- a/apple/testing/default_runner/ios_sample_test_runner.template.sh
+++ b/apple/testing/default_runner/ios_sample_test_runner.template.sh
@@ -40,7 +40,7 @@ NEW_SIM_ID=$(xcrun simctl create "$RANDOM_NAME" "iPhone 6" "$SIM_VERSION")
 # Clean up our simulator and temp directory if we fail along the way
 function cleanup {
   rm -rf "${TEST_TMP_DIR}"
-  xcrun simctl delete $"$NEW_SIM_ID"
+  xcrun simctl delete "$RANDOM_NAME"
 }
 trap cleanup ERR EXIT
 
--- a/apple/testing/default_runner/ios_sample_test_runner.template.sh
+++ b/apple/testing/default_runner/ios_sample_test_runner.template.sh
@@ -33,20 +33,33 @@ TEST_TMP_DIR="$(mktemp -d "${TMPDIR:-/tmp}/tests.XXXXXX")"
 
 # Create a simulator with a random name and the latest iOS SDK
 RANDOM_NAME="$(cat /dev/urandom | tr -dc 'a-zA-Z0-9' | fold -w 8 | head -n 1)"
-SIM_VERSION="$(xcrun simctl list runtimes | grep "SimRuntime\.iOS" | sed 's/^iOS [0-9\\.]* (\([0-9\.]*\).*/\1/' | tail -n 1)"
+SDK_VERSION="$(xcrun --sdk iphonesimulator --show-sdk-version | tr "." "-")"
+SIMULATOR_RUNTIME="com.apple.CoreSimulator.SimRuntime.iOS-$SDK_VERSION"
+NEW_SIM_ID=$(xcrun simctl create "$RANDOM_NAME" "iPhone 6" "$SIMULATOR_RUNTIME")
 
-NEW_SIM_ID=$(xcrun simctl create "$RANDOM_NAME" "iPhone 6" "$SIM_VERSION")
+# Shut down and clean up our simulator and temp directory even if we fail along the way.
+did_cleanup=0
 
-# Clean up our simulator and temp directory if we fail along the way
 function cleanup {
-  rm -rf "${TEST_TMP_DIR}"
-  xcrun simctl delete "$RANDOM_NAME"
+  if [[ "$did_cleanup" == "0" ]]; then
+    did_cleanup=1
+    rm -rf "${TEST_TMP_DIR}"
+    xcrun simctl shutdown "$RANDOM_NAME"
+    xcrun simctl delete "$RANDOM_NAME"
+  fi
 }
+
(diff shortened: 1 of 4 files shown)


sergiocampama:
I think the problem you're seeing is that you have multiple versions of xcode installed, and it gets dizzy because there are multiple version of the simulator daemon running. Do you only have 8.3.3 and 9.0b?

jgavris:
Yeah, I have both. I have only one instance of the simulator daemon running, it's the Xcode 8.3.3 version. And I've set the path with xcode select to that as well. Does `master` pass that test for you? Does this branch?

dflems:
@jgavris I've seen this before. Generally when you use simulators in a new version of Xcode, it will cleanly `kill the CoreSimulatorService` daemon that was left running by the other version. It appears that Xcode 9's service does not get shut down like others have in the past. I wrote a little script that tears everything down to a clean state.

```sh
#!/bin/sh

function kill_app() {
  killall -9 "$1" 2>/dev/null && echo "Killed $1.app"
}

function kill_launchd_services() {
  while launchctl list | grep -q "$1"; do
    for name in `launchctl list | grep "$1" | awk '{print $3}'`; do
      echo "Killing $name"
      launchctl remove "$name"
      launchctl stop "$name"
    done
    sleep 1
  done
}

kill_app Xcode
kill_app Simulator
kill_launchd_services launchd_sim
kill_launchd_services CoreSimulatorService
```

jgavris:
Thanks @dflems! I just gave that a try at the top of the runner script here. It still fails with the same error message. It seems like they fixed a lot of things Xcode 9 😁

jgavris:
👋  can someone take a look here? I've changed the unit tests to actually invoke `bazel test` and I believe that is exposing a failure to run the tests (on Xcode 8). This passes for me on the Xcode 9 beta 2.

sergiocampama:
I'm sorry for the lack of review, I just returned from vacation and haven't had the time to review, will try to take a look this week

jgavris:
Checking in here again, could you give this a try?

sergiocampama:
I believe that error has to do with the sandboxing bazel does when running the actions. I can get Xcode 8.3.3 to run the tests when I use `--test_strategy=standalone --strategy=TestRunner=standalone`. I'm very happy to know that Xcode 9 fixes this :)

apple/testing/default_runner/ios_sample_test_runner.template.sh
```diff
@@ -33,21 +33,34 @@ TEST_TMP_DIR="$(mktemp -d "${TMPDIR:-/tmp}/tests.XXXXXX")"
...
 # Create a simulator with a random name and the latest iOS SDK
 RANDOM_NAME="$(cat /dev/urandom | tr -dc 'a-zA-Z0-9' | fold -w 8 | head -n 1)"
-SIM_VERSION="$(xcrun simctl list runtimes | grep "SimRuntime\.iOS" | sed 's/^iOS [0-9\\.]* (\([0-9\.]*\).*/\1/' | tail -n 1)"
+SDK_VERSION="$(xcrun --sdk iphonesimulator --show-sdk-version | tr "." "-")"
+SIMULATOR_RUNTIME="com.apple.CoreSimulator.SimRuntime.iOS-$SDK_VERSION"
+NEW_SIM_ID=$(xcrun simctl create "$RANDOM_NAME" "iPhone 6" "$SIMULATOR_RUNTIME")
 
-NEW_SIM_ID=$(xcrun simctl create "$RANDOM_NAME" "iPhone 6" "$SIM_VERSION")
+# Shut down and clean up our simulator and temp directory even if we fail along the way.
+did_cleanup=0
```
sergiocampama: why is this needed? wouldn't the trap only get called once?

jgavris:
So you cannot get 8.3.3 to run with sandboxing enabled? That's my experience as well.

apple/testing/default_runner/ios_sample_test_runner.template.sh
```diff
@@ -33,21 +33,34 @@ TEST_TMP_DIR="$(mktemp -d "${TMPDIR:-/tmp}/tests.XXXXXX")"
...
 }
+
 trap cleanup ERR EXIT
 
 # Wait a bit so that the newly created simulator can pass from the Creating
 # state to the Shutdown state.
 sleep 2
 
+# Ensure simulator is booted.
+xcrun simctl boot "$NEW_SIM_ID"
```
sergiocampama: why boot up the simulator? when running logic only tests, (by using the xctest agent directly), the sim doesn't have to be booted, which is actually faster

sergiocampama:
yeah, you need to disable sandboxing for running tests with Xcode 8.3.3

sergiocampama:
Closing as we are using a new ios test runner

PR #85, opened 2017-06-22
Remove warning, rules are now working with latest Bazel release

steren:


allevato:
Before removing this, I want to tag a release that we document as being compatible with Bazel 0.5.0—I just haven't had the time to do so yet.

steren:
Of course, that makes sense.
0.5.1 is the Bazel release you should be using, as 0.5.0 had troubles on macOS.

allevato:
We've now tagged the 0.1.0 release that is compatible with Bazel 0.6.1, so I'll merge this.

PR #96, opened 2017-07-18
Remove bazel version warning.

tadeegan:
This seems outdated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,12 +1,5 @@
 # Apple Rules for [Bazel](https://bazel.build)
 
-> :warning: **NOTE**: At the time of this writing, the most recent Bazel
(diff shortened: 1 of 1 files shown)


allevato:
Thanks for the pull request!

This section needs to be updated, but not completely removed, because the rules in this repo at HEAD currently don't work with the Bazel 0.5.2 release (AFAIK). It's still best to build Bazel from HEAD until things stabilize more between both the Bazel code base and these rules (because we're constantly moving Apple-related functionality out of Bazel and into this repo, so there will be some gaps between Bazel releases where the repo may not work).

Once things settle a bit, we would like to be more aggressive with tagging releases that we know are stable with respect to a particular Blaze release.

allevato:
Duplicates #85.

PR #100, opened 2017-07-26
Rename set to depset

vladmos:


allevato:
Thanks for the cleanup!

PR #107, opened 2017-08-29
Enable tests for CI

allevato:


allevato:
Note: I'll be iterating over this PR to figure out why `rules_apple` has been failing on CI in our initial attempts to set it up. Please do not merge until I say it's ready.

damienmg:
jenkins test this please

allevato:
@damienmg In case you were curious, I've tracked down the problem. Since Bazel CI tries to build "//..." by default, this was including `objc_library` targets that used different platforms other than the default (which is iOS). So trying to _directly_ build a library containing watchOS or macOS code fails—they only build correctly when an `apple_binary` uses a split transition to propagate the correct platform type down to those libraries.

So the fix is to explicitly targets that we want to build and test, which I'm working on now.

PR #114, opened 2017-10-13
Return to using security tool for extracting provisioning plist from profile.

rahul-malik:
In High Sierra there are encoding issues when reading profiles using the
openssl tool but the security tool works correctly. Builds that
previously worked in Sierra are failing in High Sierra due to this
depending on the decrypted value of your profile.

This, strangely, only happens for some of our provisioning profiles. The security tool version works for all of them.

### DIFF
--- a/apple/bundling/plist_support.bzl
+++ b/apple/bundling/plist_support.bzl
@@ -33,14 +33,7 @@ def _extract_provisioning_plist_command(ctx, provisioning_profile):
     # plist without a signature.
     return "cat " + bash_quote(provisioning_profile.path)
   else:
-    # The `security cms -D -i` command previously used here spits out a message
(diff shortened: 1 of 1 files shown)


bazel-io:
Can one of the admins verify this patch?

apple/bundling/plist_support.bzl
```diff
@@ -40,7 +40,7 @@ def _extract_provisioning_plist_command(ctx, provisioning_profile):
     # be somewhat preferable to one that implies failure. Regardless, we capture
```
sergiocampama: can you update this comment? it's not true after this change

rahul-malik: updated

rahul-malik:
@sergiocampama - Removed the comment, let me know if you need anything else to merge

PR #118, opened 2017-11-08
Export

sergiocampama:


apple/bundling/plist_support.bzl
```diff
@@ -33,7 +33,14 @@ def _extract_provisioning_plist_command(ctx, provisioning_profile):
     # plist without a signature.
     return "cat " + bash_quote(provisioning_profile.path)
   else:
-    extract_plist_cmd = ("security cms -D -i " +
```
rahul-malik: @sergiocampama - This seems to undo the PR that merged yesterday for High Sierra.

sergiocampama: Yup, we didn't notice at first (or tools may be lacking in this aspect) but I cherrypicked your change back at the top of master

PR #135, opened 2018-02-16
Add a travis config for testing.

thomasvl:
Also adds examples/BUILD to provide some targets for CI to
use since bazel build|test will try to build some things
that don't really build on their own.


thomasvl:
Oh, if we add tags to the `device`, `simulator`, etc. variants of the tests we create and then use bazel's  test tag filter to spin those out into different configs.  While it would shorten each run, I the bazelbuild org is likely limited to a max of two builds on macos at a time (across the whole org), so we're competing for resources, and it wouldn't help throughput as much.  So multiple configs for organization is more worth while that trying to hope for more parallel building.

examples/BUILD
```diff
@@ -0,0 +1,40 @@
...
+        "//examples/macos/HelloWorld",
+        "//examples/macos/HelloWorldSwift",
+        "//examples/multi_platform/Buttons:Buttons",
+        "//examples/multi_platform/Buttons:ButtonsMac",
+        "//examples/multi_platform/Buttons:ButtonsTV",
+        "//examples/tvos/HelloWorld",
+        "//examples/watchos/HelloWorld",
+    ],
+)
+
```
sergiocampama: remove extra line, it will trip up buildozer later

thomasvl: Done

thomasvl:
Went ahead and added the info to accept `//test/...` as failing.

PR #152, opened 2018-03-13
Codesign bug was fixed.

pwnartist:
When compiling iOS application for iphoneos with flags
`--ios_multi_cpus=armv7,arm64`, all of compiling have been succeeded.
In the phase of codesign, bazel cannot find the credential file,
`.mobileprovision`.
This seems to be caused by no copy of the credential in the sandbox
building the source codes.

In details, the sandbox would copy the owned credential as
`embedded.mobileprovision` into sandbox directory.
Therefore, this code used `find . -name 'embedded.mobileprovision'` as
a provisioning profile instead of a specified path in `BUILD` rules,
which cannot be found in sandbox.

### DIFF
--- a/apple/bundling/codesigning_support.bzl
+++ b/apple/bundling/codesigning_support.bzl
@@ -50,7 +50,7 @@ def _extract_provisioning_plist_command(ctx, provisioning_profile):
     # The whole output for that fallback command group is then rerouted to
     # STDERR which is only printed if the command actually failed (security and
     # openssl print information into stderr even if the command succeeded).
-    profile_path = shell.quote(provisioning_profile)
+    profile_path = provisioning_profile
     extract_plist_cmd = (
         "(security cms -D -i %s || " % profile_path +
         "openssl smime -inform der -verify -noverify -in %s)" % profile_path)
--- a/apple/bundling/codesigning_support.bzl
+++ b/apple/bundling/codesigning_support.bzl
@@ -39,7 +39,7 @@ def _extract_provisioning_plist_command(ctx, provisioning_profile):
   if mock_support.is_provisioning_mocked(ctx):
     # If provisioning is mocked, treat the provisioning profile as a plain XML
     # plist without a signature.
-    return "cat " + shell.quote(provisioning_profile.path)
+    return "cat " + shell.quote(provisioning_profile)
   else:
     # NOTE: Until the bundling rules are updated to merge entitlements support
     # and signing, this extraction command should be kept in sync with what
@@ -50,7 +50,7 @@ def _extract_provisioning_plist_command(ctx, provisioning_profile):
     # The whole output for that fallback command group is then rerouted to
     # STDERR which is only printed if the command actually failed (security and
     # openssl print information into stderr even if the command succeeded).
-    profile_path = shell.quote(provisioning_profile.path)
+    profile_path = shell.quote(provisioning_profile)
     extract_plist_cmd = (
         "(security cms -D -i %s || " % profile_path +
         "openssl smime -inform der -verify -noverify -in %s)" % profile_path)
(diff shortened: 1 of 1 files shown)


pwnartist:
I signed it!

thomasvl:
Can you provide an example of what was failing or better describe the problem?  From the looks of things, you might not have set the `provisioning_profile` attribute on the your target and that is why you are trying to just use a file system `find` to pick one that happens to be sitting around.

pwnartist:
I set the `provisioning_profile` to `org_artofthings_test.mobileprovision` into `ios/BUILD`.  

you can show "provisioning profile could not found in sandbox" at "https://github.com/pwnartist/bazel-ios-app". 
Lower version of bazel ignores the error and go to build reminders. 
0.11.1-homebrew of bazel stop to going build because the not found error.

Error messages was same as belows:

```shell
ERROR: /Users/hackartist/Data/devel/bazel-ios-app/ios/BUILD:15:1: Processing and signing iOS application: test failed (Exit 1)
Could not extract plist from provisioning profile
security: unable to open "ios/org_artofthings_test.mobileprovision" for reading: No such file or directory
Can't open input file ios/org_artofthings_test.mobileprovision
140735902237576:error:02001002:system library:fopen:No such file or directory:/BuildRoot/Library/Caches/com.apple.xbs/Sources/libressl/libressl-22/libressl/crypto/bio/bss_file.c:122:fopen('ios/org_artofthings_test.mobileprovision', 'rb')
140735902237576:error:2006D080:BIO routines:BIO_new_file:no such file:/BuildRoot/Library/Caches/com.apple.xbs/Sources/libressl/libressl-22/libressl/crypto/bio/bss_file.c:125:
bazel-out/darwin-fastbuild/bin/ios/test.archive-root/Payload/test.app/Frameworks/libswiftCore.dylib: replacing existing signature
bazel-out/darwin-fastbuild/bin/ios/test.archive-root/Payload/test.app/Frameworks/libswiftCore.dylib: unknown error -1=ffffffffffffffff
Target //ios:test failed to build
```

thomasvl:
Looking at your BUILD file, the way you [set the attribute](https://github.com/pwnartist/bazel-ios-app/blob/master/ios/BUILD#L19) it would expect to find the file in that directory.  But I don't see it in your repo, was the file physically in that directory when you did your build?

pwnartist:
Practically, I compiled with my provisioning profile. Due to security reason, I did not push my provisioning file.
Additionally, `bazel` successfully copies `org_artofthings_test.mobileprovision` into sandbox env. as `embbedded.mobileprovision`.

pwnartist:
Why, I used `find` to select a mobileprovision, is that I am unskilled for skylark, which is a language for `bazel`. If you have an idea, please let me know the graceful solution for that.

allevato:
I believe the issue you've found has a much simpler fix—in [this function](https://github.com/bazelbuild/rules_apple/blob/master/apple/bundling/bundler.bzl#L386), we pass the entitlements file but not the provisioning profile as an input to the bundling action. So the sandbox doesn't get to see the latter.

The use of `find` here isn't the right way to solve it so I'm going to close out this PR, but thanks for pointing us at the issue! We'll roll out a fix shortly.

thomasvl:
#153 should have fixed the issue @allevato mentioned.

pwnartist:
@allevato , could you examine how I can pass `skip_signing` attribute from `BUILD` file to [this function](https://github.com/bazelbuild/rules_apple/blob/master/apple/bundling/bundler.bzl#L386).

pwnartist:
thank @allevato and @thomasvl, I confirmed and tested [your commit](https://github.com/bazelbuild/rules_apple/commit/3beff7faf662234c121a5fb02098b92ebae7479b).
All of signing action was completely done.

PR #159, opened 2018-03-27
Updates

sergiocampama:
* Integration tests updates
* Bundler keeps executable bits for already executable resources


thomasvl:
CLA is fine since no-reply@google.com isn't a real account.

PR #166, opened 2018-04-09
Remove unknown bundle identifier constraint

jerrymarino:
Apple recommends a reverse-domain style string, so apps that use other
formats will no longer work with this release of rules apple.

Since it is not feasible to change bundle IDs to work with
`rules_apple`, this patch removes the constraint altogether since it is
not enforced by Apple.

### DIFF
--- a/apple/bundling/bundling_support.bzl
+++ b/apple/bundling/bundling_support.bzl
@@ -284,8 +284,6 @@ def _validate_bundle_id(bundle_id):
       for ch in part:
         if ch != "-" and not ch.isalnum():
           fail("Invalid character(s) in bundle_id: \"%s\"" % bundle_id)
-  if len(bundle_id_parts) < 2:
-    fail("bundle_id isn't at least 2 segments: \"%s\"" % bundle_id)
 
 
 # Define the loadable module that lists the exported symbols in this file.
--- a/test/ios_application_test.sh
+++ b/test/ios_application_test.sh
@@ -816,24 +816,6 @@ EOF
   expect_log "Invalid character(s) in bundle_id: \"my#bundle\""
 }
 
-function test_build_fails_if_bundle_id_too_short() {
-  create_common_files
-
-  cat >> app/BUILD <<EOF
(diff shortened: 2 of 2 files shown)


jerrymarino:
+ @DavidGoldman  - what do you think Sir. We need this for Pinterest :)

rahul-malik:
+ @sergiocampama @allevato - PTAL, Pinterest has, since it's creation, used `pinterest` as it's bundle identifier and we cannot change that. This is more of a best practice than something that will actually ever be enforced by Apple

thomasvl:
`test_build_fails_if_bundle_id_too_short` needs to get removed as part of the patch: https://github.com/bazelbuild/rules_apple/blob/master/test/ios_application_test.sh#L819-L835

Looks like travis properly flagged the test as failing.

jerrymarino:
@thomasvl thank you kindly for the review, I just pushed an update for this.

apple/bundling/bundling_support.bzl
```diff
@@ -284,9 +284,6 @@ def _validate_bundle_id(bundle_id):
       for ch in part:
         if ch != "-" and not ch.isalnum():
           fail("Invalid character(s) in bundle_id: \"%s\"" % bundle_id)
-  if len(bundle_id_parts) < 2:
-    fail("bundle_id isn't at least 2 segments: \"%s\"" % bundle_id)
-
```
sergiocampama: please keep the 2 blank lines separation

sergiocampama:
Thanks Jerry! I'll merge this after we finish the pushes we're doing currently

apple/bundling/bundling_support.bzl
jerrymarino: Ah good catch - updated in the latest revision!

jerrymarino:
@sergiocampama Awesome! Thanks for the review, Sir - much appreciated 👍

rahul-malik:
Thanks @sergiocampama @thomasvl for the quick review!

Issue #172, opened 2018-04-24
Support mobileprovision with multiple signing identities


steeve:
Currently the code signing code takes the first identity in the `mobileprovision` file. This breaks device builds for teams with multiple identities in the `mobileprovision`.
https://github.com/bazelbuild/rules_apple/blob/master/apple/bundling/codesigning_support.bzl#L85

We have a patch for that that matches the available identities on the machine with the ones in the `mobileprovision` and uses that to sign. This works well.
The issue however is that since the signing with happen with the identity of the developer, the IPA will not be reproducible, and thus should not be cached (at least not remotely).
Before submitting a PR on that we'd like to make sure this is the right approach and we didn't miss the correct way to do it in a standard dev workflow.

Thanks

thomasvl:
Interesting, personally I haven't used a profile like this, I'll try to see if I can get one to play with.

It might make sense to go ahead and propose your PR to have the discussions about how to go about options for handing this.

sergiocampama:
hey @steeve, could you try passing the flag `--ios_signing_cert_name=iPhone Developer: [DEVELOPER NAME]"` (replace the developer certificate name). If this works, the easiest way to use this would be to set up a `~/.bazelrc` file with the following content:

```
build --ios_signing_cert_name="iPhone Developer: [DEVELOPER NAME]"
```

re: Remote cache, true, it would be lost but only for the output of the codesigning action, everything else should be cached. If you want to keep the cache, you may want to consider an approach where all developers have the same development certificate, which scales fairly well for large teams.

steeve:
@sergiocampama I just tried, it doesn't make a difference. In fact, this flag seems to be implemented in legacy native apple rules and rules_apple reimplements a codesigning path itself

steeve:
@thomasvl just create one with more than one person in the team, and don't be that first person, it should work. or rather, not work.

On a side note, codesigning failure doesn't make the build fail anymore. I'll open a separate PR for that.

steeve:
As for remote cache, since it's usually the leaf target, I don't think this is a big deal. If it is, or if `mobileprovision` only contains one identity, perhaps this could be a user options so that `no-cache` may or may not be set ?

sergiocampama:
[...]

PR #178, opened 2018-05-04
Add codesigntool to handle signing identity

ffreling:
This is a draft to engage conversation on the issue #172 

In order to support profiles with multiple identities, I replaced the shell command managing signing identites with a python tool that wraps `codesign`.

Naming can definitely be improved.

I am looking forward to your feedback.

### DIFF
--- a/apple/bundling/macos_rules.bzl
+++ b/apple/bundling/macos_rules.bzl
@@ -330,7 +330,7 @@ def _macos_command_line_application_impl(ctx):
         None,
     )
 
-    inputs = [ctx.file.binary]
+    inputs = [ctx.file.binary, ctx.executable._codesigningtool]
 
     platform_support.xcode_env_action(
         ctx,
--- a/apple/bundling/rule_factory.bzl
+++ b/apple/bundling/rule_factory.bzl
@@ -316,6 +316,11 @@ def _code_signing_attributes(code_signing):
         "_skip_simulator_signing_allowed": attr.bool(
             default = code_signing.skip_simulator_signing_allowed,
         ),
+        "_codesigningtool": attr.label(
+            cfg="host",
+            executable=True,
+            default=Label("@build_bazel_rules_apple//tools/codesigningtool"),
+        ),
     }
     if not code_signing.skip_signing:
         code_signing_attrs["entitlements"] = attr.label(
--- a/apple/bundling/bundler.bzl
+++ b/apple/bundling/bundler.bzl
@@ -579,6 +579,8 @@ def _process_and_sign_archive(
             progress_description,
             bundle_name,
         ),
+        no_sandbox=True,
+        no_cache=True,
         tools = [
             ctx.executable._codesigningtool,
         ]
--- a/apple/bundling/bundler.bzl
+++ b/apple/bundling/bundler.bzl
@@ -579,6 +579,9 @@ def _process_and_sign_archive(
             progress_description,
             bundle_name,
         ),
+        tools = [
+            ctx.executable._codesigningtool,
+        ]
     )
     return (work_dir, entitlements)
 
--- a/apple/bundling/codesigning_support.bzl
+++ b/apple/bundling/codesigning_support.bzl
@@ -31,113 +31,6 @@ load(
     "define_utils",
 )
 
-def _extract_provisioning_plist_command(ctx, provisioning_profile):
-    """Returns the shell command to extract a plist from a provisioning profile.
-
-    Args:
-      ctx: The Skylark context.
-      provisioning_profile: The `File` representing the provisioning profile.
-
-    Returns:
-      The shell command used to extract the plist.
-    """
-    if mock_support.is_provisioning_mocked(ctx):
-        # If provisioning is mocked, treat the provisioning profile as a plain XML
-        # plist without a signature.
-        return "cat " + shell.quote(provisioning_profile.path)
-    else:
-        # NOTE: Until the bundling rules are updated to merge entitlements support
-        # and signing, this extraction command should be kept in sync with what
-        # exists in provisioning_profile_tool.
-        #
-        # Use a fallback mechanism to call first the security command and if that
-        # fails (e.g. when running in El Capitan) call the openssl command.
-        # The whole output for that fallback command group is then rerouted to
-        # STDERR which is only printed if the command actually failed (security and
-        # openssl print information into stderr even if the command succeeded).
-        profile_path = shell.quote(provisioning_profile.path)
-        extract_plist_cmd = (
-            "(security cms -D -i %s || " % profile_path +
-            "openssl smime -inform der -verify -noverify -in %s)" % profile_path
-        )
-        return ("( " +
-                "STDERR=$(mktemp -t openssl.stderr) && " +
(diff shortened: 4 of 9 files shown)


ffreling:
`I signed it!`

thomasvl:
So the intent is rather then to pass the identity to use, to instead of it auto figured out during the signing?

If someone were using a remote cache, wouldn't this get into trouble since without the extra args, signing by different folks would be different, but the args going into the cache keys will be the same.

steeve:
Yes this is one the issues raised in #172: signing without specifying an identity is not reproducible and thus should not be cached.
Also, the `process_and_sign` rule should perhaps be split in several actions.

steeve:
Or perhaps the identity should be output in a specific file, which could then be read by the code signing tool. This way only the commands using the identity file would become naturally invalidated, should a remote cache be used with multi-identity mobileprovision.

thomasvl:
Looks like some of the test failures are real, the rules had some support for mocking signin for testing, so you'll likely need to ensure that still working.  Also I think it is the macOS command line rule that is an odd one, and doesn't use the normal factory for making it's attributes, so you likely will have to add the new tool directly to it's attributes.

As far as the caching issue, I don't think there is a way to mark something as "don't cache" from bazels pov.  So splitting out an action to extract the identity into a file and then passing that into the next action to give bazel some exposure would likely be a good thing.

That split would also help with something we've been debating; when collecting the swift support libraries, we can tell it to resign them in advance for us, but we need the identity to pass, so having an action to do the extraction could be leveraged for that also.

ffreling:
@thomasvl sorry for the delay, I was a bit busy lately.

As @steeve said, our objective is two-fold:

1) filter the possible signing identities to only select one we can sign with

The previous behavior was to select the first identity from the provisioning
profile. Now we take the first one that has a matching certificate.

2) replace the signing logic written in shell with a python tool to make it more
readable

Indeed with this approach, we broke the caching mecanism because with the same
input, we might use a separate signing identity.

We could either split the bundling process in 3:
1. unpack (+ optional archive post processing)
2. sign (with `no-cache`)
3. compress and pack

The signing stage would be explicitely tagged `no-cache`.

Or we could export the signing identity in its own file, invalidating the cache
for commands using the identity (in our case, the signing tool).

Is there a preferred approach?

thomasvl:
No worries, the busy thing hit on this end also.  😄 

Forgot to mention before, the fact that bundling and signing is monolithic is on our radar, it is becoming more of an issue for folks doing iterative development on a device.  So we're also starting to look at splitting this up to allow frameworks to be caches after they are signed, and reducing what has to be resigned on each _edit/build/debug_ cycle.

steeve:
@thomasvl sorry for the long wait, but would it be a satisfying solution for you to write off the package + codesign task as `no-cache` ?

steeve:
Namely the `ProcessAndSign` action

steeve:
I signed it !

ffreling:
I confirm that I am working with @steeve on this issue. I am ok with his changes.

steeve:
Pushed a branch to fix a typo that would break signing for the simulator

thomasvl:
From looking at the `178.patch` all the `From:` entries seem to have signed the CLA, so not sure why the bot is grumpy.

thomasvl:
Looks like some of the test failures may still be real?

steeve:
Looking into it !

steeve:
I don't get why it's failing, as the the tool is declared in the `_bundling_tool_attributes`

steeve:
Looks like I can reproduce the bug locally, will fix!

thomasvl:
Looks like it is `macos_command_line_application` that doesn't use the factory, but it does use `bundler.run()`.  So you'll need to ensure the new attribute makes it in there.

steeve:
Alright, found the issue, actually I improved the PR a bit. Tests looking good so far.

steeve:
I've added `_codesigntool` to `_code_signing_attributes`. All the tests are passing except for this one:
```
//test:ios_test_runner_unit_test                                         FAILED in 639.5s
  /private/var/tmp/_bazel_steeve/bf36c2a2daf0f473d8970113dcf65891/execroot/build_bazel_rules_apple/bazel-out/darwin-fastbuild/testlogs/test/ios_test_runner_unit_test/test.log```


The log says:
```
Target //ios:PassingUnitTestWithHost up-to-date:
  test-bin/ios/PassingUnitTestWithHost.zip
  test-bin/ios/PassingUnitTestWithHost
INFO: Elapsed time: 177.774s, Critical Path: 175.93s
INFO: 9 processes: 2 local, 7 processwrapper-sandbox.
INFO: Build completed, 1 test FAILED, 18 total actions
//ios:PassingUnitTestWithHost                                            FAILED in 169.2s
  /private/var/tmp/_bazel_steeve/bf36c2a2daf0f473d8970113dcf65891/sandbox/darwin-sandbox/28/execroot/build_bazel_rules_apple/_tmp/b57706bd1cb92d6fe477b01655fae99d/_bazel_steeve/22fb67d085c87e18a9d8f32c81210ced/execroot/build_bazel_rules_apple_integration_tests/bazel-out/darwin-fastbuild/testlogs/ios/PassingUnitTestWithHost/test.log

Executed 1 out of 1 test: 1 fails locally.
INFO: Build completed, 1 test FAILED, 18 total actions```

However I can't check the sublog.

apple/bundling/rule_factory.bzl
```diff
@@ -322,6 +322,11 @@ def _code_signing_attributes(code_signing):
       "_skip_simulator_signing_allowed": attr.bool(
           default=code_signing.skip_simulator_signing_allowed
       ),
+      "_codesigningtool": attr.label(
```
steeve: Relevant part here @thomasvl

apple/bundling/macos_rules.bzl
```diff
@@ -289,7 +289,7 @@ def _macos_command_line_application_impl(ctx):
   signing_commands = codesigning_support.signing_command_lines(
       ctx, [path_to_sign], None)
 
-  inputs = [ctx.file.binary]
+  inputs = [ctx.file.binary, ctx.executable._codesigningtool]
```
steeve: Relevant part here @thomasvl

steeve:
Ah crap wait, made a typo

steeve:
@thomasvl you need to reset the CLA again..

steeve:
I'm going to rebase this, but should the tests pass, are you satisfied with the solution so far @thomasvl ?

steeve:
I'm also going to remove @ffreling commits (he works with me) so that the CLA doesn't resets everytime. Are you fine @ffreling ?

ffreling:
@steeve I'm fine with that.

thomasvl:
Hanging this off to @sergiocampama since he's been working on the bundler changes lately and starting to think about signing improvements.

steeve:
Thanks guys!
We can open a new PR if the CLA resetting is a pain for everyone.

steeve:
hey @sergiocampama, is there anything that's missing or blocking on your side for this to move forward ?

ffreling:
Closing the issue, I'll let @steeve deal with it in his own PR.

PR #189, opened 2018-06-14
Various updates.

allevato:


thomasvl:
CL is fine, just exporting things.

allevato:
The Travis failure is a buildifier failure from an unrelated file; submitting this and we can fix that up later.

PR #194, opened 2018-06-20
ios_framework: Allow bundle_name argument to override name when constructing rpath.

### LINKED ISSUE #193: ios_framework path should use bundle_name if specified, not build target name
I tried to declare an embedded framework target like so:
```
ios_framework(
    name = "Foo.framework",
    bundle_name = "Foo",
    families = ["iphone"],
    ...
)
```

But this causes runtime errors when the framework is embedded:

**dyld: Library not loaded: (@)rpath/Foo.framework.framework/Foo.framework**
**Reason: image not found**

The binary's rpath gets set to:
```
@rpath/Foo.framework.framework/Foo.framework
```

We shouldn't assume that the bazel target name is the bundle name / binary file name.  If bundle_name is specified, it should be used instead construct the rpath.

Expected rpath:
```
@rpath/Foo.framework/Foo
```

dantreiman:
fixes #193

### DIFF
--- a/apple/ios.bzl
+++ b/apple/ios.bzl
@@ -258,7 +258,8 @@ def ios_framework(name, **kwargs):
           final framework.
     """
     linkopts = kwargs.pop("linkopts", [])
-    linkopts += ["-install_name", "@rpath/%s.framework/%s" % (name, name)]
+	bundle_name = kwargs.get("bundle_name", name)
+    linkopts += ["-install_name", "@rpath/%s.framework/%s" % (bundle_name, bundle_name)]
 
     # Link the executable from any library deps and sources provided.
     bundling_args = binary_support.create_linked_binary_target(
--- a/apple/ios.bzl
+++ b/apple/ios.bzl
@@ -258,7 +258,7 @@ def ios_framework(name, **kwargs):
           final framework.
     """
     linkopts = kwargs.pop("linkopts", [])
-	bundle_name = kwargs.get("bundle_name", name)
+    bundle_name = kwargs.get("bundle_name", name)
     linkopts += ["-install_name", "@rpath/%s.framework/%s" % (bundle_name, bundle_name)]
 
     # Link the executable from any library deps and sources provided.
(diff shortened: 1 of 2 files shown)


apple/ios.bzl
```diff
@@ -251,10 +253,11 @@ def ios_framework(name, **kwargs):
         final framework.
   """
   deps = kwargs.pop("deps", [])
-  apple_dylib_name = "%s.apple_binary" % name
+  bundle_name = kwargs.get("bundle_name", name)
+  apple_dylib_name = "%s.apple_binary" % bundle_name
```
allevato: This would change the name of the underlying `apple_binary` target, which is risky. It could more easily cause collisions between BUILD targets—especially ones that you can't directly see because they're hidden by macros).

We should already be using `bundle_name` as the binary name [here](https://github.com/bazelbuild/rules_apple/blob/master/apple/bundling/bundler.bzl#L928-L929), so your change should still produce the same result if you revert this one line back to `... % name`. Can you try that and verify?

dantreiman: You are right, and it is unnecessary change the apple_binary name.  Reverted that line, Thanks!

allevato:
My apologies for losing track of this pull request!

Thanks for making the requested change. Could you please resolve the conflicts and add a test in `ios_framework_test.sh` so that we verify that this works (and don't break it later)? Then we can merge this.

dantreiman:
@allevato I've synced this branch up and resolved conflicts.  However, I'm not sure how to go about running the tests.

I tried
```
bazel test //test:ios_framework_test
```
and I get an error:
```
ERROR: Analysis of target '//test:ios_framework_test.simulator' failed; build aborted: no such package '@bazel_tools//tools/build_defs/cc': BUILD file not found on package path
```

What am I missing?

bazel 0.15.2, Xcode 9.4.1 on macOS 10.13.6

allevato:
Please try updating to Bazel 0.16.1. Some of the more recent changes to the rules that haven't yet been cut in a release require the latest version of Bazel.

allevato:
Looks great. Thanks for the contribution!

PR #201, opened 2018-08-08
Actually delete temporary simulator after use

kastiglione:
A simulator is created for each test run, but is not deleted as expected after completion. This is happening because an `EXIT` trap is specified twice, with the second (cleanup for `$TEMP_DIR`) overwriting the first (cleanup of the temporary simulator).

This change combines the two cleanups into a single trap command.

### DIFF
--- a/apple/bundling/runners/ios_sim.template.sh
+++ b/apple/bundling/runners/ios_sim.template.sh
@@ -110,6 +110,7 @@ function CleanupSimulator() {
 }
 
 readonly STD_REDIRECT_DYLIB="$PWD/%std_redirect_dylib_path%"
+
 readonly TEMP_DIR=$(mktemp -d "${TMPDIR:-/tmp}/bazel_temp.XXXXXX")
 
 trap 'rm -rf "${TEMP_DIR}"; CleanupSimulator ${TEST_DEVICE_ID}' ERR EXIT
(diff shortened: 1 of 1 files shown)


kastiglione:
I signed it!

apple/bundling/runners/ios_sim.template.sh
```diff
@@ -109,12 +109,12 @@ function CleanupSimulator() {
...
 
-trap "CleanupSimulator ${TEST_DEVICE_ID}" EXIT
-
 readonly STD_REDIRECT_DYLIB="$PWD/%std_redirect_dylib_path%"
 
 readonly TEMP_DIR=$(mktemp -d "${TMPDIR:-/tmp}/bazel_temp.XXXXXX")
-trap 'rm -rf "${TEMP_DIR}"' ERR EXIT
+
+trap 'rm -rf "${TEMP_DIR}"' ERR
+trap 'rm -rf "${TEMP_DIR}" && CleanupSimulator ${TEST_DEVICE_ID}' EXIT
```
kastiglione: Alternatively, both `ERR` and `EXIT` should do both cleanups.

kastiglione: @sergiocampama Done.

kastiglione:
Thanks!

PR #214, opened 2018-08-14
Add output message filtering for ibtool. STDERR messages are only pri…

allevato:
…nted if ibtool terminates with a non-zero exit value.

RELNOTES: None
PiperOrigin-RevId: 208681202


allevato:
Need to add the init.py shim. Abandoning and repushing.

PR #218, opened 2018-08-16
Force bsd sed on macOS

keith:
If a user has gnu-sed in their PATH, these commands would fail since the
`-i` arguments are incompatible. Since this script will only run on
macOS, we can be sure using the absolute path will be valid.


keith:
Thanks!

Issue #226, opened 2018-09-19
Support custom substitutions in plists and entitlements files


keith:
Currently in our app we have some custom settings that we use across code, and entitlements files. To do this we define custom build settings, and reference them from our plist, and entitlements files, and then read the plist value in code. It ends up looking something like this:

```
# foo.xcconfig
APP_GROUP_SUITE_NAME = some id
```

```
# foo.entitlements
<key>com.apple.security.application-groups</key>
<array>
  <string>$(APP_GROUP_SUITE_NAME)</string>
</array>
```

```
# info.plist
<key>AppGroupSuiteName</key>
<string>${APP_GROUP_SUITE_NAME}</string>
```

```
# foo.swift
let appGroupSuiteName = Bundle.main.infoDictionary?["LFTAppGroupSuiteName"] as! String
```

I'm wondering if there's a specific reason bazel doesn't support custom substitutions to support this use case, or if this is something we could add to some of the rules.

thomasvl:
This has come up before, the complexity is always how go to about expressing it in the BUILD files.

Since you'd likely end up needing to use `select` statements to conditionally assign values, you can use those same `select` statements to provide different fragments of info.plist files and since the rules take a list of files, they will get merged.

Another thing we've seen people do is write a simple rule to wrap [expand_template](https://docs.bazel.build/versions/master/skylark/lib/actions.html#expand_template), and simply use that to do their subs and feed the result in to the other rules.

keith:
> you can use those same select statements to provide different fragments of info.plist files and since the rules take a list of files, they will get merged.

It seems like the downside for us would be checking in many plists just for this use case. For us we have a 6 configurations with different values for ~6 different keys, things like default oauth config, etc, so we'd have to check in quite a few plists.

We also currently use variables like this for our versioning, but I guess there might be an alternative to that with a different method in bazel

keith:
Using the change here https://github.com/bazelbuild/rules_apple/pull/230 I was able to implement a simple custom rule for this:

```python
def _replace_plist_variables(ctx):
    input_file = ctx.file.plist
    output_plist = ctx.outputs.out

    control = struct(
        output = output_plist.path,
        plists = [input_file.path],
        target = str(ctx.label),
        variable_substitutions = struct(**ctx.attr.replacements),
        allow_missing_variables = True,
    )

    control_file = ctx.actions.declare_file(
        "%s-control" % output_plist.basename,
        sibling = output_plist,
    )
[...]

PR #230, opened 2018-09-26
Allow plists to not be validated for missing variables

keith:
This adds a new control key for plisttool to not validate that all keys
have been replaced after processing. This is useful if you have a custom
rule that replaces some plist keys, but leaves the defaults around to be
processed later. This allows users to implement a simple rule to work
around this limitation https://github.com/bazelbuild/rules_apple/issues/226

### DIFF
--- a/tools/plisttool/plisttool.py
+++ b/tools/plisttool/plisttool.py
@@ -54,6 +54,9 @@ the following keys:
       substitutions when processing the plists. All keys/values will get
       support for the rfc1034identifier qualifier.
   target: The target name, used for warning/error messages.
+  allow_missing_variables: Don't validate all variables are replaced after
+      processing the plist. This is useful if there will be multiple processing
+      steps.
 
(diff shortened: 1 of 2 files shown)


keith:
Here's an example of a rule that uses this https://github.com/bazelbuild/rules_apple/issues/226#issuecomment-424892945

thomasvl:
It doesn't seem like plist_tool should be looped into this the problem you describe.  [expand_template](https://docs.bazel.build/versions/master/skylark/lib/actions.html#expand_template) could be wrapped in a simple rule, you could preprocess your plist with that, and then feed it to the bundle rules for plist_tool to do the final processing.  It sorta feels like you'd be going out of your way to feed things through plist_tool which is what requires it then get more controls making it more complex.

keith:
Ah I wasn't aware expand_template was this flexible. Perfect thanks.

PR #236, opened 2018-10-04
Update macOS test bundle injection path

keith:
The path to this binary changed with Xcode 10

Related https://github.com/facebook/xctool/pull/755/

### DIFF
--- a/apple/testing/default_runner/macos_test_runner.bzl
+++ b/apple/testing/default_runner/macos_test_runner.bzl
@@ -19,9 +19,17 @@ load(
     "AppleTestRunner",
 )
 
+def _get_bundle_inject_path(ctx):
+    xcode_version = ctx.attr._xcode_config[apple_common.XcodeVersionConfig].xcode_version()
+    if xcode_version and xcode_version < apple_common.dotted_version("10.0"):
+        return "Library/PrivateFrameworks/IDEBundleInjection.framework/IDEBundleInjection"
+
+    return "usr/lib/libXCTestBundleInject.dylib"
+
 def _get_template_substitutions(ctx):
     """Returns the template substitutions for this runner."""
     subs = {
+        "xctestrun_bundle_inject_path": _get_bundle_inject_path(ctx),
         "xctestrun_template": ctx.file._xctestrun_template.short_path,
     }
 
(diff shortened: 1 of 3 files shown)


sergiocampama:
sorry it took so long to push out, but this should be fixed in the PR that thomasvl mentioned. That PR branch is already passing, so it'll be merged later today

sergiocampama:
can you check that your tests pass by using the version at HEAD?

keith:
Looks like I'm seeing some new errors, likely unrelated:

```
ERROR: .../BUILD:51:1: in deps attribute of swift_runtime_linkopts rule //....swift_runtime_linkopts: rule '//...' does not exist. Since this rule was created by the macro 'macos_unit_test', the error might have been caused by the macro implementation in .../external/build_bazel_rules_apple/apple/bundling/binary_support.bzl:203:16
```

keith:
Ah it looks like this was a red herring of misconfiguration of a commented out target

keith:
Now my tests just run forever, I'll try and debug more

sergiocampama:
so it appears that Xcode 10 is trying to write data into DerivedData, which is outside of the sandbox that bazel creates. I can run the example `macos_unit_tests` but only if I pass the `--spawn_strategy=local` flag

keith:
Somehow with my change this was working locally for me.


--
Keith Smiley

sergiocampama:
Can you confirm if you can run the example tests? Or provide a small repro case where the tests hang?

keith:
Hmm yea it looks like `bazel test //test:macos_application_resources_test` fails for me either way

keith:
Actually sorry it looks like I didn't give this enough time. It looks like this test passes for me on master even without `--spawn_strategy=local`

keith:
There must be a problem with our test suite, I'll try and debug more

keith:
I am seeing this log, although it might not matter:

```
2018-10-04 23:42:26.855 xctest[16601:6628902] Bundle injected into '/Applications/Xcode-10.0.app/Contents/Developer/Platforms/MacOSX.platform/Developer/Library/Xcode/Agents/xctest' but XCInjectBundleInto environment variable set to ''
```

keith:
Ok it looks like I spoke too soon about there being an issue, it just so happens that the specific tests I'm working with crash when they fail, and that seems to add a huge amount of time to the overall runtime, so I was never patient enough to see it through. They run correctly even with the sandbox enabled.

Thanks for the fix!

PR #238, opened 2018-10-29
Update for future incompatible bazel changes

keith:
These fixes are required to run bazel 0.19.0 with
`--all_incompatible_changes`. There are a few specific transformations I
did here:

1. If the rule's attributes contained an `allow_files` array and
`single_file = True`, I converted this to `allow_single_file` with the
same array.

2. If the attribute had `non_empty = False`, I removed it because this
would be migrated to `allow_empty = True` which is the default (if we
want to be more explicit, I can add this where I removed them)

3. If the attribute had only `single_file = True` it is now
`allow_single_file = True`

4. If the attribute had `allow_files = True` and `single_file = True`,
it is now `allow_single_file = True`

### DIFF
--- a/apple/bundling/entitlements.bzl
+++ b/apple/bundling/entitlements.bzl
@@ -302,8 +302,7 @@ entitlements = rule(
             mandatory = True,
         ),
         "entitlements": attr.label(
-            allow_files = [".entitlements", ".plist"],
-            single_file = True,
+            allow_single_file = [".entitlements", ".plist"],
         ),
         "_plisttool": attr.label(
             cfg = "host",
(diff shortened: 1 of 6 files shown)


sergiocampama:
I just updated the repo to the latest internal version, could you rebase and verify this change still works? it should be ok since the new files use the new APIs

keith:
I'm also on vacation for a while, I will rebase it when I'm back.


--
Keith Smiley

keith:
Rebased and we're still looking good

sergiocampama:
I've checked in this change to the internal branch. There's a PR out with the commit, but I'm also doing some testing with that PR to see if I can reduce flaky tests.

PR #244, opened 2018-11-13
Update

sergiocampama:


sergiocampama:
The error in travis looks like a flake. Looks like other companies are also seeing it based on #245

PR #246, opened 2018-11-14
Use --ios_simulator_device argument when testing a ios test target.

chenxiao0228:
The arg also overrides device_type specified in a custom test runner.

### DIFF
--- a/apple/testing/default_runner/ios_test_runner.bzl
+++ b/apple/testing/default_runner/ios_test_runner.bzl
@@ -28,6 +28,11 @@ def _get_template_substitutions(ctx):
         "testrunner_binary": ctx.executable._testrunner.short_path,
         "device_type": ctx.attr.device_type,
     }
+    # if ios_simulator_device and ios_simulator_version are specified in the bazel test command, override device_type
+    if ctx.fragments.objc.ios_simulator_device:
+        subs["device_type"] = ctx.fragments.objc.ios_simulator_device
+    if ctx.fragments.objc.ios_simulator_device:
+        subs["os_version"] = str(ctx.fragments.objc.ios_simulator_version)
     return {"%(" + k + ")s": subs[k] for k in subs}
 
 def _get_test_environment(ctx):


rahul-malik:
This is awesome! Can we also have `--ios_simulator_version` handled as well?

chenxiao0228:
> This is awesome! Can we also have `--ios_simulator_version` handled as well?

Updated PR

apple/testing/default_runner/ios_test_runner.bzl
```diff
@@ -28,6 +28,11 @@ def _get_template_substitutions(ctx):
         "testrunner_binary": ctx.executable._testrunner.short_path,
         "device_type": ctx.attr.device_type,
     }
+    # if ios_simulator_device and ios_simulator_version are specified in the bazel test command, override device_type
```
sergiocampama: overriding the `device_type` like this is problematic, as debugging will get tricky when a test is running on an iPhone 6 when the `device_type` of the test runner is iPhone 7. another problem is overriding the device type for all tests in  running tests with `//...`, as it will override any device configured in the targets that you might have in your workspace

the approach I think this should take is to resolve `device_type` like this:

1. if `device_type` attribute is set, use that.
2. if `device_type` is not set, use the `ios_simulator_device` from the `objc` fragment
3 if the `ios_simulator_device` is not set, use a default of "iPhone 6" (for example)

same approach with `os_version`:

1. if `os_version` is set, use that
2. if `os_version` not set, use the `ios_simulator_version` value from the `objc` fragment
3. if `ios_simulator_version` is not set, use the iOS SDK value from the `apple_common.XcodeVersionConfig` provider

apple/testing/default_runner/ios_test_runner.bzl
```diff
@@ -28,6 +28,11 @@ def _get_template_substitutions(ctx):
...
+    if ctx.fragments.objc.ios_simulator_device:
+        subs["device_type"] = ctx.fragments.objc.ios_simulator_device
+    if ctx.fragments.objc.ios_simulator_device:
```
sergiocampama: I think you meant `ios_simulator_version` here

apple/testing/default_runner/ios_test_runner.bzl
chenxiao0228: Great idea. I will update the PR

chenxiao0228:
Abandoning this PR for now. Since trying to make the logic right we'd need to touch core bazel.
Creating custom test runners for now as a workaround.

PR #247, opened 2018-11-15
Update

sergiocampama:


keith:
I see this test fail locally with Xcode 10.1 as well, almost 100% of the time.

keith:
This seems to be a problem with the sequence.stickersequence. If I delete that directory, I can't seem to get the failure (after modifying the test to not expect it).

I've tried re-creating the sequence with Xcode, to see if something about the Contents.json has changed, but don't end up with any difference, and still end up with the crash.

If I run `bazel clean && bazel test //test:ios_extension_resources_test.device` without this change, I see this almost every time.

keith:
(the most recent test failure appears to be different)

sergiocampama:
I don't know what happened with the travis tests, will need to look into it further. But buildkite seems happy for now

sergiocampama:
Thanks for the tip on the stickericons error @keith, I can't repro that error on my machine though, I just get timeouts.

PR #249, opened 2018-11-20
Leak the client PATH into tests.

benjaminp:
The integration tests require the bazel executable to be on PATH.


sergiocampama:
Thanks for the change, we were just talking about it. Are you aware of which change in Bazel makes this a requirement?

sergiocampama:
Found it: https://github.com/bazelbuild/bazel/commit/3bf685794b8568042c91d789e612af34fc3c485f

benjaminp:
Yeah, this is supposed to fix https://github.com/bazelbuild/bazel/issues/6716.

It looks like `//test:ios_extension_resources_test.simulator ` is failing, but that also was happening before: https://travis-ci.org/bazelbuild/rules_apple/jobs/456041972

PR #251, opened 2018-11-21
Make apple_framework_import is_dynamic default to false

keith:
This mirrors the behavior of `objc_framework`

### DIFF
--- a/apple/internal/apple_framework_import.bzl
+++ b/apple/internal/apple_framework_import.bzl
@@ -148,7 +148,7 @@ on this target.
 """,
         ),
         "is_dynamic": attr.bool(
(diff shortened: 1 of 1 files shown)


sergiocampama:
This was designed as mandatory to make it clearer for users of the target that it is dynamic or static. While a replacement for `objc_framework`, it is a new rule, so old semantics don't need to be preserved.

kastiglione:
Was there any consideration to having two rules, like `apple_framework_import` and `apple_static_framework_import`? They essentially have different interfaces.

kastiglione:
👍 That check there is what made me wonder.

PR #253, opened 2018-11-21
Use depset for sdk_dylibs in apple_framework_import

keith:


apple/internal/apple_framework_import.bzl
```diff
@@ -109,7 +109,7 @@ def _apple_framework_import_impl(ctx):
             )
     else:
         if ctx.attr.sdk_dylibs:
-            objc_provider_fields["sdk_dylib"] = ctx.attr.sdk_dylibs
+            objc_provider_fields["sdk_dylib"] = depset(direct = ctx.attr.sdk_dylibs)
```
keith: Assuming `direct` is the right option here, but happy to update as needed

sergiocampama: please do the same below for `sdk_framework` and `weak_sdk_framework`

keith: Done!

apple/internal/apple_framework_import.bzl
```diff
@@ -109,11 +109,11 @@ def _apple_framework_import_impl(ctx):
...
         if ctx.attr.sdk_frameworks:
-            objc_provider_fields["sdk_framework"] = ctx.attr.sdk_frameworks
+            objc_provider_fields["sdk_framework"] = depset(direct = ctx.attr.sdk_frameworks)
         if ctx.attr.weak_sdk_frameworks:
-            objc_provider_fields["weak_sdk_framework"] = ctx.attr.weak_sdk_frameworks
+            objc_provider_fields["weak_sdk_framework"] = depset(direct = ctx.attr.weak_sdk_frameworks)
```
sergiocampama: nit: this line is too long, and the internal linter complains :(

using or not using `direct` isn't really a big deal, `depset([a,b])` is equivalent to `depset(direct = [a, b])`. For the sake of brevity for long lines like this, it'd be best to just remove `direct`. same for the other ones, for consistency.

keith: Done

sergiocampama:
Thanks! The change has been imported. I'll push out in a bit

apple/internal/apple_framework_import.bzl
kastiglione: I believe the positional parameter support is deprecated, and that `direct` will be required in the future.

kastiglione: https://docs.bazel.build/versions/master/skylark/lib/globals.html#depset

These are the docs that claim this. See "compatibility".

keith: Ha. I'll update this when we get hit by it with `--all_incompatible_changes`

allevato: I thought the same thing, but @sergiocampama pointed out this part of that doc to me:

> This function currently accepts a positional items parameter. It is deprecated and will be removed in the future, and after its removal direct will become a sole positional parameter of the depset function. Thus, both of the following calls are equivalent and future-proof:
>
> ```python
> depset(['a', 'b'], transitive = [...])
> depset(direct = ['a', 'b'], transitive = [...])
> ```

So it looks like the unlabeled version is safe.

kastiglione: ah, nice. Gotta work on my reading comprehension.

PR #254, opened 2018-11-21
Fix apple_framework_import docs file

keith:
This had the wrong bzl file for loading.


sergiocampama:
I submitted this change internally, it will be pushed out sometime later today (we're waiting for another set of patches before pushing)

PR #256, opened 2018-11-21
Add deps to apple_framework_import rule

keith:
This allows apple_framework_import rules to propagate dependencies that
are required for linking the framework.


keith:
The one thing this doesn't cover that we might want is the `data` attribute to propagate `apple_bundle_import` or `apple_resource_bundle` targets.

apple/internal/apple_framework_import.bzl
```diff
@@ -123,6 +126,12 @@ def _apple_framework_import_impl(ctx):
     dummy_binary = ctx.actions.declare_file("_{}.dummy_binary".format(ctx.label.name))
     ctx.actions.write(dummy_binary, "_dummy_file_")
 
+    objc_provider_fields["providers"] = [
+        dep[apple_common.Objc]
+        for dep in ctx.attr.deps
+        if apple_common.Objc in dep
```
keith: I guess I could remove this since the deps field requires it with its providers list

apple/internal/apple_framework_import.bzl
```diff
@@ -115,6 +115,9 @@ def _apple_framework_import_impl(ctx):
         if ctx.attr.weak_sdk_frameworks:
             objc_provider_fields["weak_sdk_framework"] = depset(ctx.attr.weak_sdk_frameworks)
         objc_provider_fields["static_framework_file"] = framework_imports_set
+        providers.append(
```
sergiocampama: this is missing the `framework_imports` from `deps`.

I'd recommend doing something like this before  `if ctx.attr.is_dynamic`:

```
transitive_sets = []
for dep in ctx.attr.deps:
    transitive_sets.append(dep[AppleFrameworkImportInfo].framework_imports)
```

then, inside the `if ctx.attr.is_dynamic:` replace the `if filtered_framework_imports` with:

```
if filtered_framework_imports:
  transitive_sets.append(depset(filtered_framework_imports))
```

and then just after the `if:else:` block, do:
```
providers.append(AppleFrameworkImportInfo(framework_imports = depset(transitive = transitive_sets)))
```

so that will take care of both dynamic and static cases, plus all of the `apple_framework_import` dependencies in `deps`

keith: done

apple/internal/apple_framework_import.bzl
```diff
@@ -180,6 +187,15 @@ regularly linked SDK frameworks, symbols from weakly linked frameworks do not ca
 are not present at runtime. Only applicable for static frameworks (i.e. `is_dynamic = False`).
 """,
         ),
+        "deps": attr.label_list(
+            doc = """
+A list of targets that are dependencies of the target being built, which will be
```
sergiocampama: "A list of binary import targets that are dependencies of the imported target. These dependencies will be linked in the top level binary."

Please copy this to the documentation definition in the other file

apple/internal/apple_framework_import.bzl
```diff
@@ -92,6 +92,10 @@ def _apple_framework_import_impl(ctx):
     framework_dirs_set = depset(framework_groups.keys())
     objc_provider_fields = {}
 
+    transitive_sets = []
+    for dep in ctx.attr.deps:
+        transitive_sets.append(dep[AppleFrameworkImportInfo].framework_imports)
```
sergiocampama: `if hasattr(dep[AppleFrameworkImportInfo], "framework_imports"):` to skip the providers that don't have the field

apple/internal/apple_framework_import.bzl
```diff
@@ -116,13 +118,18 @@ def _apple_framework_import_impl(ctx):
             objc_provider_fields["weak_sdk_framework"] = depset(ctx.attr.weak_sdk_frameworks)
         objc_provider_fields["static_framework_file"] = framework_imports_set
 
+    providers.append(
```
sergiocampama:
```
provider_fields = {}
if transitive_sets:
    provider_fields["framework_imports"] = depset(transitive = transitive_sets)
providers.append(
    AppleFrameworkImportInfo(**provider_fields)
)
```

this way, there's no memory overhead on empty depsets when depending on a bunch of static `apple_framework_import` targets

apple/internal/apple_framework_import.bzl
keith: Since we're requiring this in the `deps` `providers` array I don't think we need to do this? Would you still like to in case that changes in the future?

apple/internal/apple_framework_import.bzl
keith: done

apple/internal/apple_framework_import.bzl
keith: done

apple/internal/apple_framework_import.bzl
keith: Ah this is for the framework_imports field, not the provider itself because of the change below. Done

keith:
No worries, updated

PR #258, opened 2018-11-23
Create depsets outside of loop

keith:
The + operator on depset is deprecated. For this case we can instead
build up a list, and create a depset later.

### DIFF
--- a/apple/utils.bzl
+++ b/apple/utils.bzl
@@ -165,7 +165,7 @@ def group_files_by_directory(files, extensions, attr):
                 if container in grouped_files:
                     grouped_files[container] += [f]
                 else:
-                    grouped_files[container] = depset([f])
+                    grouped_files[container] = [f]
 
                 # No need to check other extensions
                 break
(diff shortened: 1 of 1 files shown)


apple/utils.bzl
```diff
@@ -165,7 +165,7 @@ def group_files_by_directory(files, extensions, attr):
                 if container in grouped_files:
                     grouped_files[container] += [f]
```
sergiocampama: can you also change this to be `grouped_files[container].append(f)` ? it'll be clearer it's a list being modified

keith: done

apple/utils.bzl
```diff
@@ -163,9 +163,9 @@ def group_files_by_directory(files, extensions, attr):
                 not_matched = False
                 container = path[:after_index]
                 if container in grouped_files:
-                    grouped_files[container] += [f]
```
kastiglione: Could this have alternatively been changed to:

```python
group = grouped_files.get(container, [])
grouped_files[container] = depset(f, transitive = group)
```

I ask this because I'm don't have a strong grasp on when to build up a list, vs when to build up a depset.

sergiocampama: yeah, that's also equivalent, but since a depset wraps a list, I believe just using a list is cleaner when possible, and then only at the end generate the depset (unless merging the depsets as you go is a cleaner implementation)

kastiglione: I think the list approach is cleaner, I'm trying to learn where it's ok to not jump straight to depsets. This gives me a sample point, thanks.

PR #259, opened 2018-11-23
Change apple_action to use run_shell

keith:


sergiocampama:
This one is not as easy to migrate, as `ctx.action` supports both `executable` and `command`, and `run_shell` only supports `command`.

We're working on cleaning this up shortly, but it involves modifying some things in bazel before we do it though.

sergiocampama:
Gonna close this for now, we're working in this internally as we speak

keith:
No worries thanks!

PR #262, opened 2018-11-27
Remove old output_licenses documentation

keith:


sergiocampama:
Imported, will be pushed out later this week

PR #263, opened 2018-11-27
Propagate .bundle directories in static frameworks

keith:
If you have a static framework that contains a resource bundle in a
structure like this:

```
Foo.framework/
Foo.framework/Foo
Foo.framework/Resources/Foo.bundle
Foo.framework/Resources/Foo.bundle/...
```

You likely want to propagate that bundle to dependent targets. For
dynamic frameworks this works by embedding the frameworks and their
contents, but for static frameworks they need to be propagated
separately.

### DIFF
--- a/apple/internal/apple_framework_import.bzl
+++ b/apple/internal/apple_framework_import.bzl
@@ -15,12 +15,24 @@
 """Implementation of apple_framework_import."""
 
 load(
-    "@build_bazel_rules_apple//apple:utils.bzl",
-    "group_files_by_directory",
+    "@bazel_skylib//lib:partial.bzl",
+    "partial",
 )
 load(
     "@bazel_skylib//lib:paths.bzl",
     "paths",
+)
+load(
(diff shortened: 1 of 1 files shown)


apple/internal/apple_framework_import.bzl
```diff
@@ -119,6 +131,21 @@ def _apple_framework_import_impl(ctx):
             objc_provider_fields["weak_sdk_framework"] = depset(ctx.attr.weak_sdk_frameworks)
         objc_provider_fields["static_framework_file"] = framework_imports_set
 
+        bundle_files = [x for x in framework_imports if ".bundle/" in x.short_path]
```
keith: Is there a more efficient solution than this? It seems like most of the functions in the `if` below require at least some files to contain `.bundle` so we can't call them without filtering first since this is optional

keith:
Fixed by https://github.com/bazelbuild/rules_apple/pull/267

PR #264, opened 2018-11-27
Remove ios_sim support for Xcode 6

keith:


kastiglione:
This is a good point to ask: What's versions of Xcode should rules_apple support? Both here and in xctestrunner, there's support for older versions, however there's no indication which versions of Xcode need to be supported, and another problem is there is code in those scripts that are clearly work arounds, but not clear which version(s) of Xcode would require them.

rahul-malik:
It seems like HEAD should support the latest major version since that’s typically required for App Store submission

sergiocampama:
Re: which versions of Xcode we support, we don't have a policy, but as @rahul-malik mentions, we try to be up to date to the latest versions of Xcode. But as you've seen, we haven't had the time to clean the repo of old Xcode workarounds

sergiocampama:

PR #269, opened 2018-12-04
Update

sergiocampama:


sergiocampama:
@thomasvl buildkite passing now

thomasvl:
Any ideas what is up with travis?  They does look like timeouts at first glance.

sergiocampama:
Not sure, it looks like the machines are somewhat slow, and that triggers timeouts all around, plus the usual flakiness of actool :(

the fact that these tests pass on buildkite (which uses imacpros and mac pros) make me believe that

PR #272, opened 2018-12-05
Fix PostProcessorError reference

kastiglione:
Fixes an error class typo `PostProcessingError` => `PostProcessorError`.

Also removes an extraneous `self` argument.


sergiocampama:

PR #279, opened 2018-12-10
Code sign support for test case

xinzhengzhang:
Code sign support for ios_unit_test and ios_ui_test.
Using for test on real device.


sergiocampama:
Why do you need this? Test bundles do not require a provisioning profile in order to run, as they're not executables. Xcode also doesn't have the option to add provisioning profiles to test targets.

xinzhengzhang:
I need run my uitest on real device. (We develop video player and the performance are total different between simulator with real one)
Test bundles do not require it only running on a simulator.
And Xcode will generate a provision file when testing on a real device  automatically (According your account logged in Xcode)
24f250d0decece711077e7f8443d601a

sergiocampama:
Ah, I see. Have you tested that this change makes this work? It's somewhat difficult to test on an iPhone since once you get it to work, it will work without these changes.

apple/testing/ios_rules.bzl
```diff
@@ -175,6 +175,9 @@ def _ios_test(
         target_name_template = "%s_test_binary",
     )
 
+    if "provisioning_profile" in kwargs:
```
sergiocampama: if you use pop instead of get, you don't need to add the attribute to the test rules above, which in any case do not get processed there.

sergiocampama: also, can you limit this to `ios_ui_test` only? `ios_unit_test` does not require a provisioning profile

xinzhengzhang:
> Ah, I see. Have you tested that this change makes this work? It's somewhat difficult to test on an iPhone since once you get it to work, it will work without these changes.

Yes, it has tested on my project.
1. test on simulator using tulsi generated project
2. test on device using tulsi generated project (Still need set provisioning profile for test runner which generated by Xcode)
3. test on simulator using `https://github.com/google/xctestrunner/`
4. test on device using device farm (Create a dummy runner using customized provisioning profile and run it on our device)

apple/testing/ios_rules.bzl
xinzhengzhang: > also, can you limit this to `ios_ui_test` only? `ios_unit_test` does not require a provisioning profile

Have you any suggestion for separating them? I have no idea where should I write it in `_ios_test` or `binary_support.create_linked_binary_target`
It seems that rules depend `entitlements` for sign and in my case we don't need to add entitlements.

sergiocampama: you can add an argument to `_ios_test` with the name `uses_provisioning_profile` and then use 

```
if uses_provisioning_profile:
    bundling_args["provisioning_profile"] = kwargs.pop("provisioning_profile", None)
```

and then add `uses_provisioning_profile = True` on `ios_ui_test` below

This way, the `ios_unit_test` will fail if the `provisioning_profile` attribute is passed, and on `ios_ui_test` tests the attribute will be filtered and only passed to the intermediate `_ios_test_bundle` target

xinzhengzhang: > you can add an argument to `_ios_test` with the name `uses_provisioning_profile` and then use
> 
> ```
> if uses_provisioning_profile:
>     bundling_args["provisioning_profile"] = kwargs.pop("provisioning_profile", None)
> ```
> and then add `uses_provisioning_profile = True` on `ios_ui_test` below
> 
> This way, the `ios_unit_test` will fail if the `provisioning_profile` attribute is passed, and on `ios_ui_test` tests the attribute will be filtered and only passed to the intermediate `_ios_test_bundle` target

Modified and rebased

apple/testing/apple_test_rules.bzl
```diff
@@ -385,6 +385,10 @@ def _apple_ui_test_attributes():
                 mandatory = True,
                 providers = [AppleBundleInfo],
             ),
+            # The provisioning_profile to codesign .xctest
+            "provisioning_profile": attr.label(
+                allow_single_file = [".mobileprovision", ".provisionprofile"],
```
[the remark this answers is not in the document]
xinzhengzhang: Fixed

PR #281, opened 2018-12-10
Remove implemented rule from 'coming soon'

keith:


keith:
Looks like something went wrong with bazel HEAD:

```
ERROR: /private/var/tmp/_bazel_travis/2b80b7ac301f661b0fa17c1c94402319/external/local_config_cc/BUILD:58:5: in apple_cc_toolchain rule @local_config_cc//:cc-compiler-ios_i386: Xcode version must be specified to use an Apple CROSSTOOL. If your Xcode version has changed recently, try: "bazel clean --expunge" to re-run Xcode configuration
```

Interestingly I've seen this locally too sometimes just between builds.

sergiocampama:
I've taken this change and added some others to this file in an internal change, will be exported soon

PR #282, opened 2018-12-10
Default iOS tests to passed versions

keith:
This makes the --ios_simulator_version and --ios_simulator_device flags,
affect which devices are used for testing if you don't otherwise provide
a device or OS version

### DIFF
--- a/apple/testing/default_runner/ios_test_runner.bzl
+++ b/apple/testing/default_runner/ios_test_runner.bzl
@@ -22,11 +22,13 @@ load(
 def _get_template_substitutions(ctx):
     """Returns the template substitutions for this runner."""
     test_env = ctx.configuration.test_env
+    os_version = ctx.attr.os_version or str(ctx.fragments.objc.ios_simulator_version)
+    device_type = ctx.attr.device_type or ctx.fragments.objc.ios_simulator_device
     subs = {
-        "os_version": ctx.attr.os_version,
+        "os_version": os_version,
         "test_env": ",".join([k + "=" + v for (k, v) in test_env.items()]),
         "testrunner_binary": ctx.executable._testrunner.short_path,
-        "device_type": ctx.attr.device_type,
+        "device_type": device_type,
     }
     return {"%(" + k + ")s": subs[k] for k in subs}
 


sergiocampama:

sergiocampama:
Actually this change breaks our current bazel tests (forgot to check buildkite before). Since our runner doesn't pass an os version, it uses the flag, which by default is set to 9.3, and buildkite does not have that os version available.

I'll revert our internal change for now to keep our tests passing while we figure out a better strategy

keith:
Is the default something that can ever be changed? We've noticed that one before too since we only support iOS 10+. I wonder if there should be a default at all.

In the more short term is passing this flag to something that does exist a viable option for the tests?

sergiocampama:
you can create a new runner rule that uses the flags only and pass that to all your test targets. Re: defaults, I think we should remove them, and have bazel return the sdk version in the current --xcode_version instead, but that change is not trivial and would take some time to propagate to released bazel

keith:
I was more thinking a solution to fix your failing tests so we could merge this. We've already created our own test runner that has this behavior but I think it's useful for others as well.

I'm also happy to make the defaults change if you think removing them is enough since that would work fine as long as something (xctestrunner) eventually picks a default

sergiocampama:
xctestrunner does indeed pick a default, the problem is that because the flags have a default, there's no way to test whether the flag was explicitly set or not. I don't think there's a way to keep this change without making the change in the flag first.

now that I think more about it, if we're going to clean up the bazel flags, we should actually remove them from bazel, since they're only used by rules_apple anyways. With that in mind, I'd like to wait for another set of features that's coming in bazel, to have Starlark defined flags. When that happens, we can migrate the flag to that feature and control the flag from rules_apple.

This should be available in Bazel in Q1, mind waiting until we have that instead?

sergiocampama:
reopening as it was closed on the update PR

keith:
This ended up being rolled back apparently https://github.com/bazelbuild/rules_apple//commit/1a5bd8e3163e6966a18a451c8bb455bb97e7a9d6

keith:
New version https://github.com/bazelbuild/rules_apple/pull/549

PR #283, opened 2018-12-10
Make ios_test_runner script pass shellcheck

keith:
Shellcheck is a very powerful shell linter. I fixed the violations of it
for this script.

### DIFF
--- a/apple/testing/default_runner/ios_test_runner.template.sh
+++ b/apple/testing/default_runner/ios_test_runner.template.sh
@@ -18,7 +18,8 @@ set -e
 
 basename_without_extension() {
   local full_path="$1"
-  local filename=$(basename "$full_path")
+  local filename
+  filename=$(basename "$full_path")
   echo "${filename%.*}"
 }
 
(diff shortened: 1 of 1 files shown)


apple/testing/default_runner/ios_test_runner.template.sh
```diff
@@ -76,14 +77,14 @@ if [[ -n "${LAUNCH_OPTIONS_JSON_STR}" ]]; then
...
-  runner_flags+=(--launch_options_json_path="${LAUNCH_OPTIONS_JSON_PATH}")
+  runner_flags+=("--launch_options_json_path=${LAUNCH_OPTIONS_JSON_PATH}")
 fi
 
 cmd=("%(testrunner_binary)s"
   "${runner_flags[@]}"
   simulator_test
-  --device_type="%(device_type)s"
-  --os_version="%(os_version)s"
+  "--device_type=%(device_type)s"
```
rahul-malik: Do we need quotes around the values for device type and os version?

keith: Since the whole argument will be passed as `'--device_type=iPhone X'` it should be fine with spaces

apple/testing/default_runner/ios_test_runner.template.sh
```diff
@@ -18,33 +18,34 @@ set -e
 
 basename_without_extension() {
   local full_path="$1"
-  local filename=$(basename "$full_path")
+  local filename
+  filename=$(basename "$full_path")
```
sergiocampama: what's the reason for this difference with the previous variable?

keith: Shellcheck validates that you don't define and assign to a variable if the assignment is a command that could also fail https://github.com/koalaman/shellcheck/wiki/SC2155

keith: (behavior is identical)

sergiocampama:
This change has been imported, will be exported soon

PR #289, opened 2018-12-13
Update

sergiocampama:


apple/internal/apple_framework_import.bzl
```diff
@@ -84,7 +61,7 @@ def _apple_framework_import_impl(ctx):
         attr = "framework_imports",
     )
 
-    _validate_single_framework(framework_groups.keys())
+    # TODO(b/120920467): Add validation to ensure only a single framework is being imported.
```
keith: is this TODO saying the fix would be to revert this change?

sergiocampama: yes. For some reason, `objc_framework` and `objc_bundle` allowed more than one imported framework/bundle, and we'd like to clean that up, but some clients are depending on this right now.

PR #295, opened 2018-12-21
Fix bazel 0.21.0 incompatible changes

keith:


apple/testing/apple_test_rules.bzl
```diff
@@ -191,7 +191,7 @@ def _collect_files(rule_attr, attr_name):
         return depset()
 
     attr_val_as_list = attr_val if type(attr_val) == type([]) else [attr_val]
```
sergiocampama: I think this would be better to be refactored as:
```
return depset(transitive = [f.files for f in attr_val_as_list])
```

and remove the `files = ...` line. that way we won't expand the depsets, and should be a bit faster. (I wrote this, I'm a little ashamed 😛 )

keith: done!

keith:
fixed and rebased!

sergiocampama:
testing this change internally

PR #296, opened 2018-12-26
Update README.md

shoumikhin:


README.md
```diff
@@ -42,13 +42,25 @@ load(
 
 apple_rules_dependencies()
 
+git_repository(
+    name = "build_bazel_rules_swift",
+    remote = "https://github.com/bazelbuild/rules_swift.git",
+    tag = "0.5.0",
+)
```
sergiocampama: these are already added in the `apple_rules_dependencies` method, please remove

README.md
```diff
@@ -42,13 +42,25 @@ load(
...
+
 load(
     "@build_bazel_rules_swift//swift:repositories.bzl",
     "swift_rules_dependencies",
 )
 
 swift_rules_dependencies()
 
+git_repository(
```
sergiocampama: same here

README.md
shoumikhin: Do you say there's actually no need to load and call both `swift_rules_dependencies` and `apple_support_dependencies`, because those are handled with `apple_rules_dependencies` above?

sergiocampama: no, you need to call `swift_rules_dependencies` and `apple_rules_dependencies`, so those are ok... only the new `git_repository` should be removed

shoumikhin: Interesting... I remember there was an error saying a git repo such and such not found, so I had to load them explicitly. Now that's not the case, pulled.

README.md
```diff
@@ -32,7 +32,7 @@ load("@bazel_tools//tools/build_defs/repo:git.bzl", "git_repository")
 git_repository(
     name = "build_bazel_rules_apple",
     remote = "https://github.com/bazelbuild/rules_apple.git",
-    tag = "0.10.0",
+    tag = "0.11.0",
```
sergiocampama: one last nit, the latest version now is 0.11.1 :)

Issue #308, opened 2019-01-22
codesign error


xinzhengzhang:
https://github.com/bazelbuild/rules_apple/blob/7f65be56ce91f9accde6247f392f2c5c2f54c70f/tools/codesigningtool/codesigningtool.py#L85

tool/codesigningtool.runfiles/build_bazel_rules_apple/tools/codesigningtool/codesigningtool.py", line 84, in _find_codesign_identities
    "security", "find-identity", "-v", "-p", "codesigning",
AttributeError: 'tuple' object has no attribute 'decode'

PR #309, opened 2019-01-22
Fix AttributeError: 'tuple' object has no attribute 'decode'

xinzhengzhang:
#308


sergiocampama:
This change has been merged internally and will be exported soon. Thanks!

PR #311, opened 2019-01-28
Change simctl command to pass full runtime

keith:
As of Xcode 12.2 beta 1 simctl no longer accepts the short version for
creating a simulator. Instead we need to pass the full runtime.


keith:
This failure looks like a flake

keith:
It sounds like this is the safest backwards compatible option, but at least for CoreSimulator bundled with Xcode 10.2, `iOS12.2` also works

PR #314, opened 2019-01-29
Use output instead of tuple.

chenxiao0228:
Fix runtime error "AttributeError: 'tuple' object has no attribute 'decode'


sergiocampama:
sorry, this was already fixed in #308 but I hadn't merged the push

PR #315, opened 2019-01-30
Fix resource documentation grouping

keith:


keith:
|before|after|
|-|-|
|image|image|

PR #318, opened 2019-01-31
Add bundle type for iOS

keith:
iOS supports bundle types for resource bundles, but rules_apple
currently doesn't build any with this type. In theory
apple_resource_bundle could use this type instead of the type it
inherits from the final rule that includes the bundle. Another possible
use case is a new rule that pre-compiled resource bundles for others to
include using apple_bundle_import.

### DIFF
--- a/apple/internal/rule_support.bzl
+++ b/apple/internal/rule_support.bzl
@@ -277,6 +277,16 @@ _RULE_TYPE_DESCRIPTORS = {
             requires_provisioning_profile = False,
             skip_signing = True,
         ),
+        apple_product_type.bundle: _describe_rule_type(
+            allowed_device_families = ["iphone", "ipad"],
+            bundle_extension = ".bundle",
+            has_infoplist = True,
+            product_type = apple_product_type.bundle,
+            requires_bundle_id = True,
+            requires_provisioning_profile = False,
+            requires_signing_for_device = False,
+            skip_signing = True,
+        ),
         # ios_ui_test
         apple_product_type.ui_test_bundle: _describe_rule_type(
             allowed_device_families = ["iphone", "ipad"],


thomasvl:
I'm not sure I'm following the value of adding this if there isn't something that makes use of it.  That usually leads to things not being setup/maintained because they aren't used, and so if/when someone does, they need correcting anyways.

keith:
Really I just submitted this because we need it because of this issue https://github.com/bazelbuild/rules_apple/issues/319

But I understand if that's something rules_apple doesn't want to support, it just means we'll have to apply this patch for the foreseeable future in our project.

keith:
@thomasvl @sergiocampama any thoughts on this? I think that since this is something rules_apple could use in the future and since it should be a very low maintenance burden and since we have a valid use case for it today, it would be nice to land even thought it will be unused in rules_apple itself for a while.

keith:
rebased

sergiocampama:
Sorry for the delay.

How are you using this rule descriptor? Where does the rule_descriptor fetching fail in the resource rule that you've created? Perhaps we can find another path to avoid requesting the rule descriptor for your use case, and you can configure the logic from within your rule.

The fact that this product type doesn't map to a resource bundle on Xcode is what's keeping me from supporting this, since it sounds more like a workaround than a proper solution. The fact that it doesn't have an associated rule is also worrying, and on our side we're thinking about loadable bundles (i.e plugins) for iOS too () which would need this product type as well.

keith:
See the example rule here for how this is used https://github.com/bazelbuild/rules_apple/issues/319

This product type is definitely supported by Xcode so I'm not sure what you mean there?

sergiocampama:
What I meant by that is that there is no resource bundle product type. There's no way to create resource bundles at all through the Xcode interface. The bundle product type does exist but for a different purpose, which is a macOS Bundle (which you can create through the Xcode New Target UI)

sergiocampama:
looking at the sample rule implementation, we could extract the rule_descriptor usage from the resources partial (and dependencies) so that you don't need this descriptor at all. I think that would be better IMO. (It might be that it's cleaner to extract all rule descriptor uses from within the partials, and configure those bits in the partial arguments instead.)

I can see 3 places where it's being used for resources:

1. [pkginfo](https://github.com/bazelbuild/rules_apple/blob/master/apple/internal/partials/resources.bzl#L82)
2. [additional plist values](https://github.com/bazelbuild/rules_apple/blob/master/apple/internal/resource_actions/plist.bzl#L283)
3. [binary plist output](https://github.com/bazelbuild/rules_apple/blob/master/apple/internal/resource_actions/plist.bzl#L322)

I think that they should be fairly simple configs to extract into the resource partial argument list and propagate them down.

keith:
I started looking at implementing this, one thing that stuck out as being a bit strange is that callers you mention are the only consumers of some of these fields from rule_support. Do you think in that case we should remove them from rule_support, meaning they only exist in the implementation of the rules that calls the partials, or do you think we should access rule_support in the rule implementations, and pass them through that way (possibly as kwargs)

keith:
I've submitted a PR removing them from rule_support entirely here https://github.com/bazelbuild/rules_apple/pull/368

PR #322, opened 2019-01-31
Fix resources being re-processed from pre-built bundles

keith:
apple_static_framework_import supports frameworks that have bundle
directories inside them. In this case these are pre-built bundles that
were shipped with the framework. Previously we were re-processing those
files like they were normal asset files. Now we propagate them directly
without processing.

### DIFF
--- a/test/ios_application_test.sh
+++ b/test/ios_application_test.sh
@@ -149,6 +149,11 @@ EOF
 Dummy resource
 EOF
 
+  mkdir -p app/fmwk.framework/fmwk.bundle
+  cat > app/fmwk.framework/fmwk.bundle/Some.plist <<EOF
+Dummy plist
+EOF
+
   mkdir -p app/fmwk.framework/Headers
   cat > app/fmwk.framework/Headers/fmwk.h <<EOF
 This shouldn't get included
--- a/test/ios_application_test.sh
+++ b/test/ios_application_test.sh
@@ -151,7 +151,14 @@ EOF
 
   mkdir -p app/fmwk.framework/fmwk.bundle
   cat > app/fmwk.framework/fmwk.bundle/Some.plist <<EOF
-Dummy plist
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+  <dict>
+    <key>Foo</key>
+    <string>Bar</string>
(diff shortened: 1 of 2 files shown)


keith:
Added!

keith:
It looks like my test has an issue on device but not for the simulator variant, debugging

keith:
Accidentally dropped a change because I had commented out some tests to iterate faster. Hopefully this works now

keith:
Ok it looks like now the only test failure is the HEAD related one

test/ios_application_test.sh
```diff
@@ -147,6 +147,11 @@ EOF
 
   cat > app/fmwk.framework/resource.txt <<EOF
 Dummy resource
+EOF
+
+  mkdir -p app/fmwk.framework/fmwk.bundle
+  cat > app/fmwk.framework/fmwk.bundle/Some.plist <<EOF
+Dummy plist
```
thomasvl:
```
<?xml version="1.0" encoding="UTF-8"?>
<plist version="1.0">
<dict>
	<key>Foo</key>
	<string>Bar</string>
</dict>
</plist>
```

because if it was processed, I'll bet things would error since it isn't a valid plist?

keith: The tests run fine for me, I was a bit surprised too. I can replace it with this anyways though

thomasvl: Without your fix, does it work?  i.e. - if the file is processed by accident, what happens then?  If it still passes, then either the test isn't testing what we think it is, or a bogus file slips thru unprocessed meaning the test isn't ensuring it processed.

keith: Yep it fails correctly without it

keith: It looks like this is expected, plutil happily converts a file like this:

```
% echo "hi" > foo.txt
% plutil -convert binary1 foo.txt
% file foo.txt
foo.txt: Apple binary property list
% cat foo.txt
bplist00Rhi
```

I'll still make the change to make it make more sense

keith: Actually there is this same case above for the framework's Info.plist, so should we leave this one or change both?

sergiocampama: just keep this one, leave the other one as is

sergiocampama: what I meant was to not modify the other tests, and make the changes that @thomasvl mentioned for this test

keith: Done

thomasvl:
FYI - we hit some failures within google with this change.  We're looking into the issues.

PR #328, opened 2019-02-07
Validate beta-reports-active in entitlements

keith:
This key is validated against your profile by iTunes Connect, so we
should too. This is a case where having it in the profile at all
requires it's in your entitlements.

### DIFF
--- a/tools/plisttool/plisttool.py
+++ b/tools/plisttool/plisttool.py
@@ -297,6 +297,11 @@ ENTITLEMENTS_APS_ENVIRONMENT_MISMATCH = (
     'match the value in the provisioning profile ("%s").'
 )
 
+ENTITLEMENTS_BETA_REPORTS_ACTIVE_MISMATCH = (
+    'In target "%s"; the entitlements "beta-reports-active" ("%s") did not '
+    'match the value in the provisioning profile ("%s").'
+)
+
 ENTITLEMENTS_HAS_GROUP_ENTRY_PROFILE_DOES_NOT = (
     'Target "%s" uses entitlements "%s" value of "%s", but the profile does '
     'not support it (["%s"]).'
@@ -1175,6 +1180,15 @@ class EntitlementsTask(PlistToolTask):
               ENTITLEMENTS_APS_ENVIRONMENT_MISMATCH % (
                 self.target, aps_environment, profile_aps_environment),
               **report_extras)
+
+    beta_reports_active = entitlements.get('beta-reports-active')
+    if beta_reports_active is not None or profile_entitlements:
+      profile_key = (profile_entitlements or {}).get('beta-reports-active')
+      if profile_key != beta_reports_active:
+        self._report(
+            ENTITLEMENTS_BETA_REPORTS_ACTIVE_MISMATCH % (
+                self.target, beta_reports_active, profile_key),
+            **report_extras)
 
     # keychain-access-groups
     self._check_entitlements_array(
--- a/tools/plisttool/plisttool.py
+++ b/tools/plisttool/plisttool.py
@@ -1187,7 +1187,7 @@ class EntitlementsTask(PlistToolTask):
               **report_extras)
 
     # If beta-reports-active is in either the profile or the entitlements file
-    # it must be in both or the upload will get rejected by Applejk
+    # it must be in both or the upload will get rejected by Apple
     beta_reports_active = entitlements.get('beta-reports-active')
     profile_key = (profile_entitlements or {}).get('beta-reports-active')
     if beta_reports_active is not None and profile_key != beta_reports_active:
--- a/tools/plisttool/plisttool.py
+++ b/tools/plisttool/plisttool.py
@@ -302,11 +302,6 @@ ENTITLEMENTS_BETA_REPORTS_ACTIVE_MISMATCH = (
     'match the value in the provisioning profile ("%s").'
 )
 
-ENTITLEMENTS_BETA_REPORTS_ACTIVE_MISSING_ENTITLEMENTS = (
-    'In target "%s"; the profile has "beta-reports-active" ("%s") so it must '
-    'also exist in the entitlements file.'
-)
-
 ENTITLEMENTS_BETA_REPORTS_ACTIVE_MISSING_PROFILE = (
     'In target "%s"; the entitlements file has "beta-reports-active" ("%s") '
     'but it does not exist in the profile.'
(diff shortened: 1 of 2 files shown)


tools/plisttool/plisttool.py
```diff
@@ -1176,6 +1181,15 @@ def _validate_entitlements_against_profile(
                 self.target, aps_environment, profile_aps_environment),
               **report_extras)
 
+    beta_reports_active = entitlements.get('beta-reports-active')
+    if beta_reports_active is not None or profile_entitlements:
```
kastiglione: Don't need this `or profile_entitlements`, since the next line handles the case where it's missing.

keith: This is for the case if `beta_reports_active is None`

tools/plisttool/plisttool.py
```diff
@@ -1176,6 +1181,14 @@ def _validate_entitlements_against_profile(
                 self.target, aps_environment, profile_aps_environment),
               **report_extras)
 
+    beta_reports_active = entitlements.get('beta-reports-active')
+    profile_key = (profile_entitlements or {}).get('beta-reports-active')
+    if profile_key != beta_reports_active:
```
thomasvl: Your comment in the cl description:

> This is a case where having it in the profile at all requires it's in your entitlements.

Seems like it should be recorded here at the checks; otherwise, the fact that this is different than the APS one above in handing it missing can look like a bug.

To that point, I think we should generate a specific message for when it is missing rather than saying _None_ doesn't match _True_, as other developers might not realize it works differently and has to be there.

thomasvl: @keith were you still interested in finishing this out?

keith: Yea sorry, haven't had a chance to get back to it

keith: Updated, let me know if that's not what you were thinking

thomasvl: Thanks

tools/plisttool/plisttool.py
```diff
@@ -1176,6 +1191,21 @@ def _validate_entitlements_against_profile(
...
+      error_msg = ENTITLEMENTS_BETA_REPORTS_ACTIVE_MISMATCH % (
+        self.target, beta_reports_active, profile_key)
+      if profile_key is None:
+        error_msg = ENTITLEMENTS_BETA_REPORTS_ACTIVE_MISSING_PROFILE % (
+          self.target, beta_reports_active)
+      elif beta_reports_active is None:
+        error_msg = ENTITLEMENTS_BETA_REPORTS_ACTIVE_MISSING_ENTITLEMENTS % (
```
thomasvl: @keith how sure are you about this case?  I'm finding a few of our apps are falling into this case, but it would appear they were otherwise working.  Are you sure one can't simply leave out the key if you don't need the entitlement?

sergiocampama: IMO, if the entitlements don't have a capability enabled in the profile, it shouldn't be an error. This just signals that the app is not fully using all the things it's allowed, but that's ok. The other way is ok (i.e. if it's in the entitlements, it must be in the profile as well)

keith: I made this an error intentionally because in our case we got rejected when the app entitlements didn't have `beta-reports-active` but the provisioning profile did.

You're saying that the case you found the provisioning profile has the key but the entitlements file does not? Note this only matters if you were uploading the build to Apple though, so if you share an entitlements file between your debug and app store builds, this might not be the desired behavior.

sergiocampama: The problem that we see here is that we have apps building with the app store profile which don't necessarily have a TestFlight version, so requiring them to add the entitlement seems wrong. the TestFlight entitlement is added by default to all app store profiles.

keith: Are you saying that if you submit them as a release build without ever submitting them to testflight you avoid the need for this entitlement? That's an interesting case. Our production profile also has `beta-reports-active` set to `true`, but maybe iTC doesn't validate it when you're distributing only

sergiocampama: I haven't seen these errors actually, so it's unclear, but we do have build tests for apps that are in the App Store which do not have this entitlement, so it makes me think that it's not validated for App Store releases.

keith: This was the content of the email:

>>>Missing beta entitlement - Your app does not include the beta-reports-active entitlement. If you intend to distribute this build via TestFlight for beta testing, please re-build this app with an App Store Distribution provisioning profile. Do not use ad-hoc profiles.

keith: But yea it makes sense that it wouldn't be for release builds. Any ideas on how we can validate this just for the testflight case? Assuming you think that it's unreasonable to require people to add this to their entitlements

thomasvl: > Any ideas on how we can validate this just for the testflight case?

We've talked about (might not have it as an issue though), _standard_ `config_setting` (or something) that the rules can key off of to know if it the build is going to be _dev_, _testflight_, or _appstore_.  Maybe also needing the idea of _enterprise_.  But then decisions could be made off of that rather than trying to get based on _opt_ vs. _dbg_, etc.  So this seems like potentially another things to justify figuring out a good way to do this.

The part that always give me some pauses, is I'd hate for folks to have to pass yet another command line flag to all bazel invokes.  But maybe with some of the bazel work around transitions/configurations/etc., this could be captured in BUILD files so it doesn't become a flag to pass every time.

keith: Yea that would make sense. I get that it's a pain though, although IMO the only way you could possibly use bazel is with a very comprehensive bazelrc file. We have ~150 LOC in that file, and pass just a single option `--config=[debug|qa|alpha|...]` that hides all of that kind of thing. But I get that our setup might also be unique

thomasvl: Ok, so to move this forward, lets make the validation more like the others (if in entitlements, as to be in profile (and match), but missing from entitlements is ok).  I've opened #342 to capture the desire/need for having something to capture where the build is headed enable more strict checks in the future.

keith: Done!

keith:
rebased

tools/plisttool/plisttool.py
```diff
@@ -1176,6 +1186,18 @@ def _validate_entitlements_against_profile(
                 self.target, aps_environment, profile_aps_environment),
               **report_extras)
 
+    # If beta-reports-active is in either the profile or the entitlements file
+    # it must be in both or the upload will get rejected by Applejk
```
thomasvl: "Applejk"

typo

keith: thanks fixed

thomasvl:
Landed internally, @sergiocampama I think is pushing a bunch of things back out shortly.

keith:
Thanks for all your help!

PR #331, opened 2019-02-12
Provide xctestrunner automatically

keith:
I'm not sure what the background of requiring this to be manual is, but this seems to work fine for our project


keith:
rebased

PR #333, opened 2019-02-21
Add feature for disabling SwiftSupport

keith:
The SwiftSupport directory in a final ipa is only necessary for builds
shipped to Apple. Any other device or enterprise builds don't need it.
For us having this extra directory greatly bloats the ipa size. This
gives an feature to control disabling it in that case.

### DIFF
--- a/apple/internal/ios_rules.bzl
+++ b/apple/internal/ios_rules.bzl
@@ -51,6 +51,10 @@ load(
     "stub_support",
 )
 load(
+    "@build_bazel_rules_apple//apple/internal/utils:defines.bzl",
+    "defines",
+)
+load(
     "@build_bazel_rules_apple//apple:providers.bzl",
     "IosApplicationBundleInfo",
     "IosExtensionBundleInfo",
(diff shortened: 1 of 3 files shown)


keith:
We don't have to go this route, we might want to do something smarter as implied in the TODOs. I'm not entirely sure how we would do the "app store vs enterprise" thing though so this _might_ end up being the best option

keith:
This is an alternative or additive solution to https://github.com/bazelbuild/rules_apple/pull/334

apple/internal/ios_rules.bzl
```diff
@@ -129,7 +135,7 @@ def _ios_application_impl(ctx):
             bundle_dylibs = True,
             # TODO(kaipi): Revisit if we can add this only for AppStore optimized builds, or at
             # least only for device builds.
-            package_swift_support = True,
+            package_swift_support = package_swift_support,
```
sergiocampama: instead of having this here in every rule that packages SwiftSupport, can you instead do this inside the swift dylibs partial?

keith:
Closing for https://github.com/bazelbuild/rules_apple/pull/336

PR #334, opened 2019-02-21
Only package SwiftSupport for device builds

keith:
For simulator builds the SwiftSupport directory isn't necessary in the
final ipa. It really isn't for all device builds either, only device
builds you will send to Apple, but we can't differentiate on that here.


keith:
This is an alternative or additive solution to https://github.com/bazelbuild/rules_apple/pull/333

keith:
Closing for https://github.com/bazelbuild/rules_apple/pull/336

Issue #304, opened 2019-01-14
[apple/testing/apple_test_rules.bzl] "env" value not passed to hosted ios_unit_test NSProcessInfo


romoore:
## Summary
The `env` values of `ios_unit_test` are not passed to `NSProcessInfo.processInfo.environment` in hosted unit tests (non-logic tests that have a custom `test_host` property set).  Using `--test_env` works as expected.

## Context/Detail

I'm trying to integrate https://github.com/material-foundation/ios-snapshot-test-case as a snapshot testing framework into our test targets.

I've attempted to [define the "env" variable "FB_REFERENCE_IMAGE_DIR" in a skylark rule for snapshot tests](https://github.com/material-components/material-components-ios/blob/f5c92fd1378d1410cb074e4a46ffdd51d8afcc95/material_components_ios.bzl#L186) (we'll be creating several dozen of these). 

```
def mdc_snapshot_test(
    name,
    deps = [],
    minimum_os_version = SNAPSHOT_IOS_MINIMUM_OS,
    visibility = ["//visibility:private"],
    size = "medium",
    **kwargs):
    
  test = "%sTest" % name
  app = "%sApp" % name

  """Declare an MDC ios_unit_test for snapshot tests."""
  ios_unit_test(
      name = test,
      deps = ["//components/private/Snapshot/TestHost"] + deps,
      minimum_os_version = minimum_os_version,
      runner = SNAPSHOT_IOS_RUNNER_TARGET,
      test_host = ":%s" % app,
      visibility = visibility,
      env = {
          "FB_REFERENCE_IMAGE_DIR":"snapshot_test_goldens/goldens",
      },
      size = size,
      **kwargs)
```

When I inspect the environment runtime in the code under test using `[NSProcessInfo processInfo].environment`, I don't see the variable available.  When I pass `--test_env="FB_REFERENCE_IMAGE_DIR=$snapshot_dir"`, I see the value in the environment values and my code behaves as expected.  

Am I doing something wrong, or [am I misreading the documentation for this variable](https://github.com/bazelbuild/rules_apple/blob/24abca60f5ec59c5336b62a61714234c23a8d17c/doc/rules-ios.md)?

keith:
This should be fixed by https://github.com/bazelbuild/rules_apple/pull/335 if you have a chance please try it and let me know if it fixes your use case!

romoore:
[...]

PR #335, opened 2019-02-21
Fix conflict test environment issues

keith:
Previously the test_env was being replaced in the test runner template
by the runner itself. At this point any value set for the `env` attr in
the test rule was not realized. Now we instead do the replacement in the
test rule itself, including the environment added by the runner.

Fixes https://github.com/bazelbuild/rules_apple/issues/304

### DIFF
--- a/apple/internal/testing/apple_test_rules.bzl
+++ b/apple/internal/testing/apple_test_rules.bzl
@@ -100,8 +100,8 @@ added to the test rules runfiles.
 Dictionary that represents the specific hardware
 requirements for this test.
 """,
-        "test_environment": """
-Dictionary with the environment variables required for the test.
+        "execution_environment": """
+Dictionary with the environment variables the test runner requires
 """,
         "test_runner_template": """
 Template file that contains the specific mechanism with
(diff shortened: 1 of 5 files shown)


keith:
I think I changed it how you asked, let me know if not

keith:
Hoping the tests are green with this push

keith:
All green! Ready for re-review @sergiocampama

keith:
Rebased and updated the tvOS runner to match

test/testdata/rules/dummy_test_runner.bzl
```diff
@@ -31,6 +31,7 @@ def _dummy_test_runner_impl(ctx):
             test_runner_template = ctx.outputs.test_runner_template,
             execution_requirements = {},
             test_environment = {},
```
sergiocampama: remove?

apple/internal/testing/apple_test_rules.bzl
```diff
@@ -102,6 +102,9 @@ requirements for this test.
 """,
         "test_environment": """
```
sergiocampama: same here, should we just remove this field since it's not being used anymore?

keith: Originally I think you recommended leaving this for backwards compatibility but I may have misunderstood. I'll remove this

sergiocampama: now that you mention, that rings a bell, but still, I'll migrate internal runners and post this in release notes so that other external clients know to update their runners, I realized I don't want to track a migration path for this

keith:
Updated!

PR #336, opened 2019-02-22
Stop always including SwiftSupport

keith:
The only case where you need the SwiftSupport in the final archive is if
you're shipping the build to Apple. Since we can't know if the build
being created is being shipped to Apple, we can instead only bundle it
1) if it's a device build and 2) if you don't opt out with the new
`apple.package_swift_support` feature.

### DIFF
--- a/apple/internal/ios_rules.bzl
+++ b/apple/internal/ios_rules.bzl
@@ -131,9 +131,7 @@ def _ios_application_impl(ctx):
             binary_artifact = binary_artifact,
             dependency_targets = embeddable_targets,
             bundle_dylibs = True,
-            # TODO(kaipi): Revisit if we can add this only for AppStore optimized builds, or at
-            # least only for device builds.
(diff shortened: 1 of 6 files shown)


keith:
Hitting this bazel flake on CI:

```
ERROR: /private/var/tmp/_bazel_travis/2b80b7ac301f661b0fa17c1c94402319/external/local_config_cc/BUILD:55:5: in apple_cc_toolchain rule @local_config_cc//:cc-compiler-tvos_x86_64: Xcode version must be specified to use an Apple CROSSTOOL. If your Xcode version has changed recently, try: "bazel clean --expunge" to re-run Xcode configuration
```

Retrying

keith:
New flake:

```
Could not determine Xcode version at all. This likely means Xcode isn't available; if you think this is a mistake, please file an issue.
```

keith:
rebased

PR #337, opened 2019-02-22
Add feature to force ipa compression

keith:
For some use cases you may want to compress the final ipa file even if
you're not building an optimized build. For example for enterprise
distribution where you want the builds to be faster.

### DIFF
--- a/apple/internal/codesigning_support.bzl
+++ b/apple/internal/codesigning_support.bzl
@@ -303,9 +303,10 @@ def _post_process_and_sign_archive_action(
         processing_tools.append(ipa_post_processor)
         ipa_post_processor_path = ipa_post_processor.path
 
-    # Only compress the IPA for optimized (release) builds. For debug builds,
-    # zip without compression, which will speed up the build.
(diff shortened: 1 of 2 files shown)


thomasvl:
I don't follow the comment:

> For some use cases you may want to compress the final ipa file even if
> you're not building an optimized build. For example for enterprise
> distribution where you want the builds to be faster.

Since it takes time to compress, how is it faster?

And do you really use _dbg_ for enterprise?  Just trying to understand, there's a few cases were we've sorta assumed it would always be a _opt_ build if it was getting any distribution (AppStore, TestFlight, Enterprise).

keith:
There's a conversation about this in the slack but our use case is that ew build enterprise builds on every push to every PR that can be used for QA. The extra 10 seconds zipping in this case is much preferred to the extra 300+ mbs of the download for a few reasons 1) we're not currently bound by CPU time for CI 2) the download times from where we store the enterprise builds aren't great. We use dbg for these builds because they're mostly to validate the build on each PR

keith:
rebased

PR #339, opened 2019-02-26
Lock apple_support to commit

keith:
Bazel warns about this unless you lock this to a specific version:

```
DEBUG: Rule 'build_bazel_apple_support' indicated that a canonical reproducible form can be obtained by modifying arguments commit = "77c9d5371949352c00bc74d4755c93e00199172b", shallow_since = "1547834515 -0500" and dropping ["tag"]
```

More info: https://github.com/bazelbuild/bazel/issues/7389

### DIFF
--- a/apple/repositories.bzl
+++ b/apple/repositories.bzl
@@ -107,7 +107,8 @@ def apple_rules_dependencies(ignore_version_differences = False):
         git_repository,
         name = "build_bazel_apple_support",
         remote = "https://github.com/bazelbuild/apple_support.git",
-        tag = "0.5.0",
+        commit = "dabf718975760b4d45bd7db3e18bc12e7b4ef485",
+        shallow_since = "1551297696 -0500",
         ignore_version_differences = ignore_version_differences,
     )
 


keith:
Alternatively we could use http_archive with the tar of the sources from GitHub

sergiocampama:
I'm not sure why that DEBUG statement is appearing, do you know where it's from? given that it doesn't say the same for skylib or rules_swift. I'd like to understand what direction the bazel team is taking before making this change, since it goes against easier versioning

keith:
Yea I thought about that before submitting this. In our case those are locked by other dependencies so the `_maybe` isn't hitting. I can update those as well.

keith:
rebased

sergiocampama:
It looks like we're going to switch to the `http_archive` model, which would include the `sha256` of the archives. We're waiting for the skylib owners to come back from vacation to coordinate the rollout

keith:
Submitted a PR for that here: https://github.com/bazelbuild/rules_apple/pull/361

PR #340, opened 2019-02-26
Migrate to new code coverage provider

keith:
This fixes an incompatible change detailed here: https://github.com/bazelbuild/bazel/issues/7347

This relies on bazel 0.23.0


sergiocampama:
we also already have this change internally, will push out later today

PR #341, opened 2019-02-26
Move to provider list for objc providers

keith:


sergiocampama:
thanks! we already made this change internally, will push out later today

PR #353, opened 2019-03-14
Allow specifying custom umbrella header name in ios_static_framework

thii:
Previously, if your framework target has the same name with one of its
public headers, Bazel would yield an error like 'Multiple files would be
placed at "Buttons.framework/Headers/Buttons.h" in the bundle,
which is not allowed'. This patch adds a new attribute to the
`ios_static_framework` rule that allow specifying a custom umbrella header
name for the bundled framework.

Examples:

    objc_library(
        name = "ButtonsLib",
        srcs = [
            "Buttons.m",
        ],
        hdrs = [
            "Buttons.h",
        ],
        module_name = "Buttons",
    )

    ios_static_framework(
        name = "ButtonsStaticFramework",
        bundle_name = "Buttons",
        hdrs = [
            "Buttons.h",
        ],
        umbrella_header_name = "Buttons-umbrella.h",
        minimum_os_version = "8.0",
        deps = [":ButtonsLib"],
    )

### DIFF
--- a/apple/internal/partials/static_framework_header_modulemap.bzl
+++ b/apple/internal/partials/static_framework_header_modulemap.bzl
@@ -121,9 +121,10 @@ def _static_framework_header_modulemap_partial_impl(ctx, hdrs, binary_objc_provi
 
     bundle_files = []
 
-    umbrella_header_name = None
+    umbrella_header_name = ctx.attr.umbrella_header_name
     if hdrs:
-        umbrella_header_name = "{}.h".format(bundle_name)
+        if not umbrella_header_name:
+            umbrella_header_name = "{}.h".format(bundle_name)
         umbrella_header_file = intermediates.file(ctx.actions, ctx.label.name, umbrella_header_name)
         _create_umbrella_header(
             ctx.actions,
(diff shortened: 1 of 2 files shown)


sergiocampama:
Doesn't this go against the convention for umbrella headers in distributable frameworks, where the umbrella header has the same name as the bundle name? 

Shouldn't the proper solution for the Buttons example be to rename the `Buttons.h` header to `ButtonsLib.h` or something along those lines?

thii:
> Shouldn't the proper solution for the Buttons example be to rename the Buttons.h header to ButtonsLib.h or something along those lines?

This is one of the solutions. In our use case, we were building an external library and we wanted to avoid making custom patches as much as we can. In addition to renaming the header file, you would have to also patch the import statements in `.m` files.

On a second thought, for libraries that only have one public header, and that header file has the same name with the framework, we probably don't need an umbrella header. I think a better solution would be allowing to disable the generation of an umbrella header when create a framework for distribution.

thii:
Submitted a patch about not to bundle the umbrella header when appropriate (https://github.com/bazelbuild/rules_apple/pull/359). Please take a look.

sergiocampama:
is this pr still open or you wanted to close it?

thii:
Closed in favor of https://github.com/bazelbuild/rules_apple/pull/359.

PR #359, opened 2019-03-15
Don't bundle the umbrella header if there is only one public header which already has the same name

thii:


sergiocampama:
can you also add documentation about the umbrella header into the `ios_static_framework` docs? Currently there is no documentation about how umbrella headers are processed, please add a section on how it's processed and what happens when there is a single header with the name of the bundle.

thii:
Did you mean the inline documentation of `ios_static_framework` rule? Because I think the `rules-ios.md` file is autogenerated.

thomasvl:
> Did you mean the inline documentation of `ios_static_framework` rule? Because I think the `rules-ios.md` file is autogenerated.

They aren't generated.  That's on the list, but stardoc can't handing everything yet.

thii:
Updated. Please take a look.

apple/internal/partials/static_framework_header_modulemap.bzl
```diff
@@ -131,9 +131,16 @@ def _static_framework_header_modulemap_partial_impl(ctx, hdrs, binary_objc_provi
             sorted(hdrs),
         )
 
-        bundle_files.append(
-            (processor.location.bundle, "Headers", depset(hdrs + [umbrella_header_file])),
-        )
+        # Don't bundle the umbrella header if there is only one public header
+        # which has the same name
+        if len(hdrs) == 1 and hdrs[0].basename == umbrella_header_name:
```
keith: can we do this comparison case insensitively to solve https://github.com/bazelbuild/rules_apple/issues/408 ? I'm not sure if that would cause any other issues or not

keith:
@sergiocampama can you re-review?

apple/internal/partials/static_framework_header_modulemap.bzl
```diff
@@ -131,9 +131,16 @@ def _static_framework_header_modulemap_partial_impl(ctx, hdrs, binary_objc_provi
             sorted(hdrs),
         )
 
-        bundle_files.append(
-            (processor.location.bundle, "Headers", depset(hdrs + [umbrella_header_file])),
-        )
+        # Don't bundle the umbrella header if there is only one public header
+        # which has the same name (case-insensitive)
+        if len(hdrs) == 1 and hdrs[0].basename.lower() == umbrella_header_name.lower():
```
keith: what's the need to limit this to only take effect if there's a single header? isn't this logic still valid if just one of the headers has the same name in a glob of headers?

keith: If it is ok to check for any number of headers this works for that:

```
if any([x.basename.lower() == umbrella_header_name.lower() for x in hdrs]):
```

thii: > what's the need to limit this to only take effect if there's a single header? isn't this logic still valid if just one of the headers has the same name in a glob of headers?

Yeah if that one header is the umbrella header, which can't be guaranteed.
 
This is a specific change to the case where the framework only produces one header. I agree that a more complete solution would be to allow users to specify a custom name for the umbrella header.

sergiocampama: the `bundle_name`  attribute can sorta be used for this, as it will change the output bundle and umbrella header name without changing the target name. Given the convention of importing `#import <FRAMEWORK_NAME/FRAMEWORK_NAME.h>` having an option to provide the name of the umbrella header that is different from the `.framework` name might seem counterintuitive for users of that framework.

it's also weird that on the CgRPC case, where cgrpc.h exists, users would `#import <CgRPC/CgRPC.h>` but in reality the header would be named `cgrpc.h`, which in a way abuses the fact that macOS is case insensitive.

keith: FWIW clang warns if you import the header with the wrong case last time I checked. But I'm definitely thinking about this more from the modules side where I'd use `@import CgRPC` there instead, but I understand the concern.

sergiocampama: we had a small chat with @keith and we arrived at this flow for umbrella headers:

in summary what we arrived at is:

1. check for umbrella_header attribute (which will be added by @keith in a later PR), if it is there, use it.
2. check if there's a single header with the same name _and case_, if there use it (what this PR implements).
3. generate an umbrella header with the same name as the bundle and import all the other headers (what is already implemented)

what do you think @thii? does this work for you? this means reverting the case sensitivity check that was added to this PR

keith: Sorry about misleading you @thii!

thii: Sounds good to me. I’ll revert the case sensitivity check.

keith: Here's the follow up PR with the new attribute https://github.com/bazelbuild/rules_apple/pull/437

PR #361, opened 2019-03-18
Switch to http_archive rule for dependencies

keith:
This stops bazel's log about downloading a dependency that's not locked
with a shallow_since field when using git_repository

### DIFF
--- a/apple/repositories.bzl
+++ b/apple/repositories.bzl
@@ -14,8 +14,7 @@
 
 """Definitions for handling Bazel repositories used by the Apple rules."""
 
-load("@bazel_tools//tools/build_defs/repo:git.bzl", "git_repository")
-load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_file")
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive", "http_file")
 
 def _colorize(text, color):
     """Applies ANSI color codes around the given text."""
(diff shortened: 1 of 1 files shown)


keith:
Theoretically the shas of these tar files can change with backend github changes, realistically it's pretty unlikely and rare, but we might want to wait until all of these repo's have vendored versions of their source instead

sergiocampama:
this is a step in the right direction, thanks @keith! I'm currently planning on doing this for all our repos and was waiting until the skylib maintainer was back from vacation to do a bottom up approach.

If you notice, the last release also had a new release asset called `rules_apple.0.14.0.tar.gz` which has been processed to have the contents at the root, so no `strip_prefix` is needed, and the release notes also contains the sha256 of that file, making it easy to integrate. 

I'll merge this now, but will slightly change once we update our dependencies as well.

keith:
Nbd if you want to wait on this until there are uploaded releases.

apple/repositories.bzl
```diff
@@ -96,34 +95,37 @@ def apple_rules_dependencies(ignore_version_differences = False):
           incompatible versions of depended-upon repositories will be silenced.
     """
     _maybe(
-        git_repository,
+        http_archive,
         name = "bazel_skylib",
-        remote = "https://github.com/bazelbuild/bazel-skylib.git",
-        tag = "0.7.0",
         ignore_version_differences = ignore_version_differences,
+        sha256 = "2c62d8cd4ab1e65c08647eb4afe38f51591f43f7f0885e7769832fa137633dcb",
```
steeve: you might now want to ping sha256 on github generated archives (they can vary)

steeve: oops, just saw that was said before, sorry for the noise

sergiocampama:
We've migrated to `http_archive` for all of our repos now.

PR #366, opened 2019-03-20
AppleBundleVersionInfo load statement fix for ci tests

rahul-malik:
- "@build_bazel_rules_apple//apple:providers.bzl" is the new location of
AppleBundleVersionInfo but there was a test that is failing on CI

### DIFF
--- a/test/apple_bundle_version_test.sh
+++ b/test/apple_bundle_version_test.sh
@@ -22,7 +22,7 @@ function set_up() {
   mkdir -p pkg
 
   cat > pkg/saver.bzl <<EOF
-load("@build_bazel_rules_apple//apple:versioning.bzl",
+load("@build_bazel_rules_apple//apple:providers.bzl",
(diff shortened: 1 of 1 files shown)


test/apple_bundle_version_test.sh
```diff
@@ -48,8 +48,8 @@ saver = rule(
 EOF
 
   cat > pkg/BUILD <<EOF
-load("@build_bazel_rules_apple//apple:versioning.bzl",
-     "apple_bundle_version")
+load("@build_bazel_rules_apple//apple:providers.bzl",
```
sergiocampama: this is actually ok, since it's loading the rule, not the provider

rahul-malik: Doh, will update soon (just walked away from my laptop)

rahul-malik: Updated now!

PR #368, opened 2019-03-24
Move some resource attributes

keith:
This moves some rule_support attributes specific to the resources
partial out of rule_support and into the rules themselves. This breaks
the dependency on rule_support from this partial, making it easier to
call directly from other rules.

This is an alternative to https://github.com/bazelbuild/rules_apple/pull/318

### DIFF
--- a/apple/internal/ios_rules.bzl
+++ b/apple/internal/ios_rules.bzl
@@ -125,6 +125,7 @@ def _ios_application_impl(ctx):
             plist_attrs = ["infoplists"],
             targets_to_avoid = ctx.attr.frameworks,
             top_level_attrs = top_level_attrs,
+            requires_pkginfo = True,
         ),
         partials.settings_bundle_partial(),
         partials.swift_dylibs_partial(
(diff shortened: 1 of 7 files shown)


keith:
It looks like the test failure is unrealted

sergiocampama:
Mmm I don't like the fact that some configuration is removed from rule descriptors, and the amount of changes required for it. How about adding a new method in `partials/resources.bzl` named `resources_partial_no_descriptor` that takes the `binary_plist` and `requires_pkginfo`, then `resources_partial` is modified to take in a `rule_descriptor`, which in turn proxies into `resources_partial_no_descriptor` with the appropriate values? That way the only modifications you need are in `partials/resources.bzl` and the `plist` processing file.

keith:
I'm happy to do that but that feels like a bit more of a workaround than this one did. Just want to gut check again that you don't want to just add a bundle descriptor?


--
Keith Smiley

sergiocampama:
Yeah, I'm certain of not having a bundle descriptor since it's going to be used for resources, and we're having some discussions internally about supporting iOS bundles in a similar manner as `macos_bundle`, which would behave completely different, but under the same product type

keith:
With the recent resource partial changes you no longer need to inject a descriptor to have this behavior, so this is obsoleted

PR #372, opened 2019-03-29
Update

sergiocampama:


sergiocampama:
> Looks like some real build breaks?

fixed with the updated dependencies

Issue #374, opened 2019-04-03
dwarfdump broken in Xcode 10.2


keith:
When using rules_apple with bitcode and bazel 0.24.0, we rely on dwarfdump output https://github.com/bazelbuild/rules_apple/blob/2a8ec18e75a243a98eb169aea9af338cf68bee90/apple/internal/partials/bitcode_symbols.bzl#L63-L71

But when passing `-arch` with Xcode 10.2 this no longer returns anything:

```
% DEVELOPER_DIR=/Applications/Xcode-10.1.0.app dwarfdump -u --arch=arm64 bazel-out/apl-darwin_x86_64-opt/bin/Modules/Catalog/Catalog.__internal__.apple_binary_lipobin
UUID: EBD44D3E-C41A-36DC-B20E-47801D52D6AF (arm64) bazel-out/apl-darwin_x86_64-opt/bin/Modules/Catalog/Catalog.__internal__.apple_binary_lipobin
% DEVELOPER_DIR=/Applications/Xcode-10.2.0.app dwarfdump -u --arch=arm64 bazel-out/apl-darwin_x86_64-opt/bin/Modules/Catalog/Catalog.__internal__.apple_binary_lipobin
```

An alternative is to stop passing `-arch`, which seems to work correctly, and parse the output for the specific arch instead

PR #375, opened 2019-04-03
Work around dwarfdump regression

keith:
In Xcode 10.2 filtering by architecture no longer works correctly.
Instead we can stop filtering and parse the output.

Fixes https://github.com/bazelbuild/rules_apple/issues/374

### DIFF
--- a/apple/internal/partials/bitcode_symbols.bzl
+++ b/apple/internal/partials/bitcode_symbols.bzl
@@ -62,8 +62,8 @@ def _bitcode_symbols_partial_impl(
             # Get the UUID of the arch slice and use that to name the bcsymbolmap file.
             copy_commands.append(
                 ("cp {bitcode_file} " +
-                 "${{OUTPUT_DIR}}/$(dwarfdump -u -arch {arch} {binary} " +
-                 "| cut -d' ' -f2).bcsymbolmap").format(
+                 "${{OUTPUT_DIR}}/$(dwarfdump -u {binary} " +
+                 "| grep \"({arch})\" | cut -d' ' -f2).bcsymbolmap").format(
                     arch = arch,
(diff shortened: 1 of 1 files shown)


sergiocampama:
looks like this doesn't pass tests... can you change `grep {arch}` to `grep \"({arch})\"`? I've tested that it passes the tests

keith:
These failures appear to be unrelated:

```

File: apple/internal/partials/bitcode_symbols.bzl [hunk 1]
```diff
@@ -62,8 +62,8 @@ def _bitcode_symbols_partial_impl(
             # Get the UUID of the arch slice and use that to name the bcsymbolmap file.
             copy_commands.append(
                 ("cp {bitcode_file} " +
-                 "${{OUTPUT_DIR}}/$(dwarfdump -u -arch {arch} {binary} " +
-                 "| cut -d' ' -f2).bcsymbolmap").format(
+                 "${{OUTPUT_DIR}}/$(dwarfdump -u {binary} " +
+                 "| grep \"({arch})\" | cut -d' ' -f2).bcsymbolmap").format(
```
kastiglione: I bet `dwarfdump-classic` would work here. I think the problem is they switched over to `llvm-dwarfdump`.

keith: Interesting! It looks like 10.1 doesn't have that though so we'd have to switch on that

kastiglione: > It looks like 10.1 doesn't have that

🤦‍♂️ 

This is nicer anyway, was just pointing out what likely caused the break.

kastiglione:
I created a diff in llvm to fix the problem: https://reviews.llvm.org/D60252

PR #410, opened 2019-04-26
Allow using a different test runner

ob:
This change allows changing the test runner. For example, to use GitHub.com/linkedin/bluepill you can add this to your `WORKSPACE`:

```
git_repository(
  name = "bptestrunner",
  remote = "https://github.com/ob/bluepill.git",
  commit = "5e9c6ae",
)
```

And use this in your `BUILD.bazel` file:
```
ios_test_runner(
    name = "bp_test_runner_parallel",
    device_type = "iPhone 6s",
    extra_args = ["--num-sims", "4"],
    testrunner = "@bptestrunner//bptestrunner:bptestrunner.par",
)

ios_unit_test(
  name = "some_name",
  runner = "bp_test_runner_parallel",
  # other attributes here... 
)
```

When using KIF or EarlGray, I have found that setting `--num-sims` to run tests in parallel saves about 40% of the test time. For unit tests, parallelism slows tests down so it's better to set `--num-sims` to 1.

### DIFF
--- a/apple/testing/default_runner/ios_test_runner.bzl
+++ b/apple/testing/default_runner/ios_test_runner.bzl
@@ -24,9 +24,13 @@ def _get_template_substitutions(ctx):
     subs = {
         "device_type": ctx.attr.device_type,
         "os_version": ctx.attr.os_version,
-        "testrunner_binary": ctx.executable._testrunner.short_path,
+        "testrunner_binary": ctx.executable.testrunner.short_path,
     }
-    return {"%(" + k + ")s": subs[k] for k in subs}
+    ret = {"%(" + k + ")s": subs[k] for k in subs}
+    if len(ctx.attr.extra_args) > 0:
+        extra_args = " ".join(["'" + arg + "'" for arg in ctx.attr.extra_args])
+        ret["%(extra_args)s"] = extra_args
+    return ret
 
 def _get_execution_environment(ctx):
     """Returns environment variables the test runner requires"""
@@ -52,7 +56,7 @@ def _ios_test_runner_impl(ctx):
         ),
         DefaultInfo(
             runfiles = ctx.runfiles(
-                files = [ctx.file._testrunner],
+                files = [ctx.file.testrunner],
             ),
         ),
     ]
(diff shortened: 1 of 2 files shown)


sergiocampama:
I'd actually like to keep this out of rules_apple. 

Perhaps this can be an opportunity to create a new community based repo with new test runners? All you should need to do is create a new `ios_bluepill_testrunner` (or similar) rule that propagates the `AppleTestRunnerInfo` provider and add a target of this rule to the `runner` attribute in `ios_unit_test`/`ios_ui_test`.

This plugin approach means that new runners don't need to be added to rules_apple (and we can avoid the potential copyright and infrastructure issues that arise from adding new third party dependencies )

ob:
Fair enough, but note that this technically doesn't introduce any third-party dependencies. All it does is expose the `testrunner` attribute. I'll create a separate repo with a different rule.

Thanks!

kastiglione:
> and we can avoid the potential copyright and infrastructure issues that arise from adding new third party dependencies

@sergiocampama are you able to elaborate on this at all?

sergiocampama:
Disclaimer: I'm kinda tired and thought that this PR was adding the bluepill dependency somehow into the rules :) which is what I didn't want to happen. If that was the case, my concern would be that by adding new dependencies, we would need to ensure that we're allowed to depend on them because of licensing (which I'm no expert in, and don't want to deal with them if possible :) )

This is clearly not the case in this particular PR, but my recommendation is still the same. New test runner technologies (e.g. bluepill) should receive their own rules so that there can be a tighter integration. In this case this seems to work ok because bluepill looks like it shares the same flags as xctestrunner, but that may not be the case in the future, meaning that new changes would need to be added to this script to support future versions of bluepill.

This also sets the precedent of adding integrations into the `ios_test_runner` rule for test runners other than xctestrunner, so you can see how in the extreme, the `ios_test_runner` rule would become a support-all rule where none of the runners are supported quite right.

Since the testing rules were designed to have pluggable runners through the `AppleTestRunnerInfo` provider, I believe this is the better approach to add support for other test runner technologies.

PR #412, opened 2019-04-30
Exlclude identities having revoked certs

kastiglione:
I happened to have both an invalid and a valid cert in my keychain. The invalid cert had been revoked, but was still in my keychain. This caused the build to fail the code signing action. This change prevents the underlying issues by doing the following:

1. Always validating the given identity (`--ios_signing_cert_name`) against the output of `security find-identity`
2. Filtering the output of `security find-identity` to exclude lines that contain `CSSMERR_TP_CERT_REVOKED`

In my case, the build was with `--ios_signing_cert_name="iPhone Developer"`, but my keychain has two matches for this, one of which has been revoked. I don't know how I got into this state, but it shouldn't prevent code signing.

Relevant output that was used to help determine a root cause and fix:

```
% security find-identity -v -p codesigning                                   
  1) <...> "lldb_codesign"
  2) <...> "iPhone Developer: Dave Lee (3FD7YP60A6)" (CSSMERR_TP_CERT_REVOKED)
  3) <...> "iPhone Developer: Dave Lee (9FBWYY4H42)"
     3 valid identities found
```

### DIFF
--- a/tools/codesigningtool/codesigningtool.py
+++ b/tools/codesigningtool/codesigningtool.py
@@ -78,17 +78,26 @@ def _get_identities_from_provisioning_profile(mpf):
         yield _certificate_fingerprint(identity.data)
 
 
-def _find_codesign_identities():
+def _find_codesign_identities(identity=None):
     """Finds code signing identities on the current system."""
     ids = []
     output, stderr = _check_output([
         "security", "find-identity", "-v", "-p", "codesigning",
     ])
     output = output.decode("utf-8").strip()
+    pattern = "(?P<hash>[A-F0-9]{40})"
+    if identity:
+        pattern += r'\s+"(?P<full_name>.*?{}.*?)"'.format(identity)
+    regex = re.compile(pattern)
     for line in output.splitlines():
-        m = re.search(r"([A-F0-9]{40})", line)
+        # CSSMERR_TP_CERT_REVOKED comes from Security.framework/cssmerr.h
+        if "CSSMERR_TP_CERT_REVOKED" in line:
+            continue
+        m = regex.search(line)
         if m:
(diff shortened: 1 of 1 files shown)


kastiglione:
There are some downstream failures, so this isn't enough as is. Will update shortly.

kastiglione:
Updated to more properly handle this case. Ready for review.

kastiglione:
@sergiocampama The PR says "This branch is out-of-date with the base branch". Do I have to rebase, or does the copybara tooling take care of this?

sergiocampama:
It'll probably work, since I don't think there are other changes pending in this particular file.

Could you squash the PR though? The new flow will merge instead of squash and rebase, so it'll appear as 4 different commits in the master branch.

kastiglione:
squashed (and rebased)

tools/codesigningtool/codesigningtool.py
```diff
@@ -78,17 +78,26 @@ def _get_identities_from_provisioning_profile(mpf):
...
+def _find_codesign_identities(identity=None):
     """Finds code signing identities on the current system."""
     ids = []
     output, stderr = _check_output([
         "security", "find-identity", "-v", "-p", "codesigning",
     ])
     output = output.decode("utf-8").strip()
+    pattern = "(?P<hash>[A-F0-9]{40})"
+    if identity:
+        pattern += r'\s+"(?P<full_name>.*?{}.*?)"'.format(identity)
```
thomasvl: fyi - we'll tweak this as it lands, I think this should call `re.escape` on `identity` before inserting it into the regex string.  Since it comes from the developers account, it likely could have characters that would otherwise having meaning in a regex.

PR #421, opened 2019-05-03
Remove test runner working directory after tests

keith:
Otherwise these directories accumulate.


keith:
For reference the macOS tests remove this dir:

https://github.com/bazelbuild/rules_apple/blob/6273f0de390158a8bbf299ba5c60644e4293daa6/apple/testing/default_runner/macos_test_runner.template.sh#L42-L45

Issue #408, opened 2019-04-26
Case insensitive headers in `ios_static_framework`s


keith:
When building a `ios_static_framework` for the [`CgRPC`](https://github.com/bazelbuild/rules_swift/blob/32001f34b291c00435e151c5c0d990a514e6eefd/third_party/com_github_grpc_grpc_swift/BUILD.overlay#L37-L56) target, we've hit an interesting issue. Building it results in this error:

```
ERROR: .../BUILD.bazel:78:1: Processing and signing CgRPCFramework failed (Exit 1)
replace bazel-out/darwin-fastbuild/bin/external/grpc_swift/CgRPCFramework_archive-root/CgRPC.framework/Headers/CgRPC.h? [y]es, [n]o, [A]ll, [N]one, [r]ename:  NULL
(EOF or read error, treating as "[N]one" ...)
Target @grpc_swift//:CgRPCFramework failed to build
Use --verbose_failures to see the command lines of failed build steps.
```

Because the [unprocessed archive](https://github.com/bazelbuild/rules_apple/blob/f92ca9dff623caac6598c624e44823f972a9e91f/tools/bundletool/process_and_sign.sh.template#L13) ends up with 2 headers:

```
CgRPC.framework/CgRPC
CgRPC.framework/Headers/cgrpc.h
CgRPC.framework/Headers/CgRPC.h
CgRPC.framework/Modules/module.modulemap
```

This is because of the framework name I'm producing (CgRPC) and their default header being [`cgrpc.h`](https://github.com/grpc/grpc-swift/blob/d5f0c46e4fd067588d17f57894b8b0630c9daca1/Sources/CgRPC/shim/cgrpc.h)

keith:
Maybe one solution for this case would be to allow users to provide the name of the umbrella header in the framework, and use that in the `module.modulemap` therefore removing the need to create a `CgRPC.h`.

keith:
Another temporary workaround for this case is renaming the header with a genrule:

```python
genrule(
    name = "main-header",
    srcs = ["Sources/CgRPC/shim/cgrpc.h"],
    outs = ["cgrpc-main.h"],
    cmd = "cp $< $@",
)
```

Which results in:

```
CgRPC.framework/
CgRPC.framework/Headers/
CgRPC.framework/Headers/cgrpc-main.h
CgRPC.framework/Headers/CgRPC.h
CgRPC.framework/CgRPC
CgRPC.framework/Modules/
[...]

PR #437, opened 2019-05-16
Add umbrella_header to ios_static_framework

keith:
This attribute allows users to specify an umbrella header to use for the
framework in place of generating a default one. This is useful for
frameworks that have multiple headers, including a natural umbrella
header, that has a conflicting name with the one that would be
generated.

The idea for this came out of the discussion on https://github.com/bazelbuild/rules_apple/pull/359

Fixes https://github.com/bazelbuild/rules_apple/issues/408

### DIFF
--- a/apple/internal/partials/static_framework_header_modulemap.bzl
+++ b/apple/internal/partials/static_framework_header_modulemap.bzl
@@ -115,14 +115,19 @@ def _create_umbrella_header(actions, output, headers):
     content = "\n".join(import_lines) + "\n"
     actions.write(output = output, content = content)
 
-def _static_framework_header_modulemap_partial_impl(ctx, hdrs, binary_objc_provider):
+def _static_framework_header_modulemap_partial_impl(ctx, hdrs, umbrella_header, binary_objc_provider):
     """Implementation for the static framework headers and modulemaps partial."""
     bundle_name = bundling_support.bundle_name(ctx)
 
     bundle_files = []
 
     umbrella_header_name = None
-    if hdrs:
+    if umbrella_header:
+        umbrella_header_name = umbrella_header.basename
+        bundle_files.append(
+            (processor.location.bundle, "Headers", depset(hdrs + [umbrella_header])),
+        )
+    elif hdrs:
         umbrella_header_name = "{}.h".format(bundle_name)
         umbrella_header_file = intermediates.file(ctx.actions, ctx.label.name, umbrella_header_name)
         _create_umbrella_header(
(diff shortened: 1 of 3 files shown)


apple/internal/partials/static_framework_header_modulemap.bzl
```diff
@@ -115,14 +115,19 @@ def _create_umbrella_header(actions, output, headers):
...
-def _static_framework_header_modulemap_partial_impl(ctx, hdrs, binary_objc_provider):
+def _static_framework_header_modulemap_partial_impl(ctx, hdrs, custom_umbrella_header, binary_objc_provider):
     """Implementation for the static framework headers and modulemaps partial."""
     bundle_name = bundling_support.bundle_name(ctx)
 
     bundle_files = []
 
     umbrella_header_name = None
-    if hdrs:
+    if custom_umbrella_header:
```
keith: I'm happy to combine this logic with the logic below, but I didn't find a great way of doing that where it was still clear to read

apple/internal/rule_factory.bzl
```diff
@@ -525,6 +525,13 @@ use only extension-safe APIs.
 A list of `.h` files that will be publicly exposed by this framework. These headers should have
 framework-relative imports, and if non-empty, an umbrella header named `%{bundle_name}.h` will also
 be generated that imports all of the headers listed here.
+""",
+            ),
+            "custom_umbrella_header": attr.label(
```
keith: I have no preference on the name here, this could also just be `umbrella_header` if folks wanted that

keith:
Should I update the docs here too, or are those generated?

apple/internal/rule_factory.bzl
sergiocampama: just call it `umbrella_header`

apple/internal/rule_factory.bzl
```diff
@@ -525,6 +525,13 @@ use only extension-safe APIs.
...
+                allow_single_file = [".h"],
+                doc = """
+A single `.h` file to use in place of generating an umbrella header for the framework. If this is not
+provided a default header will be used instead.
```
sergiocampama: "A single .h file to use as the umbrella header for this framework. Usually, this header will have the same name as this target, so that clients can load the header using the `#import <MyFramework/MyFramework.h>` format. If this attribute is not specified, an umbrella header will be generated under the same name as this target."

keith:
Updated!

keith:
I don't think these red tests are my doing, it appears to be an upstream breakage

PR #479, opened 2019-06-18
Fix framework imports with Swift modules

keith:
With bazel 0.27.0 and the flip of `incompatible_objc_framework_cleanup`
https://github.com/bazelbuild/bazel/issues/7944 framework imports of
Swift frameworks broke because the modules were included and processed
as module map files. This change makes them available in the build
but without processing them. They are handled automatically by clang /
swift since the framework is included using `-F`

RELNOTES: Fix Swift frameworks for Bazel 0.27.0 and incompatible_objc_framework_cleanup

### DIFF
--- a/apple/internal/apple_framework_import.bzl
+++ b/apple/internal/apple_framework_import.bzl
@@ -63,8 +63,13 @@ def _classify_framework_imports(framework_imports):
             # This matches /Headers/ and /PrivateHeaders/
             header_imports.append(file)
             continue
-        if "/Modules/" in file_short_path:
-            module_map_imports.append(file)
+        if file_short_path.endswith(".swiftmodule") or file_short_path.endswith(".swiftinterface"):
+            # Add Swift's module files to header_imports so that they are correctly included in the build
+            # by Bazel but they aren't processed in any way
+            header_imports.append(file)
+            continue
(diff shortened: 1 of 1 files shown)


sergiocampama:
@googlewalt ^^

we made the wrong assumption that all `/Modules/` files are modulemaps, which for Swift they might not be

apple/internal/apple_framework_import.bzl
```diff
@@ -63,8 +63,13 @@ def _classify_framework_imports(framework_imports):
             # This matches /Headers/ and /PrivateHeaders/
             header_imports.append(file)
             continue
-        if "/Modules/" in file_short_path:
-            module_map_imports.append(file)
+        if file_short_path.endswith(".swiftmodule") or file_short_path.endswith(".swiftinterface"):
+            # Add swift's module files to header_imports so that they are correctly included in the build
+            # by bazel but they aren't added to any arguments
```
sergiocampama: ... but they are not processed in any other way.

sergiocampama: capital B in Bazel and S in Swift

keith:
I added `RELNOTES` to the description, let me know if that's not the right format

PR #483, opened 2019-07-09
Use latest version in README

ob:
I think many people just copy and paste the README into their WORKSPACE file. It should use the latest version. It would be ideal to automate updating it on every release, but I don't know how releases are cut. Are they cut manually?

Also, without the protobuf stuff the WORKSPACE file didn't work, so I added it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,12 +26,14 @@ replacing the version number in the `tag` attribute with the version of the
 rules you wish to depend on:
 
 ```python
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -54,6 +56,13 @@ load(
 )
 
 apple_support_dependencies()
+
+load(
```
sergiocampama: can you add a comment here that this is a dependency of rules_swift?

sergiocampama:
already done in another commit

PR #486, opened 2019-07-10
Handle ambiguous signer in codesigningtool.py

kastiglione:
This resolves a case we hit recently. Someone had more than one developer certificate for with the exact same common name, including the hex string identifier. All were valid. Calling `codesign` with an ambiguous signing identity is an error.

This change addresses this case by passing the unique hash. An example `codesign` invocation could look like:

```sh
codesign --sign "434FA21B6A3CE7FE758E7EA843ED159B24E8C43F" ...
```

### DIFF
--- a/apple/internal/rule_support.bzl
+++ b/apple/internal/rule_support.bzl
@@ -288,6 +288,16 @@ _RULE_TYPE_DESCRIPTORS = {
             requires_provisioning_profile = False,
             skip_signing = True,
         ),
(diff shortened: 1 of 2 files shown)


kastiglione:
I've simplified this from the original PR commit.

kastiglione:
@sergiocampama In the past I've squashed before commit, do you want me to squash this?

sergiocampama:
in general yeah, I just forgot this time, but it's no big deal :)

thanks for the change!

PR #492, opened 2019-07-15
Add protobuf_deps to README instructions

keith:
This is required by rules_swift


keith:
https://github.com/bazelbuild/rules_apple/pull/483

PR #511, opened 2019-07-25
Add option to disable environment plist generation

keith:
This generation is expensive if you have a lot of targets that need it.
By disabling this on our project we saved ~400 seconds of parallel build
time, which resulted in 20 seconds being saved overall.

### DIFF
--- a/doc/common_info.md
+++ b/doc/common_info.md
@@ -172,6 +172,13 @@ the build to Apple. If you want to disable bundling SwiftSupport in your ipa for
 other device or enterprise builds, you can pass
 `--define=apple.package_swift_support=no` to `bazel build`
 
+### Disable environment plist generation
+
+For App Store builds Apple requires your plist contains contains version
+entries of the tools used to build your app. In order to save time
+generating this plist for non App Store builds you can pass
+`--define=apple.add_environment_plist=no` to `bazel build`
+
 ### Codesign Bundles for the Simulator {#apple.codesign_simulator_bundles}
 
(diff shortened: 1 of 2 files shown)


doc/common_info.md
```diff
@@ -172,6 +172,13 @@ the build to Apple. If you want to disable bundling SwiftSupport in your ipa for
 other device or enterprise builds, you can pass
 `--define=apple.package_swift_support=no` to `bazel build`
 
+### Disable environment plist generation
+
+For App Store builds Apple requires your plist contains some keys in
```
sergiocampama: ... contains version entries of the tools used ...

thomasvl:
I wonder if we can have a target within the rules that generates the file and make its output a attribute of all bundles?  That way it builds in one place so it builds/caches?  As long as there are transitions in place, the SDK/target platform should be applied.  Likewise the Xcode version used should also feed into the caching of the result correct.  i.e. - instead of a variable to suppress it, we get something that is generated once per repo for a given sdk/xcode/etc. config, and then reused from the cache after that.

keith:
That has been our internal workaround, I made a `generate_environment_plist` rule and I depend on that from all the targets. As long as nothing in the file depends on specific target attributes (which it doesn't look like it does) I think that would be fine

keith:
Submitted a WIP of that approach https://github.com/bazelbuild/rules_apple/pull/522

keith:
Closing for that PR

PR #522, opened 2019-08-05
Move to shared environment.plist generation

keith:
This introduces a new rule that bundling rules depend on to generate the
environment plist a single time. This file is required when uploading
builds to Apple, and doesn't change if you build many similar targets.
This can greatly decrease build times if you have a lot of targets that
require this.

### DIFF
--- a/apple/internal/BUILD
+++ b/apple/internal/BUILD
@@ -29,7 +29,32 @@ filegroup(
 )
 
 environment_plist(
-    name = "environment_plist",
+    name = "environment_plist_ios",
+    platform_type = "ios",
+    visibility = [
+        "//visibility:public",
+    ],
+)
+
+environment_plist(
+    name = "environment_plist_macos",
+    platform_type = "macos",
+    visibility = [
+        "//visibility:public",
+    ],
+)
+
+environment_plist(
+    name = "environment_plist_watchos",
+    platform_type = "watchos",
+    visibility = [
+        "//visibility:public",
+    ],
+)
+
+environment_plist(
+    name = "environment_plist_tvos",
+    platform_type = "tvos",
     visibility = [
         "//visibility:public",
     ],
--- a/apple/internal/BUILD
+++ b/apple/internal/BUILD
@@ -5,6 +5,11 @@
 
 licenses(["notice"])
 
+load(
+    "@build_bazel_rules_apple//apple/internal:environment_plist.bzl",
+    "environment_plist",
+)
+
 # Consumed by bazel tests.
 filegroup(
     name = "for_bazel_tests",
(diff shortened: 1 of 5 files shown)


keith:
This is an alternative to https://github.com/bazelbuild/rules_apple/pull/511

apple/internal/BUILD
```diff
@@ -22,3 +27,10 @@ filegroup(
         "//apple:__subpackages__",
     ],
 )
+
+environment_plist(
```
sergiocampama: we'll need to create 4 of these targets, once per platform, otherwise `macos_application` would use the default ios one since `platform_type` is not being set anywhere else. 

and then the attribute can be something like:

```
"_environment_plist": attr.label(
            default = "@build_bazel_rules_apple//apple/internal:environment_plist_{}".format(platform),
)
```

you'll need to extract this attribute from the `_COMMON_PRIVATE_TOOL_ATTRS` dictionary into  the `_create_apple_bundling_rule` and `_create_apple_binary_rule` methods. You can create a new method for common attributes that uses the platform.

Once we migrate to transitions, we'll be able to simplify this greatly :)

apple/internal/environment_plist.bzl
```diff
@@ -0,0 +1,52 @@
...
+environment_plist = rule(
+    attrs = dicts.add(
+        rule_factory.common_tool_attributes,
+        {
+            "_environment_plist": attr.label(
+                cfg = "host",
+                executable = True,
+                default = Label("@build_bazel_rules_apple//tools/environment_plist"),
+            ),
+            "platform_type": attr.string(mandatory = True),
```
sergiocampama: can you add a small line for docs here?

apple/internal/environment_plist.bzl
```diff
@@ -0,0 +1,52 @@
...
+            "--output",
+            ctx.outputs.plist.path,
+        ],
+    )
+
+environment_plist = rule(
+    attrs = dicts.add(
+        rule_factory.common_tool_attributes,
+        {
+            "_environment_plist": attr.label(
```
sergiocampama: to avoid confusion, can this be `_environment_plist_tool` instead?

apple/internal/environment_plist.bzl
```diff
@@ -0,0 +1,52 @@
...
+                cfg = "host",
+                executable = True,
+                default = Label("@build_bazel_rules_apple//tools/environment_plist"),
+            ),
+            "platform_type": attr.string(mandatory = True),
+        },
+    ),
+    fragments = ["apple"],
+    outputs = {"plist": "%{name}.plist"},
+    implementation = _environment_plist,
```
sergiocampama: same here, small docs for the rule.

apple/internal/environment_plist.bzl
```diff
@@ -0,0 +1,52 @@
+load(
```
sergiocampama: please also copy the copyright stuff from another bzl file

keith:
Looks like the `macos_binary_infoplist` rule doesn't follow the normal rule_factory flow, so it's failing because of that. I'll add this new attr there

PR #549, opened 2019-08-16
Default iOS tests to passed versions

keith:
This makes the --ios_simulator_version and --ios_simulator_device flags,
affect which devices are used for testing if you don't otherwise provide
a device or OS version

This reverts commit 1a5bd8e3163e6966a18a451c8bb455bb97e7a9d6.

### DIFF
--- a/apple/testing/default_runner/ios_test_runner.bzl
+++ b/apple/testing/default_runner/ios_test_runner.bzl
@@ -21,9 +21,11 @@ load(
 
 def _get_template_substitutions(ctx):
     """Returns the template substitutions for this runner."""
+    os_version = ctx.attr.os_version or str(ctx.fragments.objc.ios_simulator_version)
+    device_type = ctx.attr.device_type or ctx.fragments.objc.ios_simulator_device
     subs = {
-        "device_type": ctx.attr.device_type,
-        "os_version": ctx.attr.os_version,
+        "device_type": device_type,
+        "os_version": os_version,
         "testrunner_binary": ctx.executable._testrunner.short_path,
     }
     return {"%(" + k + ")s": subs[k] for k in subs}
--- a/apple/testing/default_runner/ios_test_runner.bzl
+++ b/apple/testing/default_runner/ios_test_runner.bzl
@@ -21,11 +21,11 @@ load(
 
 def _get_template_substitutions(ctx):
     """Returns the template substitutions for this runner."""
-    os_version = ctx.attr.os_version or str(ctx.fragments.objc.ios_simulator_version)
-    device_type = ctx.attr.device_type or ctx.fragments.objc.ios_simulator_device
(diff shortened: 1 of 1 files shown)


keith:
blocked on https://github.com/bazelbuild/bazel/pull/9191

apple/testing/default_runner/ios_test_runner.bzl
```diff
@@ -21,9 +21,11 @@ load(
 
 def _get_template_substitutions(ctx):
     """Returns the template substitutions for this runner."""
+    os_version = ctx.attr.os_version or ctx.fragments.objc.ios_simulator_version or ""
+    device_type = ctx.attr.device_type or ctx.fragments.objc.ios_simulator_device or ""
     subs = {
-        "device_type": ctx.attr.device_type,
-        "os_version": ctx.attr.os_version,
+        "device_type": device_type,
+        "os_version": str(os_version),
```
thomasvl: Why does this one need `str()`?

keith: Otherwise the value from the flag is a DottedVersion

keith:
@thomasvl I think this should be good to merge now

apple/testing/default_runner/ios_test_runner.bzl
keith: Yes the None case is why I'm doing it this way

keith:
The failures definitely appear to be related to this change, but I'm not sure why ATM

thomasvl:
Is your branch up to date?  Maybe something else has changed, and I think kokoro might try doing PRs off the branch point?

keith:
It is up to date yes

keith:
I'm not sure why this change would affect the default simulator version for UI tests since I removed the defaults in bazel itself and the tests aren't specifying a version, but if it did affect that the issue might be that 12.2 (which is what the tests are trying to use) isn't from the `xcode-select`'d Xcode on the CI machines, and it isn't installed globally, but it is the default version for the `--xcode_version` being used (although this shouldn't be the case since we're not specifying one.

thomasvl:
Seeing the failures on other things now also, so it isn't this cl.  I'll return to this after figuring out what's up on BuildKite.

keith:
So actually right now those aren't exposed to starlark. From reading the code it appears to be because bazel wants to deprecate the individual accesses for https://github.com/bazelbuild/bazel/blob/a330febdddbae5751b6bb0cc0a6a84f451df2304/src/main/java/com/google/devtools/build/lib/skylarkbuildapi/apple/ObjcConfigurationApi.java#L62-L74 instead, which rules_apple doesn't currently consume at all.

But yes, in theory we should, but also right now tvOS tests (publicly) are a no-op 😄 https://github.com/bazelbuild/rules_apple/blob/97951dce38d36316e870950eaa0c7a5583353a87/apple/testing/default_runner/tvos_test_runner.template.sh

PR #559, opened 2019-08-23
Handle possible empty values for --ios_simulator_device and --ios_simulator_version when invoking bazel run.

swiple-rules-gardener:
Handle possible empty values for --ios_simulator_device and --ios_simulator_version when invoking bazel run.


keith:
Thanks for fixing this!

PR #566, opened 2019-08-27
Enable linking of imported Swift static frameworks

### LINKED ISSUE #557: Support Swift implemented frameworks in apple_static_framework_import
Can we make `apple_static_framework_import` support frameworks that are implemented in Swift? Currently, using a framework implemented in Swift will result in linker errors, for missing libraries such as `swiftFoundation`. Perhaps by adding something like a `swift = True` keyword argument?

kastiglione:
This supports the case where a non-Swift target, such as an `ios_application`, has a `apple_static_framework_import` dependency whose binary depends on the Swift runtime.

Resolves #557.


apple/internal/apple_framework_import.bzl
```diff
@@ -223,6 +236,17 @@ def _apple_static_framework_import_impl(ctx):
     if ctx.attr.weak_sdk_frameworks:
         objc_provider_fields["weak_sdk_framework"] = depset(ctx.attr.weak_sdk_frameworks)
 
+    imported_swiftmodules = [
+        f
+        for f in header_imports
+        if _is_swiftmodule(f.basename)
+    ]
```
keith: even tho it's a bit more imperative it might be nice to build this list in the for loop above instead

apple/internal/apple_framework_import.bzl
```diff
@@ -223,6 +236,17 @@ def _apple_static_framework_import_impl(ctx):
...
+
+    if imported_swiftmodules:
+        swift_toolchain = ctx.attr._toolchain[SwiftToolchainInfo]
+        swift_linkopts = swift_common.swift_runtime_linkopts(True, swift_toolchain)
+        objc_provider_fields["linkopt"] = depset(swift_linkopts)
```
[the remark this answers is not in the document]
sergiocampama: I think @keith 's right and we need at least the `SwiftUsageInfo` provider so that this [check](https://github.com/bazelbuild/rules_apple/blob/master/apple/internal/partials/swift_dylibs.bzl#L121) is true and we consider the binary that's linking against the framework to be considered for Swift stdlib dylib processing. 

I also think that returning this provider will also add the correct linkopts, since they're added using the same check.

So I'd change this block to something like:

```
# If we find any Swift module within the imported headers, add the SwiftUsageInfo provider
# to signal the top level targets that there is Swift involved in the build.
for file in header_imports:
    if _is_swift_module(f.basename):
        toolchain = ctx.attr._toolchain[SwiftToolchainInfo]
        providers.append(SwiftUsageInfo(toolchain = toolchain))
        break
```

that way we don't iterate over all files unnecesarily, and we let the downstream logic to add the required linkopts. Can you try this out and verify it fits your needs?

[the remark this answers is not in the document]
kastiglione: I didn't know about `SwiftUsageInfo`, thanks.

apple/internal/apple_framework_import.bzl
kastiglione: Do you mean in `_classify_framework_imports()`?

keith: ah nvm, better to leave here

kastiglione: Actually, it's being removed per @sergiocampama's comment.

apple/internal/apple_framework_import.bzl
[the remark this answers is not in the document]
kastiglione: Updated.

kastiglione:
Updated to add tests.

kastiglione:
The test failures look unrelated.

test/ios_framework_import_test.sh
```diff
@@ -112,6 +112,31 @@ EOF
...
+  SharedClass *sharedClass = [[SharedClass alloc] init];
+  [sharedClass doSomethingShared];
+  return 0;
+}
+EOF
+
+    do_build ios //app:app || fail "Should build"
+}
+
+
```
sergiocampama: nit: remove extra line

kastiglione:
Fixed the nit. Thanks for reviewing.

README.md
```diff
@@ -82,6 +82,7 @@ ios_application(
     bundle_id = "com.example.app",
     families = ["iphone", "ipad"],
     infoplists = [":Info.plist"],
+    minimum_os_version = "10.0",
```
kastiglione: mistake, removing.

sergiocampama:
I'm a bit worried about that generated header without the disclaimer at the top, but I understand why it is that way. if our tools complain, I'll fix it internally

kastiglione:
If necessary I can change the script to have a disclaimer.

sergiocampama:
no need, we figured it out :)

kastiglione:
This caused a problem in cases where a Swift implemented `apple_static_framework_import` was a dep of an app that had other swift deps. Fix is here: https://github.com/bazelbuild/rules_swift/pull/306

PR #567, opened 2019-08-27
Pass imported swiftmodules to ld with -add_ast_path flag

kastiglione:
When using imported swift frameworks, and when building `dbg`, pass the `.swiftmodule` files to the linker using the `-add_ast_path`. This enables lldb expression evaluation using such modules.


apple/internal/apple_framework_import.bzl
```diff
@@ -223,6 +256,25 @@ def _apple_static_framework_import_impl(ctx):
...
+    imported_swiftmodules = [
+        f
+        for f in header_imports
+        if _is_swiftmodule(f.basename)
+    ]
+
+    if imported_swiftmodules:
+        swift_toolchain = ctx.attr._toolchain[SwiftToolchainInfo]
+        swift_linkopts = swift_common.swift_runtime_linkopts(True, swift_toolchain)
+        if ctx.var["COMPILATION_MODE"] == "dbg":
```
keith: rules_swift does this for fastbuild too https://github.com/keith/rules_swift/blob/2c84d8615ca7046c343d747545b7ec77fe4a7c71/swift/internal/api.bzl#L315-L335

apple/internal/apple_framework_import.bzl
```diff
@@ -58,6 +58,26 @@ def _is_swiftmodule(path):
...
+    # Assume that one Swift module or interface is valid.
+    if len(swiftmodule_files) == 1:
+        return swiftmodule_files
+
+    # Filter to specific architectures, for example "x86_64.swiftmodule".
+    cpu_prefix = cpu + "."
+    matches = [
+        f
+        for f in swiftmodule_files
+        if f.basename.startswith(cpu_prefix)
```
sergiocampama: is this format a requirement or a standard? could we find swift modules that do not follow this naming convention?

apple/internal/apple_framework_import.bzl
```diff
@@ -245,6 +265,14 @@ def _apple_static_framework_import_impl(ctx):
     if imported_swiftmodules:
         swift_toolchain = ctx.attr._toolchain[SwiftToolchainInfo]
         swift_linkopts = swift_common.swift_runtime_linkopts(True, swift_toolchain)
+        if ctx.var["COMPILATION_MODE"] == "dbg":
+            cpu = ctx.fragments.apple.single_arch_cpu
+            ast_files = _swiftmodules_for_cpu(imported_swiftmodules, cpu)
+            objc_provider_fields["link_inputs"] = depset(ast_files)
+            swift_linkopts.extend([
```
sergiocampama: @allevato is there a way to reuse something in rules_swift to propagate the swift modules and linkopts without repeating the linker flags here?

apple/internal/apple_framework_import.bzl
sergiocampama: @allevato same here, could we reuse some logic from rules_swift here to avoid duplicating the logic to check when to add these flags?

apple/internal/apple_framework_import.bzl
allevato: Not with the APIs that exist today. The catch is that the way module ASTs are handled for debugging is platform-specific; on Apple platforms the .swiftmodule files are passed to the linker with `-add_ast_path`, but on Linux, a separate `swift -modulewrap` action is run that embeds the module in a `.o` file and that gets passed to the linker. When using the _compilation_ API this is all handled in rules_swift for you, but we don't provide hooks to take an already precompiled module and do the right thing.

It's definitely not a bad idea to expose a standalone API to generalize this (the `swift_import` rule would benefit from it), but it's extremely low-priority.

apple/internal/apple_framework_import.bzl
allevato: Ditto vvv

apple/internal/apple_framework_import.bzl
kastiglione: Requirement, I'm not sure. Standard? Seems so. The swift repo testing declares a variable for this purpose: `target-swiftmodule-name`.

https://github.com/apple/swift/blob/cd84bc17ba5bce7cef6e560c4930644f44d46625/test/lit.cfg#L1612

And from [docs/Testing.md](https://github.com/apple/swift/blob/cd84bc17ba5bce7cef6e560c4930644f44d46625/docs/Testing.md):

> `%target-swiftmodule-name` and `%target-swiftdoc-name`: the basename of swiftmodule and swiftdoc files for a framework compiled for the target (for example, `arm64.swiftmodule` and `arm64.swiftdoc`).

apple/internal/apple_framework_import.bzl
kastiglione: > `swift-modulewrap` action is run that embeds the module in a `.o` file and that gets passed to the linker

TIL!

apple/internal/apple_framework_import.bzl
allevato: It looks like the future is getting more complicated than this.

The code that handles module loading is in https://github.com/apple/swift/blob/master/lib/Serialization/SerializedModuleLoader.cpp.

For frameworks, it looks like `$MODULE_NAME.swiftmodule/$ARCH.swiftmodule` is somewhat [legacy behavior](https://github.com/apple/swift/blob/master/lib/Serialization/SerializedModuleLoader.cpp#L43-L45).

The [modern behavior](https://github.com/apple/swift/blob/master/lib/Serialization/SerializedModuleLoader.cpp#L40) (which looks like it may have been introduced in Swift 5.1) does normalization with the full triple, not just the architecture, then uses the above as a fallback. That normalization is implemented [here](https://github.com/apple/swift/blob/master/lib/Basic/Platform.cpp#L310), and generates names like `x86_64-apple-ios-simulator.swiftmodule`.

There's also an [edge case](https://github.com/apple/swift/blob/master/lib/Serialization/SerializedModuleLoader.cpp#L47-L53) where the compiler will look for `arm.swiftmodule` instead of a more specific `arm64`, `armv7`, etc. variant.

So, depending on who builds the framework in question, if it emits modules with the newer names, it looks like this code may fail to find them.

kastiglione: Great info thanks Tony.

_For future reference_, (since it's good documentation) here are the same links but pointing to the commits at this time:

1. [lib/Serialization/SerializedModuleLoader.cpp](https://github.com/apple/swift/blob/ca1348e6ae2ae872b92b8fc1cb7e3febb7363f4e/lib/Serialization/SerializedModuleLoader.cpp)
2. [legacy behavior](https://github.com/apple/swift/blob/ca1348e6ae2ae872b92b8fc1cb7e3febb7363f4e/lib/Serialization/SerializedModuleLoader.cpp#L43-L45)
3. [modern behavior](https://github.com/apple/swift/blob/ca1348e6ae2ae872b92b8fc1cb7e3febb7363f4e/lib/Serialization/SerializedModuleLoader.cpp#L40)
4. [normalization](https://github.com/apple/swift/blob/ca1348e6ae2ae872b92b8fc1cb7e3febb7363f4e/lib/Basic/Platform.cpp#L310)
5. [edge case](https://github.com/apple/swift/blob/ca1348e6ae2ae872b92b8fc1cb7e3febb7363f4e/lib/Serialization/SerializedModuleLoader.cpp#L47-L53)

kastiglione: It's too bad the `ld` can't be given a `.swiftmodule` bundle, and use this logic themselves. I think I'll file a FB for this.

kastiglione: In our case, we have our own rule that generates `.swiftmodule` bundles, which follows the "legacy" convention, so that's one case we can easily cover. I don't think we need to worry about the edge case, unless requested by someone, at the same time it's not that complicated if we wanted to support it. For the modern behavior, I think we could defer until someone asks for it.

kastiglione: Should this bundle layout logic live in rules_swift?

kastiglione: I filed FB7221978 for adding .swiftmodule directory support to ld.

allevato: Even though the compiler supports .swiftmodule directories in general `-I` search paths, they're mostly used by Apple platforms for framework bundles and I don't believe any other platforms use fat libraries in the same way. So I don't think there's much gained by splitting a small part of this change across two repositories, when rules_swift otherwise wouldn't use that code.

kastiglione:
I've rebased and modified on top of #566.

kastiglione:
- [ ] Decide how to handle `dbg`/`fastbuild` logic duplicated here and in rules_swift
- [ ] Same, but for `-add_ast_path`
- [ ] Is handling only the "legacy" naming convention ok?
- [ ] Update tests to confirm the `-add_ast_path` was applied to the binary

apple/internal/apple_framework_import.bzl
```diff
@@ -59,6 +59,15 @@ def _is_swiftmodule(path):
     """Predicate to identify Swift modules/interfaces."""
     return path.endswith(".swiftmodule") or path.endswith(".swiftinterface")
 
+def _swiftmodules_for_cpu(swiftmodule_files, cpu):
+    """Select the cpu specific swiftmodule."""
+    for swiftmodule in swiftmodule_files:
+        root, _ = paths.split_extension(swiftmodule.basename)
+        if root == cpu:
```
keith: I think this might have been mentioned elsewhere but does the CPU correctly handle the armv7 case? I think in that case the swiftmodule is `arm.swiftmodule` 🙃  https://github.com/ios-bazel-users/ios-bazel-users/blob/master/swift_static_framework/swift_static_framework.bzl#L8-L13

kastiglione: Thanks, will update.

kastiglione:
I changed approach slightly. Instead of implementing it all inline, I duplicated and customized the interfaces from rules_swift that we'd like to ultimately reuse. This makes the callsite cleaner, and makes future migration – should those APIs become public, easier. It also calls out explicitly that this behavior is duplicated, as a reminder that hopefully one day we can reuse rules_swift.

apple/internal/apple_framework_import.bzl
kastiglione: `arm` is supported in c4ab505895a5cf37da737b8cd62b56084596c74e

sergiocampama:
I'm ok with supporting only the legacy behavior for now.  If adding a test is not too complex, could you add one?

kastiglione:
Yep, was intending to add tests. Will do that now.

test/ios_framework_import_test.sh
```diff
@@ -202,7 +202,16 @@ let sharedClass = SharedClass()
...
+    do_build ios --compilation_mode=dbg //app:app || fail "Should build"
+
+    unzip_single_file "test-bin/app/app.ipa" "Payload/app.app/app" |
+        nm -aj - | \
+            grep -e /app_main.swiftmodule \
+                 -e /iOSSwiftStaticFramework.swiftmodule/x86_64.swiftmodule | \
+            wc -l | \
+            grep -e '^\s*2\s*$' > /dev/null \
+            || fail "Could not find .swiftmodule AST references in binary; " \
+                    "linkopts may have not propagated"
```
kastiglione: I tacked this test on to an existing test. Hopefully this is ok, the main thing I'm unsure of is that I forced compilation mode to `dbg`. How do these tests get a compilation mode? Do they get called with each compilation mode?

kastiglione: Also, suggestions welcome for how to structure this shell code 😬

kastiglione:
@sergiocampama I'm considering making a patch for this locally, but I wanted to check if you had any additional feedback.

test/ios_framework_import_test.sh
```diff
@@ -202,7 +202,24 @@ let sharedClass = SharedClass()
...
+        unzip_single_file "test-bin/app/app.ipa" "Payload/app.app/app"
+            | nm -aj - | grep .swiftmodule
+    )
+
+    local swiftmodule
+    local swiftmodules=(
+        app_main.swiftmodule
+        iOSSwiftStaticFramework.swiftmodule/x86_64.swiftmodule
+    )
+    for swiftmodule in "${swiftmodule[@]}"; do
```
sergiocampama: `swiftmodules[@]`

kastiglione: 🤦‍♂

test/ios_framework_import_test.sh
```diff
@@ -202,7 +202,24 @@ let sharedClass = SharedClass()
...
 
-    do_build ios //app:app || fail "Should build"
+    do_build ios --compilation_mode=dbg //app:app || fail "Should build"
+
+    local symbols=$(
+        unzip_single_file "test-bin/app/app.ipa" "Payload/app.app/app"
+            | nm -aj - | grep .swiftmodule
+    )
+
+    local swiftmodule
```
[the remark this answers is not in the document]
kastiglione: Yes, without declaring it local it's defaults to being a global variable that escapes this function's scope. Demonstration: https://gist.github.com/kastiglione/ad86cc84d810652cb68ba07218a5464d

I should move this declaration next to the loop

test/ios_framework_import_test.sh
kastiglione: thanks for catching this

kastiglione:
Fixed

kastiglione:
Do you want me to rebase again?

kastiglione:
Rebased.

kastiglione:
Rebased again.

kastiglione:
@thomasvl would you mind looking at this? I think everything is good and covered here. I've rebased against master just now. I can squash if you'd like.

We've been using this as a patch for weeks now and it's been working as expected.

kastiglione:
Happy to answer any questions, add more comments or docs etc.

apple/internal/apple_framework_import.bzl
```diff
@@ -222,11 +260,22 @@ def _apple_static_framework_import_impl(ctx):
...
+    if swiftmodule_imports:
+        toolchain = ctx.attr._toolchain[SwiftToolchainInfo]
+        providers.append(SwiftUsageInfo(toolchain = toolchain))
+
+        if _is_debugging(ctx):
+            cpu = ctx.fragments.apple.single_arch_cpu
+            swiftmodule = _swiftmodule_for_cpu(swiftmodule_imports, cpu)
+            if not swiftmodule:
+                fail("ERROR: Missing imported swiftmodule for {}".format(cpu))
+            objc_provider_fields.update(_ensure_swiftmodule_is_embedded(swiftmodule))
```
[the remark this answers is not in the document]
kastiglione: I think yes, this could match a `.swiftinterface`. I'll fix

kastiglione:
> I'm guessing you have some partial-Bazel build setup where some of your libraries are built outside of Bazel and then imported into your workspace, which is why you need this unusual support, and if that's the case a flag is probably the most palatable thing to do here. But I want to verify that first.

We have have a few things built outside of our main workspace, and imported as prebuilt components. Some are built with bazel, and others are not. Sooner or later we'll migrate those that are not yet built with bazel. Once we're there, we can either continue to prebuild, and try to switch over to `.swiftinterface`, or skip the prebuild and merge those libraries into a single build. It's unclear yet what we'd like to do there. Both are improvements, but we're not yet there.

For the flag, are you thinking a feature? I'm happy to add that.

kastiglione:
Let me know if I've given enough info, happy to elaborate.

allevato:
That lines up with pretty much what I was expecting, based on my knowledge of your use of the rules.

> For the flag, are you thinking a feature?

Features can be toggled on a per-target/package basis, but I was thinking of something even more aggressive, like one of the new [Starlark build configuration](https://docs.bazel.build/versions/master/skylark/config.html) flags, which would have to be enabled for the entire build. Something like `--@build_bazel_rules_apple//apple:allow_swiftmodules_in_framework_imports` that you'd have to stash in your .bazelrc and is suitably long and descriptive enough to explain what's going on.

(Right now we use `--define` flags for build-global stuff, but that was to get around the lack of Starlark-defined custom command line flags until recently. Now that SBC is a thing, we need to migrate them all over.)

But that's not something you need to add in this PR, because this rule before this PR _already_ allowed .swiftmodules in those frameworks, and you're just totally reasonable adding debugging support for those. I'm just thinking longer term about something we should do to require folks to acknowledge and opt-in to the inherent binary incompatibility of this if we have to support it.

apple/internal/apple_framework_import.bzl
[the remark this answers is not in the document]
kastiglione: Fixed in dcbdf4b0bfdb2825e936490b5f05df7a4a1ea895

kastiglione:
The CI failures are relevant, I'll follow up this morning.

kastiglione:
Fixed issue.

Also re-ran `test/testdata/frameworks/update_frameworks.sh` to generate a 5.1 swiftmodule in the test fixtures.

allevato:
> Also re-ran `test/testdata/frameworks/update_frameworks.sh` to generate a 5.1 swiftmodule in the test fixtures.

Having a prebuilt .swiftmodule file in there isn't going to work; it version-locks the test to a specific version of Xcode, and makes it impossible for Buildkite to update Xcode without breaking the test. It also means we can't have Buildkite and Google's internal CI running different versions of Xcode, which is a non-starter.

(And this is why I'm nervous about this whole feature to begin with.)

I don't have a great answer for this. Since you actually need it to be a valid module (because you're building code that depends on it), you can't just have it be a dummy file. @thomasvl, do you have any ideas?

kastiglione:
For what it's worth, master already is in this state.

I could look to generate these fixtures pre- test.

allevato:
Ah, so it is... from the diff on the files changed page I couldn't tell if it was being added or just 100% changed.

I wonder if you could do this as two phases in the test. First, have a `swift_library` fixture that you build and that creates a module, then you copy that module into the test's workspace and write out a scratch framework import target in a BUILD file in the same test area, pointing to that module? If that works, that would guarantee they're always the same compiler.

kastiglione:
Sounds good

kastiglione:
`iOSSwiftStaticFramework.framework` is now dynamically generated from the test. Also rebased.

test/ios_framework_import_test.sh
```diff
@@ -88,6 +88,50 @@ EOF
...
+swift_library(
+    name = "iOSSwiftStaticFrameworkLibrary",
+    module_name = "iOSSwiftStaticFramework",
+    srcs = ["@build_bazel_rules_apple//test/testdata/frameworks:swift_sources"],
+)
+EOF
+
+  do_build ios \
+    --compilation_mode=dbg \
+    --ios_minimum_os=11.0 \
```
kastiglione: It would be nice to have this defined in a single place, any suggestions?

allevato: Which part is "this" referring to, the minimum OS? Why do you want to move it elsewhere, I don't see it being used anywhere else?

kastiglione: Yes the minimum OS.

The min OS has to be compatible with what the app uses, so this min OS is duplication.

Without specifying 11.0 as the min OS, the swiftmodule was defaulting to 13.1, which fails when being consumed by this test's app target, which is set to iOS 11.

allevato: Right, the other usage was hidden by the collapsed part of the diff. (I'll never get used to GitHub's review interface.)

I'd just stash it in a variable at the top of this test file and use it throughout:

```
readonly IOS_VERSION=11.0

...

    minimum_os_version = "$IOS_VERSION",

...

    --ios_minimum_os="$IOS_VERSION" \
```

kastiglione: That's what I was looking for, thanks.

kastiglione:
Thanks for helping us get this shaped up :)

PR #572, opened 2019-08-30
Pass is_test to swift_runtime_linkopts

keith:
rules_swift uses this flag to determine if extra linker options specific
to Swift should be propagated. Previously this worked just because there
is some duplication in which flags are passed. With Xcode 11 a new
directory has been added that we only want to pass to the linker for
tests.


keith:
Seems like the red is an upstream issue

PR #586, opened 2019-09-09
Append when redirecting to $TEST_log

kastiglione:
Ensure that all content sent to `$TEST_log` is present after the test.

### DIFF
--- a/test/apple_shell_testutils.sh
+++ b/test/apple_shell_testutils.sh
@@ -423,8 +423,8 @@ function do_action() {
        "$@" \
   )
 
-  echo "Executing: bazel ${action} ${bazel_options[*]}" > "$TEST_log"
-  bazel "${action}" "${bazel_options[@]}" > "$TEST_log" 2>&1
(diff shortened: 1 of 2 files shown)


test/apple_shell_testutils.sh
```diff
@@ -423,8 +423,8 @@ function do_action() {
        "$@" \
   )
 
-  echo "Executing: bazel ${action} ${bazel_options[*]}" > "$TEST_log"
-  bazel "${action}" "${bazel_options[@]}" > "$TEST_log" 2>&1
+  echo "Executing: bazel ${action} ${bazel_options[*]}" >> "$TEST_log"
+  bazel "${action}" "${bazel_options[@]}" >> "$TEST_log" 2>&1
```
kastiglione: It was here that I noticed this. I'm not sure if all of these should be `>>`, or if one should be `>` to start anew. I defaulted to making all redirects to `$TEST_log` be appending.

sergiocampama: good find! it should be `>>` for all of them

PR #589, opened 2019-09-10
Add error when no swift dylibs are copied

keith:
When Xcode changes how the Swift libraries are handled, we may need to
update rules_apple to be compatible. In the case of Xcode 11 no Swift
libraries were copied, meaning apps would crash only on OS versions
before 12.4. This way we find this issue sooner.


tools/swift_stdlib_tool/swift_stdlib_tool.sh
```diff
@@ -124,4 +124,10 @@ readonly TEMP_DIR=$(mktemp -d "${TMPDIR:-/tmp}/swift_stdlib_tool.XXXXXX")
 trap 'rm -rf "$TEMP_DIR"' EXIT
 
 copy_swift_stdlibs "$TEMP_DIR"
+echo "$TEMP_DIR"
```
sergiocampama: I don't think this is needed

tools/swift_stdlib_tool/swift_stdlib_tool.sh
```diff
@@ -124,4 +124,10 @@ readonly TEMP_DIR=$(mktemp -d "${TMPDIR:-/tmp}/swift_stdlib_tool.XXXXXX")
...
+if ! ls "$TEMP_DIR/libswift"* > /dev/null 2>&1; then
+  echo "error: No swift libraries were copied, please report this on https://github.com/bazelbuild/rules_apple/issues" >&2
```
sergiocampama: can you change this to 

`error: Expected to copy Swift Standard Lib dylibs, but none were copied.`

given this code will run in other contexts, let's avoid having a mechanism for filing issues

keith:
Updated!

keith:
I think I fixed it, the problem was the swift_library target's source is unused so it wasn't linked to the final binary. I'm not sure if we care about this case but in theory this could happen in the wild as well. And this would error on those cases if the final binary ended up not having any Swift symbols in it

PR #606, opened 2019-09-16
Add full workspace name to dependency

keith:
This is necessary if you want to run the tests from another workspace


keith:
Hrm, it looks like this might break these tests? But I'm not sure why?

PR #608, opened 2019-09-17
Use Zip64 in bundles.

swiple-rules-gardener:
Use Zip64 in bundles.


keith:
I found that with python 2 there is a bug where this crashes. https://github.com/lyft/envoy-mobile/pull/155#issuecomment-504246359

thomasvl:
> I found that with python 2 there is a bug where this crashes.

@keith thanks for the heads up, did you find any other solution?

keith:
Nothing besides running this with python 3. We sacrificed some debug info to make the framework smaller for the time being.

PR #612, opened 2019-09-23
Add documentation on bundle size limits.

swiple-rules-gardener:
Add documentation on bundle size limits.

### DIFF
--- a/doc/common_info.md
+++ b/doc/common_info.md
@@ -1,8 +1,19 @@
 # Apple Bazel definitions - Common Information
 
-## Controlling Builds
+## Build limits
 
-Most aspect of your builds should be controlled via the attributes you set on
+According to
+[Apple's documentation](https://help.apple.com/app-store-connect/#/dev611e0a21f),
+iOS and tvOS bundle sizes cannot exceed 4GB in size (limit for watchOS is 75MB).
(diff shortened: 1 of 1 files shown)


doc/common_info.md
```diff
@@ -1,8 +1,19 @@
...
 
-Most aspect of your builds should be controlled via the attributes you set on
+According to
+[Apple's documentation](https://help.apple.com/app-store-connect/#/dev611e0a21f),
+iOS and tvOS bundle sizes cannot exceed 4GB in size (limit for watchOS is 75MB).
+Because of this, rules_apple does not support building zipped archives that
+would be larger than 4GB. If your build outputs would be larger than 4GB (e.g.
+test bundles) you'll need to reduce the number of dependencies to fit this limit
+(e.g. for test bundles, you can split the targets so that output size is smaller
+than 4GB).
```
keith: If this was added because of the discussion here https://github.com/bazelbuild/rules_apple/pull/608 it's a bit misleading since static framework bundles are definitely allowed to be larger than 4gb as long as they don't actually affect the bundle size that much, like if most of that is debug symbols

thomasvl: Valid point, but this tweak was more or the general case (test bundles, app bundles, etc.).

PR #621, opened 2019-10-03
Add support for Swift based ios_static_frameworks and tvos_static_frameworks.

swiple-rules-gardener:
Add support for Swift based ios_static_frameworks and tvos_static_frameworks.


apple/internal/aspects/swift_static_framework_aspect.bzl
```diff
@@ -0,0 +1,152 @@
...
+
+load(
+    "@build_bazel_rules_swift//swift:swift.bzl",
+    "SwiftInfo",
+)
+
+SwiftStaticFrameworkInfo = provider(
+    fields = {
+        "module_name": "The module name for the single swift_library dependency.",
+        "swiftinterfaces": """
```
keith: I think we might want to propagate swiftmodules as well. That's at least required for using `@testable` with frameworks

apple/internal/aspects/swift_static_framework_aspect.bzl
```diff
@@ -0,0 +1,152 @@
...
+            # aspect would have errored out before.
+            if not generated_header:
+                # Headers should be generated equally for all archs, so just take any of them.
+                generated_header = swiftinfo.transitive_generated_headers.to_list()[0]
+
+            swiftdocs[arch] = swiftinfo.transitive_swiftdocs.to_list()[0]
+            swiftinterfaces[arch] = swiftinfo.transitive_swiftinterfaces.to_list()[0]
+
+        # Make sure that all dictionaries contain at least one module before returning the provider.
+        if all([generated_header, swiftdocs, swiftinterfaces]):
```
keith: generated_header isn't required to exist if you set `swift.no_generated_header` because you never depend on the framework from objc, does that end up being required here?

apple/internal/partials/swift_static_framework.bzl
```diff
@@ -0,0 +1,114 @@
...
+    bundle_header = intermediates.file(
+        ctx.actions,
+        ctx.label.name,
+        "{}.h".format(expected_module_name),
+    )
+    file_support.symlink(ctx, generated_header, bundle_header)
+    bundle_files.append((processor.location.bundle, "Headers", depset([bundle_header])))
+
+    modulemap = intermediates.file(ctx.actions, ctx.label.name, "module.modulemap")
+    ctx.actions.write(modulemap, _modulemap_contents(expected_module_name))
```
keith: It would be nice if this wasn't required since it's not needed if you don't depend on this from objc

apple/internal/partials/swift_static_framework.bzl
```diff
@@ -0,0 +1,114 @@
...
+
+    expected_module_name = bundling_support.bundle_name(ctx)
+    if expected_module_name != swift_static_framework_info.module_name:
+        fail("""
+error: Found swift_library with module name {actual} but expected {expected}. Swift static \
+frameworks expect a single swift_library dependency with `module_name` set to the same \
+`bundle_name` as the static framework target.\
+""")
+
+    generated_header = swift_static_framework_info.generated_header
```
keith: It would be nice if this wasn't required since it's not needed if you don't depend on this from objc

apple/internal/partials/swift_static_framework.bzl
```diff
@@ -0,0 +1,114 @@
...
+        file_support.symlink(ctx, swiftdoc, bundle_doc)
+        bundle_files.append((processor.location.bundle, modules_parent, depset([bundle_doc])))
+
+    bundle_header = intermediates.file(
+        ctx.actions,
+        ctx.label.name,
+        "{}.h".format(expected_module_name),
+    )
+    file_support.symlink(ctx, generated_header, bundle_header)
+    bundle_files.append((processor.location.bundle, "Headers", depset([bundle_header])))
```
keith: It would be nice if this wasn't required since it's not needed if you don't depend on this from objc

doc/rules-ios.md
```diff
@@ -1077,6 +1077,28 @@ simultaneously contain simulator and device architectures (that is, you can
...
 
+`ios_static_framework` supports Swift, but there are some constraints:
+
+* `ios_static_framework` with Swift only works with Xcode 11 and above, since
+  the required Swift functionality for module compatibility is available in
+  Swift 5.1.
+* `ios_static_framework` only supports a single direct `swift_library` target
+  that does not depend transitively on any other `swift_library` targets. The
+  Swift compiler expects a framework to contain a single Swift module, and each
+  `swift_library` target is its own module by definition.
```
keith: Shouldn't this framework still be allowed to depend on other modules as long as those are also vendored as other static frameworks?

apple/internal/aspects/swift_static_framework_aspect.bzl
thomasvl: But I thought the interface files have to be used to deal with different compiler versions, i.e. - this shouldn't be included in something that is to be distributed, and that's the point of this rule, no?

doc/rules-ios.md
thomasvl: But this is going to link everything it depends on, which can only be *_library targets, so how would it depend on one?

apple/internal/aspects/swift_static_framework_aspect.bzl
keith: Hrm yea good question, maybe the expectation is that you don't use vendored libraries like this with @testable, so you don't need it.

doc/rules-ios.md
keith: Ah right I forgot this rule works this way. Internally our rule, that looks a lot like this one https://github.com/lyft/envoy-mobile/blob/master/bazel/swift_static_framework.bzl, doesn't fully link everything but does expect the consumer to have the dependent frameworks in their `deps`

apple/internal/aspects/swift_static_framework_aspect.bzl
thomasvl: Testing the code going into this static framework should go into targets that directly depend on the `swift_library`, so `@testable` in those will work.

If one wanted to write tests on the distributed binary this makes, it should be testing the public interfaces and not going to internal details.

keith: I guess it depends on your use case. We currently distribute binary frameworks from libraries internally that have mock code baked in and internal, in those cases we do import the frameworks with `@testable`

apple/internal/aspects/swift_static_framework_aspect.bzl
sergiocampama: removed the headers being required

apple/internal/partials/swift_static_framework.bzl
sergiocampama: removed modulemaps being required, they will be generated only if the header exists

apple/internal/partials/swift_static_framework.bzl
sergiocampama: removed this requirement

apple/internal/partials/swift_static_framework.bzl
sergiocampama: removed this requirement

doc/rules-ios.md
sergiocampama: in theory we could do this with `avoid_deps` but support for that is not trivial as we'd need to subtract the correct transitive `swiftinterface` files from the provider.

thii: @sergiocampama Could you let us know what needs to be done to support `avoid_deps`? This limitation is kind of blocking us from updating to the new `rules_apple` version.

Issue #605, opened 2019-09-13
`apple_dynamic_framework_import` doesn't strip unnecessary architectures


keith:
Currently if you depend on an `apple_dynamic_framework_import` rule, and you try and build a release app and upload it to the App Store, it gets rejected because the framework has architectures for the simulator that cannot be bundled with a release build. The error looks something like this:

```
The main bundle Mach-O contains arm64(bitcode and machine code), while the nested bundle Mach-O contains X86_64ALL(machine code).
```

To fix this we need to strip out the unused architectures when copying the framework.

keith:
Input from @sergiocampama on a possibly implementation option:

>>>the framework import partial could be expanded to accept the app's binary (or the watchOS extension binary for that platform), and it could register an action to replace the framework binary with a new binary which is lipo'ed to only contain the given binary architectures

keith:
We'll have to make sure the architectures from the app's extensions are included as well, likely right now those would only ever overlap so it shouldn't be an issue

PR #623, opened 2019-10-05
Modify script to strip out unneeded slices from dynamic frameworks

nblievers:
This change removes unnecessary slices from embedded dynamic frameworks. This saves space in the final archive, and will also solve this issue: https://github.com/bazelbuild/rules_apple/issues/605

### DIFF
--- a/apple/internal/partials/framework_import.bzl
+++ b/apple/internal/partials/framework_import.bzl
@@ -27,6 +27,14 @@ load(
     "bundle_paths",
 )
 load(
+    "@build_bazel_rules_apple//apple/internal:intermediates.bzl",
+    "intermediates",
+)
+load(
+    "@build_bazel_rules_apple//apple/internal:outputs.bzl",
+    "outputs",
+)
+load(
     "@bazel_skylib//lib:partial.bzl",
     "partial",
 )
@@ -35,7 +43,7 @@ load(
     "paths",
 )
 
-def _framework_import_partial_impl(ctx, targets, targets_to_avoid):
+def _framework_import_partial_impl(ctx, targets, targets_to_avoid, extra_binaries):
     """Implementation for the framework import file processing partial."""
     _ignored = [ctx]
 
@@ -60,6 +68,8 @@ def _framework_import_partial_impl(ctx, targets, targets_to_avoid):
             files_to_bundle = [x for x in files_to_bundle if x not in avoid_files]
 
     bundle_files = []
+    slicer_args = []
+    main_binary = outputs.binary(ctx)
     for file in files_to_bundle:
         framework_path = bundle_paths.farthest_parent(file.short_path, "framework")
         framework_relative_path = paths.relativize(file.short_path, framework_path)
@@ -69,13 +79,38 @@ def _framework_import_partial_impl(ctx, targets, targets_to_avoid):
         if framework_relative_dir:
             parent_dir = paths.join(parent_dir, framework_relative_dir)
 
-        bundle_files.append(
-            (processor.location.framework, parent_dir, depset([file])),
-        )
+        # check to see if the the parent is "Foo.[extension]" and the file is "Foo", thus "Foo.framework/Foo", so the binary within the framework.
+        if paths.replace_extension(parent_dir, "") == file.basename:
+            stripped = intermediates.file(
+                ctx.actions,
+                ctx.label.name,
+                paths.join("_imported_frameworks", file.basename),
+            )
+            bundle_files.append(
+                (processor.location.framework, parent_dir, depset([stripped])),
+            )
+
+            args = slicer_args + ["--in", file.path, "--out", stripped.path]
+            all_binaries = extra_binaries + [main_binary]
+            for binary in all_binaries:
+                args.append(binary.path)
+
+            ctx.actions.run(
+                inputs = [file] + all_binaries,
+                tools = [ctx.executable._realpath],
+                executable = ctx.executable._dynamic_framework_slicer,
+                outputs = [stripped],
+                arguments = args,
+                mnemonic = "DynamicFrameworkSlicer",
+            )
+        else:
+            bundle_files.append(
+                (processor.location.framework, parent_dir, depset([file])),
+            )
 
     return struct(bundle_files = bundle_files)
 
-def framework_import_partial(targets, targets_to_avoid = []):
+def framework_import_partial(targets, targets_to_avoid = [], extra_binaries = []):
     """Constructor for the framework import file processing partial.
 
     This partial propagates framework import file bundle locations. The files are collected through
@@ -85,6 +120,8 @@ def framework_import_partial(targets, targets_to_avoid = []):
         targets: The list of targets through which to collect the framework import files.
         targets_to_avoid: The list of targets that may already be bundling some of the frameworks,
             to be used when deduplicating frameworks already bundled.
+        extra_binaries: Extra binaries to consider when collecting which archs should be
+            preserved in the imported dynamic frameworks.
 
     Returns:
         A partial that returns the bundle location of the framework import files.
@@ -93,4 +130,5 @@ def framework_import_partial(targets, targets_to_avoid = []):
         _framework_import_partial_impl,
         targets = targets,
         targets_to_avoid = targets_to_avoid,
+        extra_binaries = extra_binaries,
     )
--- a/tools/dynamic_framework_slicer/README
+++ b/tools/dynamic_framework_slicer/README
@@ -1,0 +1,5 @@
+dynamic_framework_slicer takes a directory containing .app bundle structures,
+iterates over the frameworks in the .app, and ensures that each framework only
+contains the same slices as the app binary.
+
+dynamic_framework_slicer can run on Darwin only.
--- a/tools/dynamic_framework_slicer/dynamic_framework_slicer.sh
+++ b/tools/dynamic_framework_slicer/dynamic_framework_slicer.sh
@@ -1,0 +1,88 @@
+#!/bin/bash
+#
+# Copyright 2019 The Bazel Authors. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# dynamic_framework_slicer copies the required dylibs for the given
+# binaries into the specified output path. This tool will also remove any
+# architecture slices which are not used by the given binaries, reducing binary
+# size.
+#
+# This script only runs on darwin and you must have Xcode installed.
+#
+# --out <file>      - the filename where the resultant binary should be placed.
+# --in <file>       - the filename of the input binary
+# <positional_args> - any positional argument is considered to be a
+#                              path to a binary to be processed
+
+set -euo pipefail
+
+declare -a BINARIES
+IN=""
+OUT=""
+while [[ $# -gt 0 ]]; do
+  arg="$1"
+  case $arg in
+      --in)
+      readonly IN="$2"
+      shift
+      ;;
+      --out)
+      readonly OUT="$2"
+      shift
+      ;;
+      *)
+      BINARIES+=("$1")
+      ;;
+  esac
+  shift
+done
+
+if [[ -z "$IN" || -z "$OUT" || "${#BINARIES[@]}" -eq 0 ]]; then
+    echo "Usage: $0 --in path/to/infile --out path/to/outfile binaries..."
+    exit 1
+fi
+
+if [[ ! -f "$IN" ]]; then
+    echo "Expected regular file with --in"
+    exit 1
+fi
+
+# Strip out any unnecessary slices from embedded dynamic frameworks to save space
+
+# Gather all binary slices
+declare -a all_bin_slices
+all_bin_slices=$(xcrun lipo -info "${BINARIES[@]}" | cut -d: -f3 | awk '{ for(i = 1; i <= NF; i++) { print $0; } }' | sort -u)
+
+# Gather the slices in the framework
+declare -a framework_slices
+framework_slices=$(xcrun lipo -info "$IN" | cut -d: -f3 | awk '{ for(i = 1; i <= NF; i++) { print $0; } }' | sort -u)
+
+if [[ $(echo -n $framework_slices | wc -w) -eq 1 || "$all_bin_slices" == "$framework_slices" ]]; then
+    # If we only have one slice or the slices match exactly, we don't need to do anything
+    cp "$IN" "$OUT"
+else
+    # Figure out what we should strip
+    for slice in $framework_slices; do
+        if echo "$all_bin_slices" | grep -q "$slice" ; then
+            slices_needed+=($slice)
+        fi
+    done
+
+    declare -a lipo_args
+    for slice in $all_bin_slices; do
+        lipo_args+=(-extract $slice)
+    done
+    xcrun lipo "$IN" "${lipo_args[@]}" -output "$OUT"
+fi
--- a/apple/internal/ios_rules.bzl
+++ b/apple/internal/ios_rules.bzl
@@ -60,6 +60,7 @@ load(
 )
 load(
     "@build_bazel_rules_apple//apple:providers.bzl",
+    "AppleBundleInfo",
     "IosApplicationBundleInfo",
     "IosExtensionBundleInfo",
     "IosFrameworkBundleInfo",
@@ -123,6 +124,7 @@ def _ios_application_impl(ctx):
         ),
         partials.framework_import_partial(
             targets = ctx.attr.deps + ctx.attr.extensions + ctx.attr.frameworks,
+            extra_binaries = [x[AppleBundleInfo].binary for x in ctx.attr.extensions],
         ),
         partials.resources_partial(
             bundle_id = bundle_id,
--- a/apple/internal/rule_factory.bzl
+++ b/apple/internal/rule_factory.bzl
@@ -171,6 +171,11 @@ _COMMON_PRIVATE_TOOL_ATTRS = dicts.add(
             executable = True,
             default = Label("@build_bazel_rules_apple//tools/xctoolrunner"),
         ),
+        "_dynamic_framework_slicer": attr.label(
+            cfg = "host",
+            executable = True,
+            default = Label("@build_bazel_rules_apple//tools/dynamic_framework_slicer"),
+        ),
     },
     apple_support.action_required_attrs(),
 )
--- a/apple/internal/tvos_rules.bzl
+++ b/apple/internal/tvos_rules.bzl
@@ -52,6 +52,7 @@ load(
 )
 load(
     "@build_bazel_rules_apple//apple:providers.bzl",
+    "AppleBundleInfo",
     "TvosApplicationBundleInfo",
     "TvosExtensionBundleInfo",
     "TvosFrameworkBundleInfo",
@@ -101,6 +102,7 @@ def _tvos_application_impl(ctx):
         ),
         partials.framework_import_partial(
             targets = ctx.attr.deps + embeddable_targets,
+            extra_binaries = [x[AppleBundleInfo].binary for x in ctx.attr.extensions],
         ),
         partials.resources_partial(
             bundle_id = bundle_id,
--- a/tools/BUILD
+++ b/tools/BUILD
@@ -17,6 +17,7 @@ filegroup(
         "//tools/whitelists/function_transition_whitelist:for_bazel_tests",
         "//tools/wrapper_common:for_bazel_tests",
         "//tools/xctoolrunner:for_bazel_tests",
+        "//tools/dynamic_framework_slicer:for_bazel_tests",
     ],
     visibility = [
         "//:__pkg__",
--- a/tools/dynamic_framework_slicer/BUILD
+++ b/tools/dynamic_framework_slicer/BUILD
@@ -1,0 +1,20 @@
+licenses(["notice"])
+
+sh_binary(
+    name = "dynamic_framework_slicer",
+    srcs = ["dynamic_framework_slicer.sh"],
+    # Used by the rule implementations, so it needs to be public; but
+    # should be considered an implementation detail of the rules and
+    # not used by other things.
+    visibility = ["//visibility:public"],
+)
+
+# Consumed by bazel tests.
+filegroup(
+    name = "for_bazel_tests",
+    testonly = 1,
+    srcs = glob(["**"]),
+    visibility = [
+        "//tools:__pkg__",
+    ],
+)


kastiglione:
cc @supermarin

tools/bundletool/process_and_sign.sh.template
```diff
@@ -30,6 +30,40 @@ if [[ -n "$POST_PROCESSOR" ]]; then
   "$POST_PROCESSOR" "$WORK_DIR"
 fi
 
+# Strip out any unnecessary slices from embedded dynamic frameworks to save space
+for i in "$WORK_DIR"/*.app; do
+    # Lets figure out what slices the app has
+    app_name=$(basename $i .app)
+    slices=$(xcrun lipo -info $i/$app_name | cut -d: -f3)
```
keith:
```suggestion
    slices=$(xcrun lipo -info "$i/$app_name" | cut -d: -f3)
```

keith: I think `app_slices` would be more clear here since there's also `framework_slices` below

tools/bundletool/process_and_sign.sh.template
```diff
@@ -30,6 +30,40 @@ if [[ -n "$POST_PROCESSOR" ]]; then
   "$POST_PROCESSOR" "$WORK_DIR"
 fi
 
+# Strip out any unnecessary slices from embedded dynamic frameworks to save space
+for i in "$WORK_DIR"/*.app; do
+    # Lets figure out what slices the app has
+    app_name=$(basename $i .app)
```
keith:
```suggestion
    app_name=$(basename "$i" .app)
```

tools/bundletool/process_and_sign.sh.template
```diff
@@ -30,6 +30,40 @@ if [[ -n "$POST_PROCESSOR" ]]; then
...
+
+    # Now lets make sure any included frameworks match
+    for f in $i/Frameworks/*.framework; do
+        # We need to check we have overlapping architectures
+        framework_name=$(basename $f .framework)
+        framework_slices=$(xcrun lipo -info $f/$framework_name | cut -d: -f3)
+        # Sanity check that the framework has at least the slices the app has
+        for s in $slices ; do
+            echo $framework_slices | grep -q $s
+            if [ $? -ne 0 ]; then
```
keith:
```suggestion
            if ! echo "$framework_slices" | grep -q "$s"; then
```

tools/bundletool/process_and_sign.sh.template
```diff
@@ -30,6 +30,40 @@ if [[ -n "$POST_PROCESSOR" ]]; then
...
+                echo "App has slice $s, but $f does not. Aborting..."
+                exit -1
```
keith:
```suggestion
                exit 1
```

tools/bundletool/process_and_sign.sh.template
```diff
@@ -30,6 +30,40 @@ if [[ -n "$POST_PROCESSOR" ]]; then
...
+    # Now lets make sure any included frameworks match
+    for f in $i/Frameworks/*.framework; do
+        # We need to check we have overlapping architectures
+        framework_name=$(basename $f .framework)
+        framework_slices=$(xcrun lipo -info $f/$framework_name | cut -d: -f3)
+        # Sanity check that the framework has at least the slices the app has
+        for s in $slices ; do
+            echo $framework_slices | grep -q $s
+            if [ $? -ne 0 ]; then
+                echo "App has slice $s, but $f does not. Aborting..."
```
keith:
```suggestion
                echo "error: $i has slice $s, but $f does not" >&2
```

tools/bundletool/process_and_sign.sh.template
```diff
@@ -30,6 +30,40 @@ if [[ -n "$POST_PROCESSOR" ]]; then
...
+        # Now make the framework match the slices the app has
+        for s in $slices ; do
+            lipo_args="$lipo_args -extract $s"
+        done
+        xcrun lipo $f/$framework_name $lipo_args -output $temp_file
+        if [ $? -ne 0 ]; then
+            echo lipo failed, aborting...
+            exit 1
+        fi
+        file $temp_file
```
keith: Left over from debugging?

tools/bundletool/process_and_sign.sh.template
```diff
@@ -30,6 +30,40 @@ if [[ -n "$POST_PROCESSOR" ]]; then
...
+            fi
+        done
+
+        # Now make the framework match the slices the app has
+        for s in $slices ; do
+            lipo_args="$lipo_args -extract $s"
+        done
+        xcrun lipo $f/$framework_name $lipo_args -output $temp_file
+        if [ $? -ne 0 ]; then
+            echo lipo failed, aborting...
```
keith:
```suggestion
            echo "error: lipo of $framework_name failed" >&2
```

tools/bundletool/process_and_sign.sh.template
```diff
@@ -30,6 +30,40 @@ if [[ -n "$POST_PROCESSOR" ]]; then
...
+        # Sanity check that the framework has at least the slices the app has
+        for s in $slices ; do
+            echo $framework_slices | grep -q $s
+            if [ $? -ne 0 ]; then
+                echo "App has slice $s, but $f does not. Aborting..."
+                exit -1
+            fi
+        done
+
+        # Now make the framework match the slices the app has
```
keith: Should we declare `lipo_args=""` here?

tools/bundletool/process_and_sign.sh.template
```diff
@@ -30,6 +30,40 @@ if [[ -n "$POST_PROCESSOR" ]]; then
...
+        for s in $slices ; do
+            lipo_args="$lipo_args -extract $s"
+        done
+        xcrun lipo $f/$framework_name $lipo_args -output $temp_file
```
keith: Looks like `$temp_file` is never declared?

tools/bundletool/process_and_sign.sh.template
```diff
@@ -30,6 +30,40 @@ if [[ -n "$POST_PROCESSOR" ]]; then
...
+            echo $framework_slices | grep -q $s
+            if [ $? -ne 0 ]; then
+                echo "App has slice $s, but $f does not. Aborting..."
+                exit -1
+            fi
+        done
+
+        # Now make the framework match the slices the app has
+        for s in $slices ; do
+            lipo_args="$lipo_args -extract $s"
```
keith: If you do this as an array you can do:

```suggestion
            lipo_args+=(-extract "$s")
```

tools/bundletool/process_and_sign.sh.template
```diff
@@ -30,6 +30,40 @@ if [[ -n "$POST_PROCESSOR" ]]; then
...
+        for s in $slices ; do
+            echo $framework_slices | grep -q $s
+            if [ $? -ne 0 ]; then
+                echo "App has slice $s, but $f does not. Aborting..."
+                exit -1
+            fi
+        done
+
+        # Now make the framework match the slices the app has
+        for s in $slices ; do
```
keith:
```suggestion
        for s in $slices; do
```

tools/bundletool/process_and_sign.sh.template
```diff
@@ -30,6 +30,40 @@ if [[ -n "$POST_PROCESSOR" ]]; then
...
+            lipo_args="$lipo_args -extract $s"
+        done
+        xcrun lipo $f/$framework_name $lipo_args -output $temp_file
+        if [ $? -ne 0 ]; then
+            echo lipo failed, aborting...
+            exit 1
+        fi
+        file $temp_file
+        mv $temp_file $f/$framework_name
```
keith:
```suggestion
        mv "$temp_file" "$f/$framework_name"
```

tools/bundletool/process_and_sign.sh.template
```diff
@@ -30,6 +30,40 @@ if [[ -n "$POST_PROCESSOR" ]]; then
   "$POST_PROCESSOR" "$WORK_DIR"
 fi
 
+# Strip out any unnecessary slices from embedded dynamic frameworks to save space
+for i in "$WORK_DIR"/*.app; do
```
keith:
```suggestion
for app in "$WORK_DIR"/*.app; do
```

I think naming this `app` is a bit more clear

tools/bundletool/process_and_sign.sh.template
```diff
@@ -30,6 +30,40 @@ if [[ -n "$POST_PROCESSOR" ]]; then
...
+    app_name=$(basename $i .app)
+    slices=$(xcrun lipo -info $i/$app_name | cut -d: -f3)
+
+    # Now lets make sure any included frameworks match
+    for f in $i/Frameworks/*.framework; do
+        # We need to check we have overlapping architectures
+        framework_name=$(basename $f .framework)
+        framework_slices=$(xcrun lipo -info $f/$framework_name | cut -d: -f3)
+        # Sanity check that the framework has at least the slices the app has
+        for s in $slices ; do
```
kastiglione: Should this script be responsible for this, or should this fall through to be a linker error?

kastiglione: Or wait, this happens after linking, right? Does this even need to be checked then?

sergiocampama:
thanks for the contribution! 

a few things:

* This script should be specific for applying the post processing and codesigning steps, so we shouldn't be mixing the behavior with framework slicing.
* By putting this logic here, you're going to effectively execute this stripping step for any change in the application, and even for frameworks that will already have the correct architectures built in (i.e. `ios_framework` doesn't need any stripping) and that's going to be a constant perf hit for framework users, even when they don't use `apple_dynamic_framework_import`.

What I recommend doing instead is creating a new tool for framework import slicing, and invoking it inside the `partials/framework_import.bzl` partial. You'll need to:
* Collect all the binary files from all frameworks.
* Collect all binaries for the platform being built (i.e. app executable and extensions)
* Register an action that invokes the new tool with those inputs, and the outputs will be the sliced framework binaries, which you then inject into the files to be bundled (effectively replacing the binary with a sliced binary)

The tool then needs to:
* Inspect all executable binaries (app and extensions) and collect the used architectures.
* Slice each of the framework binaries to only contain those slices.

With this approach, you will only be executing this tool for framework imports, not for `ios_framework` outputs. This also means that the sliced binary will be cached, so that you'll run the slicing tool only when the imported frameworks change, not on every build.

tools/bundletool/process_and_sign.sh.template
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
```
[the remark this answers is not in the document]
nblievers: Fair enough, this is force of habit on my part, as it is better to use /usr/bin/env for cross platform purposes.... but that isn't relevant here!

tools/bundletool/process_and_sign.sh.template
nblievers: Yeah I'm slack with quotes... I will go and quote all the variables, they need it.

tools/bundletool/process_and_sign.sh.template
nblievers: Oops, yep.

tools/bundletool/process_and_sign.sh.template
nblievers: can't hurt.

tools/bundletool/process_and_sign.sh.template
nblievers: Well that's embarrassing.

tools/bundletool/process_and_sign.sh.template
nblievers: Hmm, that is a good point, the linker will fail if there isn't enough slices. This change was really around removing unnecessary slices, I just figured it didn't hurt to be sure.... but you are right, it can't happen without some serious issues up stream.

nblievers:
I've made most of the trivial script changes. Will look at the bigger changes next.

nblievers:
So, for anyone desperate for a solution, this works, at least for phone apps, perhaps not for watch apps. I have made most of the suggested changes (not switching to use an array for lipo args, everything else tho), so it is a no-op in the usual case, and has better quoting etc.

The other approach I am less confident about. As soon as I try and run an external script I get java stack traces, and my appetite to debug java code is small... 

@sergiocampama I can create a script and register it, but when I try to invoke it, I hit java exceptions, which don't seem particularly descriptive.  I am assuming from your description I don't need to add another partial, just extend _framework_import_partial_impl() ?

sergiocampama:
do you have an example PR with adding a tool that shows that error? I have never seen it before

nblievers:
Please see https://github.com/bazelbuild/rules_apple/pull/636 -- glad to know I have managed to hit something unusual! I have no doubt that I am doing the wrong thing... just this type of failure doesn't make it obvious to me what that is. Thanks for taking a look.

sergiocampama:
alright, there was an issue with how you're treating the files... mainly you were declaring that an input to an action was also an output, which is not allowed by Bazel (and hence the cast error from Source to Derived)

I've taken your initial approach and made [this](https://github.com/bazelbuild/rules_apple/commit/5dc5b89543902b6073652b07e804eee1953f0c88) which pipes the files and actions in a way that conforms to my initial comments, what's missing is the core of the script which collects the architectures from the input binaries and slices the framework and writes the output.

after you've made the changes to your branch, please also add tests (also, my code is a bit messy, so make sure to tidy it up a bit as well before sending it for review :) )

sergiocampama:
Also, you'll need to do similar changes for tvos  and watchos in order for those binaries to be correctly processed

nblievers:
That is awesome, thank you. I should be able to work with that and create the iOS, tvOS and watchOS bits required.

nblievers:
Okay the above commit is for iOS only, still need to do watchOS and tvOS.... but it works! I get the expected output for an iOS build. Thanks @sergiocampama for your help.

nblievers:
I added tvOS but not watchOS as it looks like watchOS framework support is a works in progress.

tools/bundletool/process_and_sign.sh.template
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/bash
```
thomasvl: accidental?

tools/dynamic_framework_slicer/README
```diff
@@ -0,0 +1,5 @@
+dynamic_framework_slicer takes a directory containing .app bundle structures,
+iterates over the frameworks in the .app, and ensures that each framework only
+contains the same slices as the app binary.
+
+bundletool can run on Darwin only.
```
thomasvl: bundletool?

tools/dynamic_framework_slicer/dynamic_framework_slicer.sh
```diff
@@ -0,0 +1,54 @@
+#!/bin/bash 
+set -eu
+
+BINARIES+=()
```
thomasvl: shouldn't need `+=` here, just `=`, right?

tools/dynamic_framework_slicer/dynamic_framework_slicer.sh
```diff
@@ -0,0 +1,54 @@
...
+
+# Strip out any unnecessary slices from embedded dynamic frameworks to save space
+
+for bin in $BINARIES; do
+    # Lets figure out what slices the app has
+    bin_slices=$(xcrun lipo -info "$bin" | cut -d: -f3 | sort)
+
+    if [ -f "$IN" ]; then
+        # We need to check we have overlapping architectures
+        framework_slices=$(xcrun lipo -info "$IN" | cut -d: -f3 |sort)
```
thomasvl: space before `sort`

tools/dynamic_framework_slicer/dynamic_framework_slicer.sh
```diff
@@ -0,0 +1,54 @@
...
+      shift
+      ;;
+      *)
+      BINARIES+=("$1")
+      ;;
+  esac
+  shift
+done
+
+# Strip out any unnecessary slices from embedded dynamic frameworks to save space
```
thomasvl: Sanity check args?  Error if missing IN/BINARIES/etc?

i.e. - if `IN` were a directory or a nonexistent file, won't this be a noop, is that good?

tools/dynamic_framework_slicer/dynamic_framework_slicer.sh
```diff
@@ -0,0 +1,54 @@
...
+            continue;
+        fi
+
+        lipo_args=""
+        # Now make the framework match the slices the app has
+        for slice in $bin_slices ; do
+            lipo_args="$lipo_args -extract $slice"
+        done
+        xcrun lipo "$IN" $lipo_args -output "$OUT"
+        if [ $? -ne 0 ]; then
```
thomasvl: Why is this needed, the script is `set -eu`, so won't a failure stop things?

tools/dynamic_framework_slicer/dynamic_framework_slicer.sh
```diff
@@ -0,0 +1,54 @@
...
+    bin_slices=$(xcrun lipo -info "$bin" | cut -d: -f3 | sort)
+
+    if [ -f "$IN" ]; then
+        # We need to check we have overlapping architectures
+        framework_slices=$(xcrun lipo -info "$IN" | cut -d: -f3 |sort)
+        
+        # The normal case should be that the framework and app have exactly the same slices.
+        # Lets check that and skip processing if that is the case
+        if [ "$bin_slices" == "$framework_slices" ]; then
+            continue;
```
thomasvl: Does the Starlark expect to find the framework in the output dir at the end, that doesn't happen in this case, no?  i.e. - if there is nothing to strip, support break?

tools/dynamic_framework_slicer/dynamic_framework_slicer.sh
```diff
@@ -0,0 +1,54 @@
...
+  shift
+done
+
+# Strip out any unnecessary slices from embedded dynamic frameworks to save space
+
+for bin in $BINARIES; do
+    # Lets figure out what slices the app has
+    bin_slices=$(xcrun lipo -info "$bin" | cut -d: -f3 | sort)
+
+    if [ -f "$IN" ]; then
```
thomasvl: use `[[` over `[` since we're in bash, that way it is internal and doesn't spawn `test` (also more consistent within  the file).

apple/internal/partials/framework_import.bzl
```diff
@@ -69,28 +82,52 @@ def _framework_import_partial_impl(ctx, targets, targets_to_avoid):
         if framework_relative_dir:
             parent_dir = paths.join(parent_dir, framework_relative_dir)
 
-        bundle_files.append(
-            (processor.location.framework, parent_dir, depset([file])),
-        )
+        if paths.replace_extension(parent_dir, "") == file.basename:
```
thomasvl: I think I'm missing something, what is this test looking specifically for?

tools/dynamic_framework_slicer/dynamic_framework_slicer.sh
```diff
@@ -0,0 +1,54 @@
...
+  case $arg in
+      --in)
+      readonly IN="$2"
+      shift
+      ;;
+      --out)
+      readonly OUT="$2"
+      shift
+      ;;
+      --realpath)
```
thomasvl: Why do you need realpath?

tools/dynamic_framework_slicer/dynamic_framework_slicer.sh
```diff
@@ -0,0 +1,54 @@
...
+        for slice in $bin_slices ; do
+            lipo_args="$lipo_args -extract $slice"
+        done
+        xcrun lipo "$IN" $lipo_args -output "$OUT"
+        if [ $? -ne 0 ]; then
+            echo "error: lipo of $framework_name failed" >&2
+            exit 1
+        fi
+    fi
+done
```
thomasvl: How does this loop work if there are >1 BINARY to check against?

On any iteration, it will strip unneeded archs, but that isn't correct.  i.e. - 

BINARY1 - archA, archB & archC
BINARY2 - archB & archC
IN = archA, archB, & archC

Loop pass 1 -
  OUT doesn't get created because the arches match
Loop pass 2 -
  OUT gets created with archB & archC

If you make the arches matching case also copy, you still can end up doing extra work on the second loop iteration because you copy arches you already had in OUT.  The better approach is likely going to be to collect all the needed arches, then compare that to the provide ones, and then copy the file or the subset of arches once.

ps - you also need to handle the case where a framework might not have an arch a binary has (because of conditional code), i.e. - BINARY has arm7, arm64, but IN only has arm64.  Currently the script looks like it might try to extract arm7 and thus fail.

tools/dynamic_framework_slicer/README
nblievers: I didn't copy this from anywhere.... *cough* :)

tools/dynamic_framework_slicer/dynamic_framework_slicer.sh
nblievers: Yep that should be fine.

tools/dynamic_framework_slicer/dynamic_framework_slicer.sh
nblievers: I don't know enough to say whether it should be an error or a no-op. I will add some argument checking to ensure the types are what we expect, and that we have arguments, but I'm not entirely sure if there are valid builds where the script will be called with the wrong arguments.

tools/dynamic_framework_slicer/dynamic_framework_slicer.sh
nblievers: Good call.

tools/dynamic_framework_slicer/dynamic_framework_slicer.sh
nblievers: No clue. This was something in the skeleton that @sergiocampama  provided. I am happy to remove if it isn't needed (it obviously isn't used in the script, but that could be a bug?).

tools/dynamic_framework_slicer/dynamic_framework_slicer.sh
nblievers: Surely if the binary has architectures that are missing from the dynamic library, that should fail the build? Actually, the link step should fail in that scenario.

Comment about multiple binaries is good, I will fix that.

apple/internal/partials/framework_import.bzl
nblievers: I'll have to leave that to @sergiocampama  to answer sorry!

tools/dynamic_framework_slicer/dynamic_framework_slicer.sh
thomasvl: Likely just because he copied something quickly to help get things going for you.  If it isn't needed, drop it.

nblievers: Roger. Will remove.

tools/dynamic_framework_slicer/dynamic_framework_slicer.sh
thomasvl: In general, be a little defense just so if someone updates the bzl without updating the script to match, things fail vs. accidentally do nothing.

tools/dynamic_framework_slicer/dynamic_framework_slicer.sh
thomasvl: The app might have conditional code specific to arm64 for some feature (say something AR/VR related) and thus ends up with that framework to provide the support, but arm7 build doesn't have the feature and thus doesn't need the dependency.  Hence the app can have arches the framework doesn't have.

nblievers: Hmm yes good call. Will have to rely on the linker for that scenario, which makes the script simpler. I wont try and validate the minimum set of slices, just remove the unused ones.

nblievers:
I re-wrote the script to handle the suggested case, so aside from the above question I can't answer, hopefully all the concerns are dealt with now.

apple/internal/partials/framework_import.bzl
thomasvl: Talking with @allevato we realized what this was after, it's checking to see if the the parent is "Foo.[extension]" and the file is "Foo", thus "Foo.framework/Foo", so the binary within the framework.

So this is correct for iOS/tvOS, but with the symlinks that happen for macOS frameworks, this would be 100% correct.  So please add a comment here for what it is doing, and that it might need a revisit if macOS ever got multiple arches and and desired this support or some other platform uses a different framework layout compared.

thomasvl:
@keith @kastiglione any more thoughts on this one?

thomasvl:
ps - it would be nice to put a test on this, but in the current setup, these can be some what painful since it needs a built binary to pull in, etc.

thomasvl:
You'll also need to do the CLA steps before we can accept the patch.

tools/dynamic_framework_slicer/dynamic_framework_slicer.sh
```diff
@@ -0,0 +1,80 @@
...
+for slice in $(xcrun lipo -info "$IN" | cut -d: -f3); do
+    framework_slices+=($slice)
+done
+
+# Figure out what, if anything, we should strip
+declare -a slices_needed
+just_copy=1
+for slice in "${framework_slices[@]}"; do
+    if echo "${all_bin_slices[@]}" | grep -q "$slice" ; then
+        slices_needed+=($slice)
```
kastiglione: Will this always end up being `slices_needed` == `all_bin_slices`? I'm wondering if `slices_needed` needs to be computed.

tools/dynamic_framework_slicer/dynamic_framework_slicer.sh
```diff
@@ -0,0 +1,80 @@
...
+    bin_slices=$(xcrun lipo -info "$bin" | cut -d: -f3)
+    
+    # Add the app slices to the overall list of required slices
+    for slice in $bin_slices ; do
+        # store the slice if we don't already have it
+        if [[ "${#all_bin_slices[@]}" -eq 0 || $( echo "${all_bin_slices[@]}" | grep -q "$slice") ]] ; then
+            all_bin_slices+=($slice)
+        fi
+    done
+done
```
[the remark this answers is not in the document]
thomasvl: Check the resolved comments (github ui isn't great for past comments), the framework could have different arches the the apps (including the apps have some arch the framework doesn't have), so the assembly is required.

tools/dynamic_framework_slicer/dynamic_framework_slicer.sh
thomasvl: Check the resolved comments (github ui isn't great for past comments), the framework could have different arches the the apps (including the apps have some arch the framework doesn't have), so the assembly is required.

tools/dynamic_framework_slicer/dynamic_framework_slicer.sh
```diff
@@ -0,0 +1,80 @@
...
+# Figure out what, if anything, we should strip
+declare -a slices_needed
+just_copy=1
+for slice in "${framework_slices[@]}"; do
+    if echo "${all_bin_slices[@]}" | grep -q "$slice" ; then
+        slices_needed+=($slice)
+    else
+        # We have a slice in the framework that does not exist in any binary
+        # so a plain copy wont work
+        just_copy=0
```
kastiglione: If the `otool` method is used, and if `slices_needed` == `all_bin_slices` does hold, then I think this block (calculating `slices_needed` and `just_copy`), and the last block, can be merged and reduced to:

```bash
if [[ "${all_bin_slices[*]}" == "${framework_slices[*]}" ]]; then
    cp "$IN" "$OUT"
else
    declare -a lipo_args
    for slice in "${all_bin_slices[@]}"; do
        lipo_args+=(-extract $slice)
    done
    xcrun lipo "$IN" "${lipo_args[@]}" -output "$OUT"
fi
```

tools/dynamic_framework_slicer/dynamic_framework_slicer.sh
kastiglione: How would the app slice successfully link if the frameworks were missing those same arch slices?

kastiglione: It seems like by this point we know the frameworks have no less slices, but potentially more slices.

thomasvl: The framework isn't needed for those archs.  My example - the app has conditional code for arm64 to support some AR/VR feature; the armv7 binary won't have any calls to the things vended from the framework, so the framework isn't needed.  The app ends up with armv7 and arm64, but the framework only has to have arm64.

tools/dynamic_framework_slicer/dynamic_framework_slicer.sh
thomasvl: @kastiglione this was the original comment thread where we got to this.

tools/dynamic_framework_slicer/dynamic_framework_slicer.sh
kastiglione: Yeah.

Our use case is stripping simulator slices, because App Store requires this. This seems like the common case, but not the only case 😕

nblievers:
AFAICT, I didn't modify the doctoring that the buildifier is complaining about, so I'm reluctant to change that.

@googlebot I signed it!

tools/dynamic_framework_slicer/dynamic_framework_slicer.sh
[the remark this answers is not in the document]
nblievers: I considered that but thought the logic was clearer like this. I initially wrote it with associative arrays, which was better (IMHO), but had forgotten that OSX's default bash is OOOOLD.

nblievers:
I don't know what the issue is with the CLA. From the page:
"It looks like you've already signed this CLA. If you'd like to edit your contact information, you may do so below."

tools/dynamic_framework_slicer/dynamic_framework_slicer.sh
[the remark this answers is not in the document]
nblievers: $(xcrun lipo -info "${BINARIES[@]}" | cut -d: -f3 | awk '{ for(i = 1; i <= NF; i++) { print $0; } }' | sort -u) works, I have switched to that.

tools/dynamic_framework_slicer/dynamic_framework_slicer.sh
nblievers: It isn't quite that simple, as @thomasvl  points out above. It is possible that a binary may have a slice that doesn't exist in the current framework, so we need to figure out what the minimum set of slices, that exist in the current framework, that we want to extract.

I have changed the logic to use something similar to the suggested otool method.

nblievers:
@googlebot I signed it!

apple/internal/partials/framework_import.bzl
```diff
@@ -69,28 +79,54 @@ def _framework_import_partial_impl(ctx, targets, targets_to_avoid):
...
 
     return struct(bundle_files = bundle_files)
 
-def framework_import_partial(targets, targets_to_avoid = []):
+def framework_import_partial(targets, targets_to_avoid = [], extra_binaries = []):
     """Constructor for the framework import file processing partial.
 
     This partial propagates framework import file bundle locations. The files are collected through
     the framework_import_aspect aspect.
-
```
thomasvl: You removed this line which is likely what trips things up, the spec is a blank line before the "Args:" block.

apple/internal/partials/framework_import.bzl
```diff
@@ -69,28 +79,54 @@ def _framework_import_partial_impl(ctx, targets, targets_to_avoid):
...
     Args:
         targets: The list of targets through which to collect the framework import files.
         targets_to_avoid: The list of targets that may already be bundling some of the frameworks,
             to be used when deduplicating frameworks already bundled.
-
```
thomasvl: And keep the blank line before the "Returns:" header also.

tools/dynamic_framework_slicer/dynamic_framework_slicer.sh
```diff
@@ -0,0 +1,61 @@
...
+      shift
+      ;;
+      *)
+      BINARIES+=("$1")
+      ;;
+  esac
+  shift
+done
+
+if [[ -z "$IN" || -z "$OUT" || -z "${BINARIES[*]}" ]]; then
```
thomasvl: actually, the array might need to check the count is >0, i think this expansion will be multiple args otherwise and the test will be invalid.

thomasvl:
Re CLA -

One patch (5dc5b89543902b6073652b07e804eee1953f0c88 ) says:

```
From: =?UTF-8?q?Sergio=20Campam=C3=A1?=
 <kaipi@davg-macpro.roam.corp.google.com>
```

So that appears to be what's tripping things up.  I can override things once the other issues are resolved.

apple/internal/partials/framework_import.bzl
nblievers: Not even sure how I managed that...!

tools/dynamic_framework_slicer/dynamic_framework_slicer.sh
nblievers: Actually multiple args was okay, it was not enough args that failed poorly. It is better now.

nblievers:
Okay I think we should be good now!

thomasvl:
Changes look good, can you do us a favor and squash the commits into one commit?  Then when our process pulls tests this internally and eventually merges it, it will be one commit and not the sequence that got here?  That way you'll still get the credit for the change in the commit history on the project/etc.

nblievers:
So I had some fun with git and I think managed to resolve it... but now a unit test fails, I am pretty sure I didn't introduce something new now, and I'm finding the unit test output a little opaque. Any tips or hints?

thomasvl:
Something happened with the unittests late yesterday, I sent out some emails to see if the machines changed in some what to try and get to the root of it, so just ignore that for now.

thomasvl:
Told buildkite to retry, hopefully it will go green again. 🤞

tools/dynamic_framework_slicer/dynamic_framework_slicer.sh
```diff
@@ -0,0 +1,61 @@
+#!/bin/bash
```
allevato: A couple more changes needed (from our internal checks):

* Can you copy the Apache license header comment from one of the other shell scripts in the `tools` directory and add it to the top of this file under the shebang line?
* Can you remove the trailing space from the end of the shebang line, and from line 55?

tools/dynamic_framework_slicer/dynamic_framework_slicer.sh
```diff
@@ -0,0 +1,88 @@
...
+# binaries into the specified output path. This tool will also remove any
+# architecture slices which are not used by the given binaries, reducing binary
+# size.
+#
+# This script only runs on darwin and you must have Xcode installed.
+#
+# --out <file>      - the filename where the resultant binary should be placed.
+# --in <file>       - the filename of the input binary
+# <positional_args> - any positional argument is considered to be a
+#                              path to a binary to be processed
```
allevato: There is a trailing whitespace at the end of this line now.

tools/dynamic_framework_slicer/dynamic_framework_slicer.sh
```diff
@@ -0,0 +1,88 @@
...
+    # If we only have one slice or the slices match exactly, we don't need to do anything
+    cp "$IN" "$OUT"
+else
+    # Figure out what we should strip
+    for slice in $framework_slices; do
+        if echo "$all_bin_slices" | grep -q "$slice" ; then
+            slices_needed+=($slice)
+        fi
+    done
+
```
allevato: It's not visible in this diff, but looking at the raw file, this line still contains trailing whitespace.

tools/dynamic_framework_slicer/dynamic_framework_slicer.sh
nblievers: heh, doh.

thomasvl:
Can you do one more squash, I think we're good otherwise.

PR #633, opened 2019-10-15
Change experimental bundletool from no-cache to no-remote

kastiglione:
Local caching doesn't have the problem of multiple signing identities, and remote-exec should be disabled for the same reason that remote cache is.


keith:
IIRC I saw something about this not being allowed for local caching since if you update local signing identities it won't cause a rebuild. But maybe that's not something we're worried about?

thomasvl:
> IIRC I saw something about this not being allowed for local caching since if you update local signing identities it won't cause a rebuild. But maybe that's not something we're worried about?

Yea, that would be an issue, but it seems like the common case savings make it worth while.

PR #634, opened 2019-10-15
"Hide" the *_bundle test rules names.

### LINKED ISSUE #624: Consider bazel query when picking rule names
We noticed recently that `kind(ios_unit_test, ...)` will return both `ios_unit_test` and `ios_unit_test_bundle`. We can avoid it by changing the query to `kind("ios_unit_test rule", ...)` but I wonder if the best thing to do is pick names that avoid such collisions, where possible. Since `ios_unit_test_bundle` is internal, it could have a name like `_ios_internal_unit_test_bundle`, etc.

swiple-rules-gardener:
"Hide" the *_bundle test rules names.

They are implementation details, so folks should really use them, this also
ensures a query with `kind(ios_unit_test, ...)` doesn't pick them up since a
query that doesn't end in " rule" is an unanchored pattern can hence can match
even just a leading underscore on the name (_ios_unit_test_bundle).

Fixes #624

### DIFF
--- a/apple/internal/testing/ios_rules.bzl
+++ b/apple/internal/testing/ios_rules.bzl
@@ -59,12 +59,16 @@ def _ios_unit_test_impl(ctx):
         IosXcTestBundleInfo(),
     ]
 
-ios_ui_test_bundle = rule_factory.create_apple_bundling_rule(
+# Declare it with an underscore so it shows up that way in queries.
+_ios_internal_ui_test_bundle = rule_factory.create_apple_bundling_rule(
     implementation = _ios_ui_test_bundle_impl,
     platform_type = str(apple_common.platform_type.ios),
     product_type = apple_product_type.ui_test_bundle,
     doc = "Builds and bundles an iOS UI Test Bundle. Internal target not to be depended upon.",
 )
+
+# Alias to import it.
+ios_internal_ui_test_bundle = _ios_internal_ui_test_bundle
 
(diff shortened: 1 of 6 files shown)


kastiglione:
For my queries, `bazel query` isn't prefix based, it's substring, so I don't believe this would fix #624.

kastiglione:
For example `bazel query 'kind(unit_test, ...)'` matches both `ios_unit_test` and `ios_unit_test_bundle`. This is why I thought something like `_ios_internal_unit_test_bundle` might be needed.

kastiglione:
Thanks for doing this though! 🍻

thomasvl:
Ugh, your right, I forget it is a pattern unless it have the magic ` rule` suffix.  This sucks more…

kastiglione:
> unless it have the magic ` rule` suffix

TIL

thomasvl:
> > unless it have the magic ` rule` suffix
> 
> TIL

If you didn't find it, it is sorta documented: https://docs.bazel.build/versions/master/query.html#kind

Isn't completely clear if it fully is a pattern still that is just anchored or what, but it is _something_.

kastiglione:
ah yeah, I knew that `rule` could be used to be more explicit, but thought you were saying there was more to it even. In any case, yes using `rule` is how we prevent matches against `ios_unit_test_bundle`.

kastiglione:
Thanks!

PR #648, opened 2019-11-01
Add support for specific simulator IDs for iOS tests

keith:
This allows users to pass configurable simulator UDIDs to run tests
with. This allows them to run test jobs on the same simulator without
the overhead of creating one each time.

Another benefit of this is if you want to setup simulators ahead of time for example for UI tests where you may need to set some simulator defaults to not show iOS' onboarding flows

### DIFF
--- a/apple/testing/default_runner/ios_test_runner.bzl
+++ b/apple/testing/default_runner/ios_test_runner.bzl
@@ -14,6 +14,7 @@
 
 """iOS test runner rule."""
 
+load("@bazel_skylib//rules:common_settings.bzl", "BuildSettingInfo")
 load(
     "@build_bazel_rules_apple//apple/testing:apple_test_rules.bzl",
     "AppleTestRunnerInfo",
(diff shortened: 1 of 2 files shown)


keith:
Example usage:

```python
string_flag(
    name = "ios_simulator_id",
    build_setting_default = "",
)

ios_test_runner(
    name = "custom_test_runner",
    simulator_id = "//:ios_simulator_id",
)
```

```
bazel test //... --//:ios_simulator_id="21461F1A-68DD-43E0-A795-D0329CB1E217"
```

I would have put this `string_flag` in rules_apple but it seems that you can not use it from external repos, so `--@build_bazel_rules_apple//apple:ios_simulator_id=...` doesn't work

thomasvl:
Can you file a bug against bazel for rules sets not being able to define these and have them work?  We'll get it raised internally.

keith:
I filed https://github.com/bazelbuild/bazel/issues/10152 about this. I think in the meantime this would probably still be a fine way to handle this, but I guess it depends on how you feel about the public API changing if this starts working because I initial made `simulator_id` a private attr referencing the flag that was in rules_apple

khandpur:
Can you use test arguments for this?

keith:
I've submitted the `--test_arg` option here https://github.com/bazelbuild/rules_apple/pull/653

keith:
#653 is the better approach

Issue #495, opened 2019-07-16
Open source internal test runner


keith:
Through a few conversations with Googlers I learned that internally a different test runner is used with rules_apple than the one that is publicly supported. I think because of this the open source version has a lack of testing which can cause [issues](https://github.com/bazelbuild/rules_apple/issues/490) that we likely wouldn't hit if the community shared the test runner that is currently used by Google. It would be great if that runner could be open sourced so that we could all contribute to it and improve a single tool, rather than being required to maintain our own test runners.

sergiocampama:
This is something outside of the rules_apple project, but given the lack of an appropriate channel to make this request, we can keep it here for tracking.

keith:
One thing I want to add to this is that iterating on / debugging xctestrunner is very difficult, here's the current steps I'm using:

1. Clone [xctestrunner](https://github.com/google/xctestrunner), [subpar](https://github.com/google/subpar), and rule_apple
2. In your local xctestrunner checkout, merge these PRs: https://github.com/google/xctestrunner/pull/7 https://github.com/google/xctestrunner/pull/8
3. In your local subpar clone, change [here](https://github.com/google/subpar/blob/9fae6b63cfeace2e0fb93c9c1ebdc28d3991b16f/subpar.bzl#L89) and add `"--interpreter", "/usr/bin/python2.7",` (this may vary depending on how your project is configured, without this xctestrunner will try to work with python3, and if you change the par_binary rule and add `python_version = "PY2"` it will try to use the `python2` executable which doesn't exist on macOS)
4. Build xctestrunner with `bazel build //:ios_test_runner.par --override_repository=subpar=/path/to/subpar`
5. Copy `bazel-bin/ios_test_runner.par` to your local bazel repo
6. In your repo's root BUILD file add:

```python
filegroup(
    name = "foo",
    srcs = ["ios_test_runner.par"],
    visibility = ["//visibility:public"],
)
```

7. In your local rules_apple clone (or your custom test runner that uses xctestrunner), change all uses of `@xctestrunner//file` to `//:foo`
8. Test your code like normal passing `--override_repository=build_bazel_rules_apple=/path/to/rules_apple`

And repeat steps 4 and 5 whenever you make new changes to xctestrunner.

jerrymarino:
👍 in the meantime, I've forked to [ios-bazel-users](https://github.com/ios-bazel-users/xctestrunner)

PR #653, opened 2019-11-05
Add support for specific simulator IDs for iOS tests

keith:
This allows users to use `--test_arg=--ios_simulator_id=ID` to run tests
on a specific simulator that they setup before running the tests. This
allows them to avoid creating and booting a new simulator on each test
run.

### DIFF
--- a/apple/testing/default_runner/ios_test_runner.template.sh
+++ b/apple/testing/default_runner/ios_test_runner.template.sh
@@ -23,6 +23,21 @@ basename_without_extension() {
   echo "${filename%.*}"
 }
 
+simulator_id=""
+while [[ $# -gt 0 ]]; do
+  arg="$1"
+  case $arg in
+    --destination=platform=ios_simulator,id=*)
+      simulator_id="${arg##*=}"
+      ;;
+    *)
+      echo "error: unsupported argument: '$1'"
+      exit 1
+      ;;
+  esac
+  shift
+done
+
 # Enable verbose output in test runner.
 runner_flags=("-v")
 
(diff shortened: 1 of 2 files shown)


keith:
This is an alternative to https://github.com/bazelbuild/rules_apple/pull/648 based on PR feedback

keith:
@khandpur can you review?

khandpur:
The way we do it is by passing this as args: `--test_arg=--destination=platform=ios_simulator,id={UUID}`

I know this is a pain especially diverging internal and external behavior. I'll chat with our internal apple rules team and see what can be done here.

keith:
I just want to use this opportunity to mention https://github.com/bazelbuild/rules_apple/issues/495 again 😄 

But in this case this test runner only supports iOS, and doesn't support devices which is why I went with the simpler arg. But I can change to parse something like that the easy way for now just to match the internal case and ignore all the other options until those are supported publicly

keith:
@khandpur I've updated to use that long arg even though no other args are supported

keith:
@khandpur can you take another look at this?

apple/testing/default_runner/ios_test_runner.template.sh
```diff
@@ -102,12 +117,23 @@ if [[ -n "${LAUNCH_OPTIONS_JSON_STR}" ]]; then
   runner_flags+=("--launch_options_json_path=${LAUNCH_OPTIONS_JSON_PATH}")
 fi
 
+target_flags=()
+if [[ -n "$simulator_id" ]]; then
+  target_flags=(
+    "test"
```
khandpur: I am not entire sure here because runners are different but should this be "simulator_test" ?

khandpur:
I just reviewed it. I am not too familiar with this particular runner so I am going to add someone else to review the arg change.

apple/testing/default_runner/ios_test_runner.template.sh
keith: I'm not sure why but `xctestrunner` doesn't accept `--id` with `simulator_test`

apple/testing/default_runner/ios_test_runner.template.sh
```diff
@@ -102,12 +117,23 @@ if [[ -n "${LAUNCH_OPTIONS_JSON_STR}" ]]; then
...
+    "--id=$simulator_id"
```
albertdai: Please also add --platform=ios_simulator to the runner. It can help xctestrunner skip the platform check.

apple/testing/default_runner/ios_test_runner.template.sh
albertdai: For the design of xctestrunner, the subcommand `simulator_test` is designed for creating new simulator for testing and the subcommand  `test` is for testing on existing simulator/device with passing id.

apple/testing/default_runner/ios_test_runner.template.sh
keith: Done!

keith:
I rebased, we'll see what happens

khandpur:
Change looks fine pending tests.

ajanuar:
any update?

keith:
I haven't attempted to write tests yet, if you want to take over and do that that'd be awesome!

keith:
Ok got a test here now

PR #656, opened 2019-11-07
Add swiftmodule files to ios_static_framework

keith:
As far as I can tell even with swiftinterface files these are still
required. Xcode still adds them to the framework when using the library
evolution flags and without them you get build errors when including the
framework with just the swiftinterface.

### DIFF
--- a/apple/internal/aspects/swift_static_framework_aspect.bzl
+++ b/apple/internal/aspects/swift_static_framework_aspect.bzl
@@ -25,6 +25,9 @@ SwiftStaticFrameworkInfo = provider(
         "swiftinterfaces": """
 Dictionary of architecture to the generated swiftinterface file for that architecture.
 """,
+        "swiftmodules": """
+Dictionary of architecture to the generated swiftmodule file for that architecture.
+""",
         "swiftdocs": """
 Dictionary of architecture to the generated swiftdoc file for that architecture.
 """,
@@ -95,6 +98,7 @@ single swift_library dependency with no transitive swift_library dependencies.\
         module_name = None
         generated_header = None
         swiftdocs = {}
+        swiftmodules = {}
         swiftinterfaces = {}
         for dep in swiftdeps:
             swiftinfo = dep[SwiftInfo]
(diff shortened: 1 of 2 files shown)


keith:
cc @allevato in case you have a better understanding around not needing this

allevato:
What kind of build errors are you seeing? That doesn't seem right—if the compiler finds a `.swiftinterface` file on the search path for a particular module, it compiles it into the implicit module cache. I've done this for simple module dependencies on the command line in other manual tests (but never in an end-to-end "static framework"/app scenario).

keith:
Here's a repro case with a framework built with Xcode that contains the swiftmodule, it builds fine, then when I remove it, you'll see the failure. Run `./repro.sh` to see.

[swiftinterfacerepro.zip](https://github.com/bazelbuild/rules_apple/files/3821138/swiftinterfacerepro.zip)

allevato:
You're hitting a particular case of a known issue in the Swift compiler. From the [Xcode 11.2 release notes](https://developer.apple.com/documentation/xcode_release_notes/xcode_11_2_release_notes):

> If a module is built with `BUILD_LIBRARIES_FOR_DISTRIBUTION` and contains a public type with the same name as the module itself, clients will fail to import the module. (19481048) (FB5863238)

In your repro case, your `Mapper` module defines a struct named `Mapper`, so you're triggering that bug.

Including the .swiftmodule files in the framework "works around" the issue because the compiler sees those first and finds them, and the serialized module loader doesn't suffer from the same bug. But that's fragile and I don't think we should bundle the .swiftmodule files just for that workaround (because there's no other reason to); if you use the framework with a future (binary-incompatible) version of the compiler fixes the bug, then it would see the differently versioned .swiftmodule file and either (1) ignore it and try the .swiftinterface file or (2) say it's invalid and refuse to load the module at all (I don't remember off the top of my head what the behavior here is).

If (2) happens, you're screwed either way. If (1) happens, then you're screwed unless that future version of the compiler fixes the .swiftinterface same module/type name bug.

Giving a type the same name as its containing module in Swift has always been a bit iffy dating back a few versions, and this bug just proves that there are still ongoing issues, so my recommendation here is to just change the name of your module.

keith:
Do you have any insight into why includes the Swiftmodule files with the frameworks it produces?

Also note I do think using frameworks with `@testable` would be another usecase for including these?

allevato:
> Do you have any insight into why [Xcode?] includes the Swiftmodule files with the frameworks it produces?

No idea. 🤷‍♂  When you invoke the Swift compiler with `-emit-module-interface` it still emits the .swiftmodule and .swiftdoc file (i.e., there's no invocation to my knowledge that *just* emits the .swiftinterface file), so maybe it's a case of Xcode just blindly grabbing all the outputs and bundling them?

> Also note I do think using frameworks with `@testable` would be another usecase for including these?

`@testable` has already been described as a ["hack"](https://forums.swift.org/t/the-future-of-testable/9655/14) by multiple people on the Swift team, so I don't think extending this rule to support that use case (i.e., an attribute meant to be used for direct-from-source builds) is technical debt worth taking on and diverges from our intended use case.

From our point of view, `ios_static_framework` exists solely to create true _distributable_ frameworks, which is one of the reasons that we didn't add any support for Swift until 5.1 when module stability was available. Our long-term plan is to drop this rule entirely in favor of something that produces .xcframework bundles instead, so we'd like to keep the line drawn that the output of these rules doesn't have any artifacts that would break compatibility guarantees or otherwise be irrelevant in that distribution setting.

keith:
Relevant input https://twitter.com/harlanhaskins/status/1196524615213842432 https://twitter.com/slava_pestov/status/1196525447506513925

PR #660, opened 2019-11-11
Handle avoid_deps in ios_static_framework for Swift

keith:
Previously ios_static_framework errored if you had multiple transitive
Swift dependencies. By using avoid_deps you actually create a framework
that doesn't include these transitive dependencies, so this case should
be valid.

### DIFF
--- a/apple/internal/aspects/swift_static_framework_aspect.bzl
+++ b/apple/internal/aspects/swift_static_framework_aspect.bzl
@@ -96,10 +96,20 @@ single swift_library dependency with no transitive swift_library dependencies.\
         generated_header = None
         swiftdocs = {}
         swiftinterfaces = {}
+        avoid_swiftinterfaces = []
+        for dep in ctx.rule.attr.avoid_deps:
+            if SwiftInfo in dep:
+                avoid_swiftinterfaces.extend(dep[SwiftInfo].transitive_swiftinterfaces.to_list())
+
         for dep in swiftdeps:
             swiftinfo = dep[SwiftInfo]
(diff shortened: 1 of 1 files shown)


apple/internal/aspects/swift_static_framework_aspect.bzl
```diff
@@ -129,7 +139,7 @@ swift_library dependency with no transitive swift_library dependencies.\
                     generated_header = swiftinfo.transitive_generated_headers.to_list()[0]
 
             swiftdocs[arch] = swiftinfo.transitive_swiftdocs.to_list()[0]
-            swiftinterfaces[arch] = swiftinfo.transitive_swiftinterfaces.to_list()[0]
+            swiftinterfaces[arch] = propagated_swiftinterfaces[0]
```
keith: Hrm I guess this kinda change has to apply for all things accessed on `swiftinfo`

apple/internal/aspects/swift_static_framework_aspect.bzl
```diff
@@ -122,14 +132,15 @@ swift_library dependency with no transitive swift_library dependencies.\
             # Collect the interface artifacts. Only get the first element from each depset since
             # they should only contain 1. If there are transitive swift_library dependencies, this
             # aspect would have errored out before.
+            # TODO: This has to be fixed before merging
```
keith: this would need a `direct_generated_headers` counterpart I believe, if this approach makes sense

keith:
https://github.com/bazelbuild/rules_apple/pull/1031

PR #671, opened 2019-12-18
Set execution requirements for Starlark rules that use apple_support based on the availability of the selected Xcode.

swiple-rules-gardener:
Set execution requirements for Starlark rules that use apple_support based on the availability of the selected Xcode.

This is part of a series of changes that set the execution requirements for actions generated by rules that require Xcodes. The dynamic scheduler will later be modified to run those actions locally/remotely accordingly, but this does not currently change behavior.

RELNOTES: None.

### DIFF
--- a/apple/internal/utils/legacy_actions.bzl
+++ b/apple/internal/utils/legacy_actions.bzl
@@ -60,7 +60,7 @@ def _kwargs_for_apple_platform(ctx, additional_env = None, **kwargs):
         execution_requirement_dicts.append(original_execution_requirements)
 
     # Add the execution requirements last to avoid clients overriding this value.
-    execution_requirement_dicts.append(apple_support.action_required_execution_requirements())
+    execution_requirement_dicts.append(apple_support.action_required_execution_requirements(ctx))
 
     processed_args["env"] = _add_dicts(*env_dicts)
     processed_args["execution_requirements"] = _add_dicts(*execution_requirement_dicts)
--- a/test/starlark_tests/rules/apple_verification_test.bzl
+++ b/test/starlark_tests/rules/apple_verification_test.bzl
@@ -137,7 +137,7 @@ def _apple_verification_test_impl(ctx):
             test_env["APPLE_TEST_ENV_{}_{}".format(key, num)] = value
 
     return [
-        testing.ExecutionInfo(apple_support.action_required_execution_requirements()),
+        testing.ExecutionInfo(apple_support.action_required_execution_requirements(ctx)),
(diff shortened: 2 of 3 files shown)


apple/internal/utils/legacy_actions.bzl
```diff
@@ -59,8 +59,14 @@ def _kwargs_for_apple_platform(ctx, additional_env = None, **kwargs):
     if original_execution_requirements:
         execution_requirement_dicts.append(original_execution_requirements)
 
+    if xcode_config.availability() == "local":
+        execution_requirement_dicts.append({"no-remote": "1"})
```
keith: Should these be no-remote-exec or is that implied by them being execution requirements?

allevato: `no-remote` implies `no-remote-exec` (see https://github.com/bazelbuild/bazel/blob/2075b000c6aa3ba32a6a2ccd2bf49f4ba5b3b748/src/main/java/com/google/devtools/build/lib/actions/ExecutionRequirements.java#L187-L191).

(Note that there are some Bazel changes that this PR relies on haven't landed yet.)

kastiglione: So `no-remote` doesn't disable remote cache?

allevato: It also implies `no-remote-cache` (see the link above).

keith: Isn't that not what we want for this?

allevato: Perhaps for data integrity reasons the intention is to not populate the remote cache with artifacts built using versions of Xcode that are only available locally, but rather than speculate further, I'll defer to @susinmotion who is the author of this change (and the related work in Bazel).

susinmotion: The idea behind this set of changes is for the dynamic scheduler to run actions locally/remotely based on the availability (local, remote or both) of the Xcode version that the user specifies. 

If a user specifies a local-only version with `--xcode_version`, we want to run all Xcode-related actions locally. Disabling remote caching in this case makes sense to me; I wouldn't expect remote cache hits from an Xcode version that isn't available remotely, and populating a remote cache with locally built artifacts seems dangerous.

The regular xcode_config shape (including the autogenerated host_xcodes) does not set remote/local requirements, regardless of the origin of the Xcode. These are only set through a new xcode_config shape that has dependencies indicating which Xcodes are available remotely and locally. Right now, none of this is turned on... which is a good time to be asking questions!

Does that address your concerns?

PR #680, opened 2019-12-20
Substitute DEVELOPMENT_LANGUAGE variable with `en` in resource bundle's Info.plist

thii:
This supports compiling Info.plist out of the box coming from Xcode.

This is similar with #475 but for `apple_resource_bundle` rule.

### DIFF
--- a/apple/internal/resource_actions/plist.bzl
+++ b/apple/internal/resource_actions/plist.bzl
@@ -119,6 +119,12 @@ def merge_resource_infoplists(ctx, bundle_name, input_files, output_plist):
         "TARGET_NAME": product_name,
     }
 
+    # The generated Info.plists from Xcode's project templates use
+    # DEVELOPMENT_LANGUAGE as the default variable substitution for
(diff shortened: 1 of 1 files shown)


keith:
FWIW we hard coded this in all our plists manually

thii:
We have a lot of them including the ones from external dependencies, and the new ones from Xcode always have this. Asking people to manually modify it every time is not so intuitive IMO.

thii:
We came up with the solution of using a genrule to modify the Info.plist before giving it to apple_resource_bundle for now. But I prefer using as less hacks as I can.

keith:
Yea seems reasonable. Especially since this is set in other places.

keith:
at some point this started happening it seems https://github.com/bazelbuild/rules_apple/blob/2c370272687c62dcf94abb244c9f320e8b6d35f2/apple/internal/resource_actions/plist.bzl#L213-L217

thii:
That's for the root Info plists. This one is still needed.

keith:
Sorry about that, folding mixed me up

PR #681, opened 2019-12-20
Substitute resource bundle's PRODUCT_BUNDLE_IDENTIFIER with a generated bundle ID

### LINKED ISSUE #679: Add bundle_id to apple_resource_bundle rule or support Info.plist variable substitution for PRODUCT_BUNDLE_IDENTIFIER
The `Info.plist` file comes with a resource bundle target out of the box coming from Xcode by default has a `PRODUCT_BUNDLE_IDENTIFIER` variable. Although bundle ID isn't needed for resource bundles (according to Bazel) it would be nice to make this work for the default targets that get generated by Xcode.

thii:
The Info.plist file in resource bundle targets generated by Xcode
contains a PRODUCT_BUNDLE_IDENTIFIER variable even though it doesn't seem
necessary. We generate a bundle ID for each target and substitute this
variable to that, so that we can build resource bundles coming from Xcode
without having to modify the Info.plist file.

Resolves #679.


keith:
IMO unless this value is something super stable for access with the NSBundle API it's nicer to have users set this by relying on other substituted variables like TARGET_NAME which is what we do

thomasvl:
This definitely feels wrong because it puts every bundle into the build.bazel namespace when used.  The chances for collisions also seems like it could be pretty high.

thii:
How about adding a bundle_id attribute?

thii:
Adding a `bundle_id` didn't seem trivial as `apple_resource_bundle` is not a packaging rule and it will only get processed when you build its depender. I modified the bundle id generating logic to derive from the parent target's label.

Users will still be able to set this themselves by hard-coding the bundle id.

acecilia:
A `PRODUCT_BUNDLE_IDENTIFIER` is mandatory in the `Info.plist` of a resource bundle if you want to use an asset catalog using `UIImage?(named name: String, in bundle: Bundle?, compatibleWith traitCollection: UITraitCollection?)`. For some more context see: https://github.com/CocoaPods/CocoaPods/issues/1264

keith:
>  PRODUCT_BUNDLE_IDENTIFIER is mandatory in the Info.plist

This is a bit misleading, what is required is the `CFBundleIdentifier` which you can create differently without having this specific replacement.

I'm super wary about a change like this because these values changing, or being inferred in an unexpected way will lead to runtime crashes. The current recommendation for how to deal with this is use a different replacement to create a stable identifier that you choose, such as by doing this in your plist:

```
<key>CFBundleIdentifier</key>
<string>com.lyft.$(PRODUCT_NAME)</string>
```

segiddins:
I agree that adding this default is potentially dangerous -- but I do think a change adding a `bundle_id` attr to `apple_resource_bundle` is something we want to do (even if it might not be a quick fix)

thii:
We pre-process our plists now so this isn't needed anymore.

PR #682, opened 2019-12-21
Expose test_environment attribute on ios_test_runner

indragiek:
Expose this attribute so that we can supply environment variables that are propagated to XCTest. `test_environment` is already declared on `AppleTestRunnerInfo`, it just needs to be exposed on `ios_test_runner`.


indragiek:
@googlebot I signed it!

keith:
What kind of variables make sense to propagate on the runner itself?

indragiek:
@keith We use it to propagate `DYLD_INSERT_LIBRARIES`, which is required to load the dynamic lib used to enable sanitizers (TSAN, etc.) in the simulator

keith:
Thanks!

keith:
Follow up tests https://github.com/bazelbuild/rules_apple/pull/868

indragiek:
@keith thank you!

PR #683, opened 2019-12-21
Handle test environment values that contain = in ios_test_runner

indragiek:
The existing implementation doesn't correctly handle environment variables like `TSAN_OPTIONS=halt_on_error=1` because the "=" in the value is treated as if it were another key-value pair. The updated logic accounts for this by matching only on the first occurrence of "=" for each env var.


keith:
Thanks! I wish there was a nicer way to do this in bash. I added a test so this won't regress. I'll merge once CI is green

keith:
@googlebot I consent.

indragiek:
@segiddins @keith thank you!

PR #685, opened 2019-12-28
Fix :for_bazel_tests visible from Tulsi aspects

thii:
Fixes https://github.com/bazelbuild/tulsi/issues/120

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -21,5 +21,5 @@ filegroup(
         "@build_bazel_apple_support//:for_bazel_tests",
         "@build_bazel_rules_swift//:for_bazel_tests",
     ],
-    visibility = ["//:__subpackages__"],
+    visibility = ["//visibility:public"],
 )


thii:
This needs to be tested using the latest Bazel from HEAD but CI is using a very old version of Bazel (pre 1.2.0)

```
2019/12/28 08:29:07 Using unreleased version at commit 11deef7582dfeec7a04ee3f7236393d9b8027367
```

thomasvl:
Is this for some integration test?  It sounds like they should just pulled in rule_apple into the testing workspace rather than trying to pull it in via the filegroup?  Long term, we'd love to get off integration tests here and thus remove these filegroups.

allevato:
> Long term, we'd love to get off integration tests here and thus remove these filegroups.

Yeah, increasing reliance on these is a bad idea because they're going to be going away as soon as we can realistically make it happen. @DavidGoldman's recommendation in the Tulsi issue is a better fix, IMO.

But I'm glad this pointed out that apparently our last-green configuration has been incorrect since November, but has gone undetected until now because we didn't check in any features that used newer stuff until recently.

thii:
Submitted https://github.com/bazelbuild/tulsi/pull/121 instead.

PR #688, opened 2020-01-06
Introduce product_tags

kastiglione:
This is a proposal to solve a problem we hit with how tags are propagated in rules_apple, and use of `bazel query`.

We'd like to have tags that apply only to core top level rules, for example `ios_application`, etc. However tags are propagated to the bundles as well. One way to solve this is have some separate tags parameter ("`product_tags`") that is not visible to bundle rules, and passed only to the core rules.

The current motivation is to query targets by tags, and then build them. As is, those results include products, and the underlying binaries and bundles. Passing these to `bazel build` results in bundles for example being built with twice, for different transition configurations. This is slower to build and uses more disk space.

### DIFF
--- a/apple/ios.bzl
+++ b/apple/ios.bzl
@@ -46,12 +46,18 @@ load(
 
 def ios_application(name, **kwargs):
     """Builds and bundles an iOS application."""
+    product_tags = kwargs.pop("product_tags", [])
+
     bundling_args = binary_support.create_binary(
         name,
         str(apple_common.platform_type.ios),
         apple_product_type.application,
         **kwargs
     )
+
+    if product_tags or "tags" in bundling_args:
+        tags = bundling_args.pop("tags", [])
+        kwargs["tags"] = tags + product_tags
 
     _ios_application(
         name = name,


kastiglione:
This RP initially changes just `ios_application`, as an example.

allevato:
> We'd like to have tags that apply only to core top level rules, for example ios_application, etc. However tags are propagated to the bundles as well.

Do you mean that the tags are propagated to the _binaries_ as well? Because the top-level rule _is_ the bundle.

This change will actually be unnecessary soon (for some definition of "soon") because we want to restructure `ios_application` and friends so that instead of being a macro that wraps an `apple_binary` target and an `_ios_application` target, it's just the `ios_application` rule directly and that rule implementation does the linking internally, producing just another `File` to bundle.

If you look at the rule implementations for the other platforms (macos, tvos, watchos), we've already (mostly) done this. The `apple_binary` target is gone and the `*os_application` rule does the linking internally. However, we still create a couple extra targets to wedge in the Swift linkopts and entitlements, and those also inherit the tags (but it sounds like those aren't as problematic for you, since they don't involve building anything). So with the binary out of the picture, this hypothetical attribute wouldn't be necessary—we just haven't done the migration for the iOS rules yet. (IIRC, the introduction of a new configuration transition increased output path lengths in ways that broke some of our Swift compiles for older versions that didn't support `@param` files yet because command lines that were previously under the system limit went over the system limit. That may not be an issue anymore as folks have moved on to newer Xcodes).

So rather than adding an attribute that will be obsolete in the near-ish future, can you temporarily filter your queries in some other way, like excluding the binary targets and only building the bundles?

kastiglione:
> can you temporarily filter your queries in some other way

yes, that's what we're doing now, it's a little less clean than we'd like, but we're not blocked.

thanks for providing all the context!

PR #691, opened 2020-01-14
Add _lcov_merger attribute to test rules

keith:
This is required to use `--combined_report=lcov` with `bazel coverage`

### DIFF
--- a/apple/internal/rule_factory.bzl
+++ b/apple/internal/rule_factory.bzl
@@ -226,6 +226,11 @@ AppleTestRunnerInfo provider.
         default = Label("@bazel_tools//tools/objc:mcov"),
         allow_single_file = True,
     ),
+    "_lcov_merger": attr.label(
+        cfg = "host",
+        executable = True,
+        default = Label("@bazel_tools//tools/test:lcov_merger"),
+    ),
 }
 
 def _common_binary_linking_attrs(rule_descriptor):


allevato:
It's a bit more complicated than this. Based on https://github.com/bazelbuild/bazel/pull/8477, we would also want this to be a late-bound default so that it only applies for coverage-enabled builds; otherwise, regular `bazel {build,test}` invocations also incur the cost of processing the `lcov_merger` dependency.

I wish it was clearer why every test rule needs to have a dependency on this tool instead of it just being something that Bazel handles on its own. I guess because it has to run the tool in the scope of the test action? If this is meant to be a supported path for coverage, there's no documentation about `lcov_merger` or the attribute on Bazel's site yet to indicate that.

Another oddity is that Bazel's definition of `cc_test` (`BazelCcTestRule.java`) is a separate implementation from the one we use in Google, and the Google one **does not** set the `[:$_]lcov_merger` attribute (nor do any of our other test rules), but coverage still works with Blaze. But, as far as I can tell, the [TestActionBuilder.java code](https://github.com/bazelbuild/bazel/blob/3924737e4d6e151e0efa4d9d560296ec1e972628/src/main/java/com/google/devtools/build/lib/analysis/test/TestActionBuilder.java#L272-L276) that references that attribute is the same in both Bazel and Blaze, and we do use `lcov_merger` internally, so it just gets into the build "some other way."

I'd like to understand why that "some other way" wasn't feasible for Bazel, because if we do have to add this for Bazel to support coverage, we'll also need to update our workflows to exclude the attribute from our internal version of the rules since it's not needed (and could potentially conflict with other stuff Blaze is doing).

keith:
We talked offline a bit here and we're going to try and find some way to late bind this attribute, since there doesn't seem to be an API to do this at the moment based on whether or not coverage is enabled.

keith:
Filed an issue for discussion here https://github.com/bazelbuild/bazel/issues/10642

keith:
For now the suggested workaround is to use `--test_env=LCOV_MERGER=/usr/bin/true`

sgammon:
with bazelbuild/bazel#14724 it seems like progress is being made on this

PR #699, opened 2020-02-04
Properly escape spaces when compiling plists

thii:
We just hit this recently when we started to build everything (the app) with `opt` mode, when the plist files are actually compiled. (Those aren't `Info.plist`s)

### DIFF
--- a/apple/internal/resource_actions/plist.bzl
+++ b/apple/internal/resource_actions/plist.bzl
@@ -87,9 +87,9 @@ def compile_plist(ctx, input_file, output_file):
     # execute the version of plutil that takes the file directly. If the file is
     # empty, it will echo an new line and then pipe it into plutil. We do this
     # to handle empty files as plutil doesn't handle them very well.
-    plutil_command = "plutil -convert binary1 -o %s --" % output_file.path
-    complete_command = ("if [[ -s {in_file} ]] ; then {plutil_command} {in_file} ; " +
-                        "elif [[ -f {in_file} ]] ; then echo | {plutil_command} - ; " +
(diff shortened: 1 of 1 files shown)


apple/internal/resource_actions/plist.bzl
```diff
@@ -87,9 +87,9 @@ def compile_plist(ctx, input_file, output_file):
     # execute the version of plutil that takes the file directly. If the file is
     # empty, it will echo an new line and then pipe it into plutil. We do this
     # to handle empty files as plutil doesn't handle them very well.
-    plutil_command = "plutil -convert binary1 -o %s --" % output_file.path
-    complete_command = ("if [[ -s {in_file} ]] ; then {plutil_command} {in_file} ; " +
-                        "elif [[ -f {in_file} ]] ; then echo | {plutil_command} - ; " +
+    plutil_command = "plutil -convert binary1 -o '%s' --" % output_file.path
```
allevato: This will be fine until someone has a filename with a single quote in it, so if we really want to handle degenerate cases...

Let's change this back to `%s`, then load `shell.bzl` from Skylib and use `shell.quote(output_file.path)`, which will handle the space case and also escape any single quotes that might be in the path.

apple/internal/resource_actions/plist.bzl
```diff
@@ -87,9 +87,9 @@ def compile_plist(ctx, input_file, output_file):
...
+    complete_command = ("if [[ -s '{in_file}' ]] ; then {plutil_command} '{in_file}' ; " +
```
allevato: Likewise, remove the single quotes you've added here and pass `shell.quote(input_file.path)` below instead.

thii:
Fixed. Please take a look.

PR #700, opened 2020-02-09
Substitute PRODUCT_BUNDLE_PACKAGE_TYPE with default value

### LINKED ISSUE #643: plisttool failes on unknown variable reference $PRODUCT_BUNDLE_PACKAGE_TYPE
## Description
A new Info.plist variable has been defined recently [here](https://forums.developer.apple.com/thread/123573) and ios_application rule failes on Info.plist that includes it. 

I am new to Apple and rules_apple, so not sure how to workaround it or contribute a fix. But if I figure, I will try to contribute. If I can have a mentor, I am happy to help as well. 

## Environemnt 
```
...
	<string>$(PRODUCT_BUNDLE_PACKAGE_TYPE)</string>
...
```

## Expected behaviour
Build runs without a problem

## Observed error
```
ERROR: In target "//test_app:app"; unknown variable reference "$(PRODUCT_BUNDLE_PACKAGE_TYPE)" while merging plists (key: "CFBundlePackageType", value: "$(PRODUCT_BUNDLE_PACKAGE_TYPE)").
```

BalestraPatrick:
I've been trying to compile a project generated with Xcode 11 with Bazel, and it's currently failing with the following error:
```
ERROR: In target "//MyProject:MyProject"; unknown variable reference "$(PRODUCT_BUNDLE_PACKAGE_TYPE)" while merging plists (key: "CFBundlePackageType", value: "$(PRODUCT_BUNDLE_PACKAGE_TYPE)")
```

I haven't tested this change directly myself (appreciate if anyone has instructions on how to do so), but the logic is similar to the existing substitutions that are in place already. I tried to find some documentation regarding `PRODUCT_BUNDLE_PACKAGE_TYPE` and its possible values, but I couldn't find anything. The value is `APPL` for iOS and Catalyst builds, so I guess it's safe to assume it as a default value.

Should fix #643.

### DIFF
--- a/apple/internal/resource_actions/plist.bzl
+++ b/apple/internal/resource_actions/plist.bzl
@@ -216,6 +216,12 @@ def merge_root_infoplists(
     # Info.plists out of the box coming from Xcode.
     substitutions["DEVELOPMENT_LANGUAGE"] = "en"
 
+    # The generated Info.plists from Xcode's project templates use
+    # PRODUCT_BUNDLE_PACKAGE_TYPE as the default variable substitution for
+    # CFBundlePackageType. We substitute this to `APPL` to support
+    # Info.plists out of the box coming from Xcode 11+.
+    substitutions["PRODUCT_BUNDLE_PACKAGE_TYPE"] = "APPL"
+
     if include_executable_name:
         substitutions["EXECUTABLE_NAME"] = product_name
         forced_plists.append(struct(CFBundleExecutable = product_name))
--- a/doc/common_info.md
+++ b/doc/common_info.md
@@ -347,6 +347,13 @@ replaced with the `-` character (i.e., `Foo Bar` will become `Foo-Bar`).
         <code>bundle_name</code> attribute is provided.</p>
       </td>
     </tr>
+    <tr>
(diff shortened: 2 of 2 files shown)


BalestraPatrick:
@googlebot I signed it!

allevato:
`APPL` is not the only possible value for this key; it depends on the product type of the target. It's documented here: https://developer.apple.com/documentation/bundleresources/information_property_list/cfbundlepackagetype

I think a proper solution would need to determine what the correct value should be for each product type that we support and then add that to the rule descriptors in apple/internal/rule_support.bzl, and then feed that through to the action that generates the Info.plist.

BalestraPatrick:
@allevato That's a good point. I actually just tried to generate an iMessage extension and Xcode replaces the value with `XPC!` which is not a supported value from what the documentation says. I guess I can go through the various `product_type`s defined [here](https://github.com/bazelbuild/rules_apple/blob/2f20f88d85c0fe217cf9a9eadfb8015b3a384dea/apple/internal/apple_product_type.bzl#L86) and check what Xcode's substituted value is.

It's my first experience with `rules_apple` but I'll try to come up with a patch.

keith:
Happy to merge a solution to this that implement's Tony's suggestion, feel free to reopen if you want!

PR #711, opened 2020-02-18
Fix incorrect entitlements validation of application identifier

thii:
This fixes the following error that we encountered when building for
devices:

```
ERROR: BUILD.bazel:9:1: ProcessEntitlementsFiles App_entitlements.entitlements failed (Exit 1)
ERROR: In target "//:App_entitlements"; the entitlements "com.apple.developer.team-identifier" ("GFPYJQXRSN") did not match the provisioning profile's "ApplicationIdentifierPrefix" ("['BW6NM26Y25']").
```

From an Apple's Technical Note, the App ID prefix isn't always equal to
the Team ID.

> Often times the prefix is equal to the Team ID though it isn't always
the case.

https://developer.apple.com/library/archive/technotes/tn2415/_index.html#//apple_ref/doc/uid/DTS40016427-CH1-TP-INSPECTING_DISTRIBUTION_BUILD_ENTITLEMENTS_WHILE_SUBMITTING_AN_APP_IN_XCODE

### DIFF
--- a/tools/plisttool/plisttool.py
+++ b/tools/plisttool/plisttool.py
@@ -1170,6 +1170,9 @@ class EntitlementsTask(PlistToolTask):
       PlistToolError: For any issues found.
     """
     # com.apple.developer.team-identifier vs profile's TeamIdentifier
+    # Not verifying against profile's ApplicationIdentifierPrefix here, because
+    # it isn't always equal to the Team ID.
+    # https://developer.apple.com/library/archive/technotes/tn2415/_index.html#//apple_ref/doc/uid/DTS40016427-CH1-ENTITLEMENTSLIST
     src_team_id = entitlements.get('com.apple.developer.team-identifier')
     if src_team_id:
       key = 'TeamIdentifier'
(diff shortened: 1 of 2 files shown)


thii:
(The current master branch isn't building: https://buildkite.com/bazel/rules-apple-darwin/builds/2229)

BalestraPatrick:
We believe that this is a fix worth merging. Our app and some extensions are using a legacy App ID prefix which don't match the Team ID. We would like to get rid of the warning shown in the original message (we can turn off the error with `entitlements_validation = entitlements_validation_mode.warn`, but we don't want to completely disable the validation to keep some level of safety).

tools/plisttool/plisttool.py
```diff
@@ -1169,16 +1169,15 @@ def _validate_entitlements_against_profile(self, entitlements):
...
-        if src_team_id not in from_profile:
-          self._report(
-              ENTITLEMENTS_TEAM_ID_PROFILE_MISMATCH % (
-                self.target, src_team_id, key, from_profile))
+      key = 'TeamIdentifier'
+      from_profile = self._profile_metadata.get(key, [])
+      if src_team_id not in from_profile:
+        self._report(
+            ENTITLEMENTS_TEAM_ID_PROFILE_MISMATCH % (
+              self.target, src_team_id, key, from_profile)),
```
brentleyjones, line 1180:
```suggestion
              self.target, src_team_id, key, from_profile))
```

brentleyjones (approved):

thii:
cc @rsahara

PR #722, opened 2020-02-28
Fix bundle tools timestamps before 1980

AlessandroPatti:
The archive timestamps is set at 1 Jan 1980 UTC timezone, making zip utilities break in PST timezone because the date would be before 1 Jan 1980.


keith:
Looks like there are a bunch of other places in bazel that set to this timestamp, do you have this same issue with those?

keith:
if anyone has more details on this feel free to re-open!

PR #725, opened 2020-03-10
Support processing Metal files

### LINKED ISSUE #696: Support processing Metal files in apple_resource_bundle
In Xcode, if you have `.metal` files in the `Compile Sources` phase, it will compile them into `.metallib` files before before bundling into the resource bundle.

thii:
This adds support for processing Metal files when they're included in a
`apple_resource_bundle` target. This currently does not support
specifying the output filename of the compiled .metallib -- the Metal
files are always compiled to a single `default.metallib` now. 

Resolves #696

### DIFF
--- a/apple/internal/partials/support/resources_support.bzl
+++ b/apple/internal/partials/support/resources_support.bzl
@@ -222,6 +222,40 @@ def _infoplists(ctx, parent_dir, files):
     else:
         return struct(files = [], infoplists = files.to_list())
 
+def _metals(ctx, parent_dir, files, output_filename = "default.metallib"):
+    """Processes metal files.
+
+    The metal files will be compiled into a Metal library named `default.metallib`.
+
+    Args:
+        ctx: The target's context.
+        parent_dir: The path under which the library should be placed.
+        files: The metal files to process.
+        output_filename: The output .metallib filename.
+
+    Returns:
+        A struct containing a `files` field with tuples as described in processor.bzl.
+    """
+    metallib_path = paths.join(parent_dir or "", output_filename)
+    metallib_file = intermediates.file(
+        ctx.actions,
+        ctx.label.name,
+        metallib_path,
+    )
+    resource_actions.compile_metals(
+        ctx,
+        files.to_list(),
+        metallib_file,
+    )
+
+    return struct(
+        files = [(
+            processor.location.resource,
+            parent_dir,
+            depset(direct = [metallib_file])
+        )],
+    )
+
 def _mlmodels(ctx, parent_dir, files):
     """Processes mlmodel files."""
 
@@ -435,6 +469,7 @@ resources_support = struct(
     asset_catalogs = _asset_catalogs,
     datamodels = _datamodels,
     infoplists = _infoplists,
+    metals = _metals,
     mlmodels = _mlmodels,
     noop = _noop,
     plists_and_strings = _plists_and_strings,
--- a/apple/internal/resource_actions/metals.bzl
+++ b/apple/internal/resource_actions/metals.bzl
@@ -1,0 +1,104 @@
+# Copyright 2020 The Bazel Authors. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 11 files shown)


thii:
I'm not sure how to fix the buildifier errors 😢. They didn't look like they're coming from this PR's changes.

keith:
can you add a dead simple test to  make sure this doesn't regress? I'm worried there will be so  few users that it will be easily breakable

apple/internal/resource_actions/metals.bzl
```diff
@@ -0,0 +1,107 @@
...
+    Args:
+      ctx: The target's rule context.
+      input_files: The input metal files.
+      output_file: The output metallib file.
+    """
+    air_files = []
+    target = _metal_apple_target_triple(ctx)
+
+    # Compile each .metal file into a single .air file
+    for input_metal in input_files:
```
segiddins, line 63: should there be an assertion that `input_files` is not empty?

apple/internal/resource_actions/metals.bzl
```diff
@@ -0,0 +1,107 @@
...
+        air_file = ctx.actions.declare_file(
+            paths.replace_extension(input_metal.basename, ".air"),
+        )
+        air_files.append(air_file)
+
+        args = [
```
segiddins, line 69: use an `args` object, so the paths can be lazily expanded?

apple/internal/resource_actions/metals.bzl
```diff
@@ -0,0 +1,107 @@
...
+            executable = "/usr/bin/xcrun",
+            inputs = [input_metal],
+            outputs = [air_file],
+            arguments = args,
+            mnemonic = "MetalCompile",
+        )
+
+    # Compile .air files into a single .metallib file, which stores the Metal
+    # library
+    args = [
```
segiddins, line 91: use an args object here?

apple/internal/partials/support/resources_support.bzl
```diff
@@ -222,6 +222,37 @@ def _infoplists(ctx, parent_dir, files):
...
+
+    Args:
+        ctx: The target's context.
+        parent_dir: The path under which the library should be placed.
+        files: The metal files to process.
+
+    Returns:
+        A struct containing a `files` field with tuples as described in processor.bzl.
+    """
+    metallib_path = paths.join(parent_dir or "", "default.metallib")
```
segiddins, line 238: is the metallib always named `default` ?

thii: Xcode does that. I'm not sure about the benefits of allowing the user to specify the name.

segiddins: Looks like it can be customized in Xcode via `METAL_LIBRARY_FILE_BASE` (defined in CoreBuildSystem.xcspec, referenced in `Metal Linker.xcspec`)

thii: Allowing to specify an output file name would require a change to the public interface of the rule, which I don’t think we should take easy. I suggest to use the default and hard-coded name for now. What do you think?

segiddins: Maybe make the name a parameter to the metal partial, and have the caller be the one to specify it? Then, presumably, the caller can take ctx into account, and a method can just return default for now, but that way we're set up for it to change later?

thii:
Done. PTAL.

apple/internal/resource_actions/metals.bzl
```diff
@@ -0,0 +1,107 @@
...
+            paths.replace_extension(input_metal.basename, ".air"),
+        )
+        air_files.append(air_file)
+
+        args = ctx.actions.args()
+        args.add("metal")
+        args.add("-c")
+        args.add("-target", target)
+        args.add("-ffast-math")
+        args.add("-o", air_file.path)
```
segiddins, line 78: this can just be `args.add("-o", air_file)`, and the `.path` will be substituted during execution. same on the line below

apple/internal/resource_actions/metals.bzl
```diff
@@ -0,0 +1,107 @@
...
+            arguments = [args],
+            mnemonic = "MetalCompile",
+        )
+
+    # Compile .air files into a single .metallib file, which stores the Metal
+    # library
+    args = ctx.actions.args()
+    args.add("metallib")
+    args.add("-o", output_file.path)
+    args.add_all([
```
segiddins, line 95: `args.add_all(air_files)` should work here, without having to expand `air_files` at all

apple/internal/resource_actions/metals.bzl
thii: Nice.

thii:
Ping for merge :-)

PR #744, opened 2020-04-03
Added support to reference an imported dynamic framework for watchOS in a watchos_extension rule.

swiple-rules-gardener:
Added support to reference an imported dynamic framework for watchOS in a watchos_extension rule.

RELNOTES: Added support to reference apple_dynamic_framework_import through a watchos_extension's dependencies.


apple/internal/watchos_rules.bzl
```diff
@@ -61,7 +61,7 @@ load(
 )
 
 def _watchos_application_impl(ctx):
-    """Experimental implementation of ios_application."""
```
sergiocampama, line 64: 😬🎉

PR #777, opened 2020-05-26
Allow to pass a custom script to the *_test_runner rules

acecilia:
Sometimes you want to modify the call to `xctestrunner`. 
One example: you prefer to use an existing simulator instead of creating a new one every time.

### DIFF
--- a/apple/testing/default_runner/ios_test_runner.bzl
+++ b/apple/testing/default_runner/ios_test_runner.bzl
@@ -42,7 +42,7 @@ def _get_execution_environment(ctx):
 def _ios_test_runner_impl(ctx):
     """Implementation for the ios_test_runner rule."""
     ctx.actions.expand_template(
-        template = ctx.file._test_template,
+        template = ctx.file.test_template,
         output = ctx.outputs.test_runner_template,
         substitutions = _get_template_substitutions(ctx),
     )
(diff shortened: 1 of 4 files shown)


keith:
I don't really think we want the implementation details of this script and its replacements to be public API. I think the recommendation in this case would be to create your own test runner rule instead (which would theoretically start as a copy of one of these .bzl files). But I'm also empathetic that there is some overhead to doing that.

The right solution to the simulator reuse is something like this: https://github.com/bazelbuild/rules_apple/pull/653

What other kinda of things are you interested in overriding this for?

acecilia:
@keith I see your point: the fact that this script (which is an internal implementation detail) is a perfect place for tweaking the rule functionality does not justify making it public API.

On the other side, this seems to be an ok alternative to provide developers with more control over the rule functionality: understanding/editing the existing starlark layer on top of `xctestrunner` is complex/imposible for the average dev (or at least for me), but changing a shell script is doable.

In a way, in this case the starlark layer bumps the complexity quiet a lot for something that ends up being just a command execution.

Thanks for the tip on selecting which simulator to run 👍 

My usecase is: running on CI I do not know the ID of the simulator to use, so what I did (in my own fork) is tweak the `ios_test_runner.template.sh` script so:
a) Using the iOS version and simulator name, obtain the ID of a matching simulator
b) Pass the ID to `xctestrunner`

Let me know what you think. Either way, at this point in time I am not actively looking to get this merged, so I am okey with what you decide

keith:
Given the PR I merged above I think you can update your logic to do something like this:

```
uuid="$(get your simulator ID)"
bazel test --test_arg=--destination=platform=ios_simulator,id=$uuid //...
```

acecilia:
Just a detail for posterity. Better use a constant simulator name instead of a simulator UUID, so the value of `--test_arg` is always the same and bazel does not discard the analysis cache between test invocations: 
```
--test_arg="--destination=platform=ios_simulator,name=$SIMULATOR_NAME"
```

PR #784, opened 2020-05-29
Add -application_extension flag in framework link

jerrymarino:
This causes the linker to validate that any dynamic libraries linked against
are safe for use in application extensions.

Without the flag, the framework won't link.


jerrymarino:
Hmm CLA issue seems email related

PR #785, opened 2020-05-29
Add -application_extension flag in framework link

jerrymarino:
This causes the linker to validate that any dynamic libraries linked against
are safe for use in application extensions.

Without the flag, the library won't link.


jerrymarino:
@googlebot I consent

thii:
According to the `apple_binary` docs, the `extension_safe` is deprecated and will be removed soon—the option can be added via the `linkopts` attribute instead.

keith:
follow up: https://github.com/bazelbuild/rules_apple/pull/865

keith:
This seems like the right use since rules_apple provides this API, but bazel itself doesn't consume it, which is fine.

PR #823, opened 2020-06-26
Fix error message formatting by specifying params

segiddins:
Noticed since `bazel build --nobuild ...` was failing


keith:
It's a race, who will win https://github.com/bazelbuild/rules_apple/pull/752/

keith:
merged mine 😛

PR #827, opened 2020-07-01
Sign simulator bundles with com.apple.security.get-task-allow

brentleyjones:
This is needed for Accessibility Inspector to work.

### DIFF
--- a/apple/internal/entitlement_rules.bzl
+++ b/apple/internal/entitlement_rules.bzl
@@ -323,13 +323,39 @@ def _entitlements_impl(ctx):
     # Only propagate linkopts for simulator builds to embed the entitlements into
     # the binary; for device builds, the entitlements are applied during signing.
     if not is_device:
+        simulator_entitlements = None
+        if _include_debug_entitlements(ctx):
+            simulator_entitlements = ctx.actions.declare_file(
+                "%s.simulator.entitlements" % ctx.label.name,
+            )
+            simulator_control = struct(
+                plists = [],
+                forced_plists = [struct(**{"com.apple.security.get-task-allow": True})],
+                output = simulator_entitlements.path,
+                target = str(ctx.label),
+            )
+            simulator_control_file = _new_entitlements_artifact(ctx, "simulator-plisttool-control")
+            ctx.actions.write(
+                output = simulator_control_file,
+                content = simulator_control.to_json(),
+            )
+            resource_actions.plisttool_action(
+                actions = actions,
+                inputs = [],
+                outputs = [simulator_entitlements],
+                platform_prerequisites = platform_prerequisites,
+                plisttool = ctx.executable._plisttool,
+                control_file = simulator_control_file,
+                mnemonic = "ProcessSimulatorEntitlementsFile",
+            )
+
         return [
             linking_support.sectcreate_objc_provider(
                 "__TEXT",
                 "__entitlements",
                 final_entitlements,
             ),
-            AppleEntitlementsInfo(final_entitlements = final_entitlements),
+            AppleEntitlementsInfo(final_entitlements = simulator_entitlements),
         ]
     else:
         return [
--- a/test/ios_application_test.sh
+++ b/test/ios_application_test.sh
@@ -335,12 +335,14 @@ EOF
 </plist>
 EOF
 
+  create_dump_codesign "//app:app" "Payload/app.app" -d --entitlements :-
+  readonly CODESIGN_OUTPUT="test-bin/app/codesign_output"
+
   if is_device_build ios ; then
     # For device builds, entitlements are in the codesign output.
-    create_dump_codesign "//app:app" "Payload/app.app" -d --entitlements :-
     do_build ios "$@" //app:dump_codesign || fail "Should build"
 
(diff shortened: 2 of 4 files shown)


kastiglione:
Thanks for finding and fixing this!

apple/internal/entitlement_rules.bzl
```diff
@@ -281,13 +281,37 @@ def _entitlements_impl(ctx):
...
+            )
+
         return [
             linking_support.sectcreate_objc_provider(
                 "__TEXT",
                 "__entitlements",
                 final_entitlements,
             ),
-            AppleEntitlementsInfo(final_entitlements = final_entitlements),
+            AppleEntitlementsInfo(final_entitlements = simulator_entitlements),
```
keith, line 314: Do we not need to also propagate the previous `final_entitlements` somehow here?

brentleyjones: I don't believe so. the way Xcode does this is embed these entitlements while linking (which we do right above) and only signed with `com.apple.security.get-task-allow`. `AppleEntitlementsInfo.final_entitlements` is what Bazel signs with, so we only want to propagate `simulator_entitlements` here. We use `final_entitlements` in `sectcreate_objc_provider`, so it doesn't get dropped as unused.

brentleyjones:
Rebased

keith:
looks like the test failures are real and the API you're calling has drifted a bit. would also be nice to cover this in a small test case if possible

brentleyjones:
I'll ping when I've added a test.

thomasvl:
Drive by - Should this replace https://github.com/bazelbuild/rules_apple/blob/master/doc/common_info.md#debugging-entitlement-support-appleadd_debugger_entitlement?  i.e. - it was meant to add this when needed, if you always force it for simulator builds, does the flag need to exist for device builds?  (or should the existing logic that honors that also check if it is a simulator build and force it there?)

brentleyjones:
This is on top of that, since even with "The rules will auto add the entitlement for non optimized builds", which results in linker sections for the simulator, we _also_ need this one flag in the entitlements. This is still gated behind `_include_debug_entitlements`, so we will only do this under the same situations that you had laid out.

thomasvl:
If you nest a device/simulator check in 268-268 and also add this to the force list at that time, you don't need to write out any other files, and the existing action flow should do all the work, no?  Should be a much shorter edit and doesn't require extra actions.

brentleyjones:
This new entitlement can't have anything but thing flag, if i recall correctly. If we sign it with the same entitlement that goes in the linking then the Accessibility Inspector is still broken. I'll confirm this again as I fix this up and add tests.

test/ios_application_test.sh
```diff
@@ -351,12 +353,26 @@ EOF
 
     readonly FILE_TO_CHECK="${TEST_TMPDIR}/dumped_entitlements"
     readonly SHOULD_CONTAIN="${FOR_SIM}"
+
+    # Simulator builds also have entitlements in the codesign output,
+    # but only `com.apple.security.get-task-allow` and nothing else
+    do_build ios "$@" //app:dump_codesign || fail "Should build"
```
brentleyjones, line 359: I couldn't make the do_build on 350 to build this as well. Open to a change that make it only build once though.

test/ios_application_test.sh
```diff
@@ -351,12 +353,26 @@ EOF
...
+      assert_not_contains "<key>keychain-access-groups</key>" "${CODESIGN_OUTPUT}"
+    fi
   fi
 
   if [[ "${SHOULD_CONTAIN}" == "y" ]] ; then
     assert_contains "<key>get-task-allow</key>" "${FILE_TO_CHECK}"
+    assert_contains "<key>keychain-access-groups</key>" "${FILE_TO_CHECK}"
   else
     assert_not_contains "<key>get-task-allow</key>" "${FILE_TO_CHECK}"
+    assert_contains "<key>keychain-access-groups</key>" "${FILE_TO_CHECK}"
```
brentleyjones, line 375: Added these to verify that the existing entitlements are included with the defaults.

test/ios_application_test.sh
```diff
@@ -351,12 +353,26 @@ EOF
...
+    # Simulator builds also have entitlements in the codesign output,
+    # but only `com.apple.security.get-task-allow` and nothing else
+    do_build ios "$@" //app:dump_codesign || fail "Should build"
+
+    if [[ "${SHOULD_CONTAIN}" == "y" ]] ; then
+      assert_contains "<key>com.apple.security.get-task-allow</key>" "${CODESIGN_OUTPUT}"
+      assert_not_contains "<key>keychain-access-groups</key>" "${CODESIGN_OUTPUT}"
+    else
+      assert_not_contains "<key>com.apple.security.get-task-allow</key>" "${CODESIGN_OUTPUT}"
+      assert_not_contains "<key>keychain-access-groups</key>" "${CODESIGN_OUTPUT}"
```
brentleyjones, line 366: This specifically tests that we are only getting the default, and not the full entitlements. This is how Xcode does it, so even if it's not a requirement I would prefer to match that.

brentleyjones (commented):
@keith Tests fixed and augmented to cover this use case.

keith (approved):

PR #828, opened 2020-07-06
Make dSYM binary the same name with the bundle name

thii:
Previously dSYM binaries were created separately for each arch
being built, with the binary name appended with the arch name (for
example, `MyApp.app.dSYM/Contents/Resources/DWARF/MyApp_arm64`). This
was slightly different with Xcode's convention, where it creates a
single universal binary for each dSYM bundle and names the binary after
the bundle name. Some crash analytics services rely on this convention
to symbolicate the crash logs.

This patch changes the debug symbols processing to match Xcode's
convention.

### DIFF
--- a/apple/internal/partials/debug_symbols.bzl
+++ b/apple/internal/partials/debug_symbols.bzl
@@ -15,6 +15,10 @@
 """Partial implementation for debug symbol file processing."""
 
 load(
+    "@build_bazel_apple_support//lib:apple_support.bzl",
+    "apple_support",
+)
+load(
     "@build_bazel_rules_apple//apple/internal:bundling_support.bzl",
     "bundling_support",
 )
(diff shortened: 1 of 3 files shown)


thii:
Would this be accepted if implemented behind a feature flag?

apple/internal/partials/debug_symbols.bzl
```diff
@@ -101,21 +105,41 @@ def _bundle_dsym_files(ctx, debug_provider, bundle_name, bundle_extension = ""):
...
-        outputs.append(output_binary)
-        ctx.actions.symlink(target_file = dsym_binary, output = output_binary)
+        inputs.append(dsym_binary)
+        args.extend([
+            "-arch",
+            arch,
+            dsym_binary.path,
+        ])
+
+    apple_support.run(
```
keith, line 135: in the case that this is only built for a single arch, should we avoid this lipo and do a cp?

keith (commented):
I think this is fine, the benefit here being making it more like xcode, we'll need to note this breaking change in the release notes

apple/internal/partials/debug_symbols.bzl
```diff
@@ -101,21 +105,41 @@ def _bundle_dsym_files(ctx, debug_provider, bundle_name, bundle_extension = ""):
...
+    output_binary = ctx.actions.declare_file(
+        "%s/Contents/Resources/DWARF/%s" % (
+            dsym_bundle_name,
+            bundle_name,
+        ),
+    )
+    outputs = [output_binary]
+
+    inputs = []
+    args = [
```
segiddins, line 117: nit: use a `ctx.actions.args()`, then you don't need to expand the file paths during analysis

segiddins (approved):
looks good to me, with the change to symlink/cp the dwarf file when there's only one input

thii:
Thanks. I've rebased and updated the change to address all comments. Please take another look.

apple/internal/partials/debug_symbols.bzl
```diff
@@ -117,30 +121,50 @@ def _bundle_dsym_files(ctx, debug_provider, bundle_name, bundle_extension = ""):
...
+        # Create a universal binary if there are more than one arch.
+        args = ctx.actions.args()
+        args.add("lipo")
+        args.add("-create")
+        args.add("-output", output_binary)
+
+        inputs = []
+        for arch, arch_outputs in outputs_map_items:
+            dsym_binary = arch_outputs["dsym_binary"]
+            args.add_all("-arch", [arch, dsym_binary])
```
keith, line 157: im surprised this bit works, i would expect all or none to be in the list

keith (approved):

PR #838, opened 2020-07-13
Add support for App Clips

### LINKED ISSUE #821: Support for App Clips
App Clips are similar to App Extensions, except they are full apps: https://developer.apple.com/documentation/app_clips/creating_an_app_clip.

We should have rules for creating App Clips and embedding them in `ios_application` targets.

brentleyjones:
Resolves #821.

The "com.apple.developer.parent-application-identifiers" entitlement isn't automatically added, though it probably could be (https://developer.apple.com/documentation/bundleresources/entitlements/com_apple_developer_parent-application-identifiers).

I haven't added tests yet, because I want to make sure this is the desired approach.

### DIFF
--- a/apple/internal/entitlement_rules.bzl
+++ b/apple/internal/entitlement_rules.bzl
@@ -23,6 +23,10 @@ load(
     "defines",
 )
 load(
+    "@build_bazel_rules_apple//apple/internal:apple_product_type.bzl",
+    "apple_product_type",
+)
+load(
     "@build_bazel_rules_apple//apple/internal:bundling_support.bzl",
     "bundling_support",
 )
@@ -130,6 +134,17 @@ def _include_debug_entitlements(ctx):
         return False
     return True
 
+def _include_app_clip_entitlements(ctx):
+    """Returns a value indicating whether app clip entitlements should be used.
+
+    Args:
+      ctx: The Starlark context.
+
+    Returns:
+      True if the app clip entitlements should be included, otherwise False.
+    """
+    return ctx.attr.product_type == apple_product_type.app_clip
+
 def _extract_signing_info(ctx):
     """Inspects the current context and extracts the signing information.
 
@@ -230,6 +245,9 @@ def _entitlements_impl(ctx):
     if _include_debug_entitlements(ctx):
         get_task_allow = {"get-task-allow": True}
         forced_plists.append(struct(**get_task_allow))
+    if _include_app_clip_entitlements(ctx):
+        app_clip = {"com.apple.developer.on-demand-install-capable": True}
+        forced_plists.append(struct(**app_clip))
 
     inputs = list(plists)
 
@@ -306,6 +324,8 @@ entitlements = rule(
         ),
         # Used to pass the platform type through from the calling rule.
         "platform_type": attr.string(),
+        # Used to pass the product type through from the calling rule.
+        "product_type": attr.string(),
         "provisioning_profile": attr.label(
             allow_single_file = [".mobileprovision", ".provisionprofile"],
         ),
--- a/apple/internal/ios_rules.bzl
+++ b/apple/internal/ios_rules.bzl
@@ -65,6 +65,7 @@ load(
 load(
     "@build_bazel_rules_apple//apple:providers.bzl",
     "IosApplicationBundleInfo",
+    "IosAppClipBundleInfo",
     "IosExtensionBundleInfo",
     "IosFrameworkBundleInfo",
     "IosImessageApplicationBundleInfo",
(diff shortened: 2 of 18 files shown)


apple/internal/ios_rules.bzl
```diff
@@ -186,6 +187,97 @@ def _ios_application_impl(ctx):
...
+    """Experimental implementation of ios_app_clip."""
+    top_level_attrs = [
+        "app_icons",
+        "launch_images",
+        "launch_storyboard",
+        "strings",
+        "resources",
+    ]
+
+    # TODO(kaipi): Replace the debug_outputs_provider with the provider returned from the linking
```
sergiocampama, line 200: I like how it appears as I'm adding newer todos to this code :p

(you can just remove the `(kaipi)` part since I'm not going to be taking a look at it any time soon :) )

apple/internal/outputs.bzl
```diff
@@ -47,6 +51,13 @@ def _archive(ctx):
     # DefaultInfo.
     return ctx.outputs.archive
 
+def _archive_for_embedding(ctx):
+    """Returns a files reference for this target's archive, when being embedded in another target."""
+    if _has_different_embedding_archive(ctx):
+        return ctx.actions.declare_file("%s.embedding.zip" % ctx.label.name)
```
segiddins, line 57: should this respect `is_experimental_tree_artifact_enabled` ?

segiddins: nvm -- see below that it wont be called if tree artifacts are enabled

apple/internal/entitlement_rules.bzl
```diff
@@ -306,6 +324,8 @@ entitlements = rule(
         ),
         # Used to pass the platform type through from the calling rule.
         "platform_type": attr.string(),
+        # Used to pass the procut type through from the calling rule.
```
thii, line 327: Minor typo
```suggestion
        # Used to pass the product type through from the calling rule.
```

apple/internal/ios_rules.bzl
```diff
@@ -186,6 +187,97 @@ def _ios_application_impl(ctx):
...
+            binary_artifact = binary_artifact,
+            debug_outputs_provider = binary_target[apple_common.AppleDebugOutputs],
+            dependency_targets = embeddable_targets,
+            package_bitcode = True,
+        ),
+        partials.clang_rt_dylibs_partial(binary_artifact = binary_artifact),
+        partials.debug_symbols_partial(
+            debug_dependencies = embeddable_targets,
+            debug_outputs_provider = binary_target[apple_common.AppleDebugOutputs],
+            package_symbols = True,
```
thii, line 227: @brentleyjones Have you added support for packaging symbols?

brentleyjones: Not explicitly. I've tried to copy whatever is being done for `ios_application` for the most part, then making sure it's embedded into the application following the same pattern as extensions. I bet I missed some of the later though.

thii: `debug_symbols_partial` doesn’t have a `package_symbols` parameter, that’s why I was wondering.

brentleyjones: Ahh, it did back where I made the patch from 😬. Removed now.

brentleyjones: @thii I was confused. Yes, internally we do have support for packaging symbols.

thii:
We're using this patch internally.

keith:
I haven't dove in to review this, but clearly it's something we should merge. So if you have any feedback plz comment here (I will soon too) and once it's reviewed and rebase we can land! 📈

brentleyjones:
> I haven't added tests yet, because I want to make sure this is the desired approach.

I would love some guidance on how to add tests for this. I haven't written tests for Bazel rules before.

keith:
I think some of the prior art around extensions should have good examples for this https://github.com/bazelbuild/rules_apple/blob/master/test/ios_extension_test.sh

brentleyjones:
I'll rebase and add tests when I get a free moment 😄.

apple/internal/processor.bzl
```diff
@@ -169,7 +171,7 @@ def _is_parent_dir_valid(invalid_top_level_dirs, parent_dir):
             return False
     return True
 
-def _archive_paths(ctx):
+def _archive_paths(ctx, embedding = False):
```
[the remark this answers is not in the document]
brentleyjones: Thanks. `outputs.has_different_embedding_archive(ctx)` was added after, I'll clean this up.

keith:
https://github.com/bazelbuild/rules_apple/pull/916

PR #853, opened 2020-07-28
Make test environment override test rule attr env and runner info

ghvg1313:
Currently test rule will respect test rule's `env` attribute over manual test_env specification, this seems a bit counter-intuitive when setting test environment explicitly like `bazel test --test_env=FOO=BAR` doesn't get what you want


ghvg1313:
@googlebot I signed it!

keith:
I believe this is expected, this mirrors the behavior of defining `copts` on targets as well I think? For example if you have:

```
cc_library(
    name = "counter",
    srcs = ["counter.cc"],
    hdrs = ["counter.h"],
    copts = ["-DFOO"],
)
```

And run:

```
bazel build //:counter --copt=-DBAR -s
```

The invocation has `-DBAR -DFOO`, so if the args were competing the `copts` on `cc_library` would win.

keith:
Related tests which current encode the behavior I'm describing https://github.com/bazelbuild/rules_apple/pull/868

PR #854, opened 2020-07-29
Test missing DEVELOPER_DIR

keith:

PR #855, opened 2020-07-29
Fix setting DEVELOPER_DIR for test runners

keith:
This execution info was ignored because bazel expects XCODE_VERSION_OVERRIDE so if you had a different Xcode version selected globally bazel would use that for tests regardless of what was passed for `--xcode_version`

Also see failed CI here https://github.com/bazelbuild/rules_apple/pull/854

And https://github.com/bazelbuild/bazel/blob/e3e46e39e17f770812872175c704aed8efcab754/src/main/java/com/google/devtools/build/lib/rules/apple/AppleConfiguration.java#L47 https://github.com/bazelbuild/bazel/blob/e3e46e39e17f770812872175c704aed8efcab754/src/main/java/com/google/devtools/build/lib/exec/local/XcodeLocalEnvProvider.java#L94-L98

### DIFF
--- a/apple/testing/default_runner/ios_test_runner.bzl
+++ b/apple/testing/default_runner/ios_test_runner.bzl
@@ -35,7 +35,7 @@ def _get_execution_environment(ctx):
     execution_environment = {}
     xcode_version = str(ctx.attr._xcode_config[apple_common.XcodeVersionConfig].xcode_version())
     if xcode_version:
-        execution_environment["XCODE_VERSION"] = xcode_version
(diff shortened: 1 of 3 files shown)


sergiocampama:
oh, I remember this... we didn't want to expose the OVERRIDE variable for some reason, but I think that doesn't apply any more... this change is not going to work as is, as internally it's using the OVERRIDE variant and then replaced on export. You'll need to ask for it to be updated in the export configs.

keith:
Now that this isn't being pushed out by google I guess we're good to go here

PR #858, opened 2020-08-06
Allow .strings extension for plists

keith:
Strings files are ascii plists, so they can be combined with plists with
no extra effort. This is especially useful for providing
'InfoPlist.strings' files to the Info.plist so you don't have to
duplicate that information.

### DIFF
--- a/apple/internal/macos_binary_support.bzl
+++ b/apple/internal/macos_binary_support.bzl
@@ -99,7 +99,7 @@ macos_binary_infoplist = rule(
         {
             "bundle_id": attr.string(mandatory = False),
             "infoplists": attr.label_list(
-                allow_files = [".plist"],
+                allow_files = [".plist", ".strings"],
                 mandatory = False,
                 allow_empty = True,
             ),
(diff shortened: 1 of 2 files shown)


thomasvl:
Aren't `InfoPlist.strings` files normally in language directories?  Why would you try to merge all those into the `Info.plist`?  How would that even work?

keith:
So the problem is that Apple's automated review process requires some keys to live in the Info.plist, and only being in the InfoPlist.strings file isn't enough. So really the intent with merging them at all is to satisfy that app review requirement without duplicating keys in both files. So really the expected usage here would just be to pass a single InfoPlist.strings file (likely your Base.lproj one). And if you accidentally pass more plisttool will correctly fail because of conflicting keys with different values

keith:
FWIW I could also see a world where you preferred this file format just because it's more readable and writable, but just a thought

keith:
Going to punt on this because it's a bit weird. If folks are interested they can use `ctx.actions.symlink` to get a "valid" file path for this.

Issue #746, opened 2020-04-07
apple_dynamic_framework_import regressions


brentleyjones:
4ec36b79fd35633581f5e1eb51c0bc7aac3f6031 regressed `apple_dynamic_framework_import` in a couple ways:
- ~Generated provisioning profiles no longer work, since they aren't listed as an input into `ImportedDynamicFrameworkProcessor`~ Fixed with #856 
- `ImportedDynamicFrameworkProcessor` needs `execution_requirements = {"no-sandbox": "1"}` when there is a provisioning profile otherwise the sandbox blocks it
- When using `ipa_processor` a framework that is stripped of its code signature is no longer re-signed (in particular for us, when we strip bitcode and Swift symbols from a framework)

liuliu:
Any ideas how to put up a fix for this other than revert? @nglevin

thii:
We’ve just hit an error because of this. We renewed our certificate recently, and the imported frameworks weren’t resigned with the new certificate. It makes it an invalid build because the app and the frameworks were signed with different certificates. This seems like a rare case, but it would be nice to be able to opt out of this feature.

brentleyjones:
@Keith This should stay open. I'll update it to say the first part is fixed.

keith:
Sorry it auto closed

keith:
https://github.com/bazelbuild/rules_apple/pull/862

thii:
@brentleyjones How are you dealing with the third issue?

brentleyjones:
@thii We stopped stripping symbols as part of our IPA Post Processor. We now strip the symbols before they are imported by Bazel.

thii:
I see. I have a patch [here](https://github.com/bazelbuild/rules_apple/pull/881) that while it does not fix the above `ipa_post_processor` issue, it can eliminate the need of doing bitcode stripping in `ipa_post_processor`.

PR #862, opened 2020-08-10
Add no-sandbox to ImportedDynamicFrameworkProcessor

keith:
Partially fixes https://github.com/bazelbuild/rules_apple/issues/746

### DIFF
--- a/apple/internal/partials/framework_import.bzl
+++ b/apple/internal/partials/framework_import.bzl
@@ -165,6 +165,7 @@ def _framework_import_partial_impl(ctx, targets, targets_to_avoid):
             outputs = [framework_zip],
             arguments = [args],
             mnemonic = "ImportedDynamicFrameworkProcessor",
+            execution_requirements = {"no-sandbox": "1"},
         )
 
         bundle_zips.append(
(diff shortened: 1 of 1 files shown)


keith:
@brentleyjones @thii can you test?

apple/internal/partials/framework_import.bzl
```diff
@@ -165,6 +165,7 @@ def _framework_import_partial_impl(ctx, targets, targets_to_avoid):
             outputs = [framework_zip],
             arguments = [args],
             mnemonic = "ImportedDynamicFrameworkProcessor",
+            execution_requirements = {"no-sandbox": "1"},
```
brentleyjones, line 168: This _works_, but we have it conditionally add in `if provisioning_profile:` instead. that way we only jump out of the sandbox when we are signing.

keith: Hrm interesting. I wonder what the ramifications are, if any, around conditionally executing in the sandbox. I've never seen that pattern before.

keith: Updated with that behavior

PR #868, opened 2020-08-11
Add tests for test env inheritance

keith:
Tests can come from a few places:

1. `--test_env`
2. `test_environment` on a test runner (as of https://github.com/bazelbuild/rules_apple/pull/682)
3. `env` on a test target

This validates the order of these is what we expect if the same key
appears in multiple of these places.

### DIFF
--- a/apple/internal/testing/apple_test_rule_support.bzl
+++ b/apple/internal/testing/apple_test_rule_support.bzl
@@ -142,8 +142,8 @@ def _apple_test_rule_impl(ctx, test_type):
     # --test_env and env attribute values, but not the execution environment variables.
     test_environment = dicts.add(
         ctx.configuration.test_env,
-        ctx.attr.env,
         getattr(runner, "test_environment", {}),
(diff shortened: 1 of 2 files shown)


keith:
The order is a bit tbd based on https://github.com/bazelbuild/rules_apple/pull/853

Also publicly until https://github.com/bazelbuild/rules_apple/pull/682 was merged rules_apple didn't have any test runners that provided this behavior. Custom test runners could have implemented this

segiddins:
I'm maybe having a bit of trouble reading the tests -- which order are we going with here?

I'd think command line overrides test bundle overrides runner makes the most sense to me?

keith:
sorry I should have been more clear in the description that the order is what's specified in my 1,2,3 above.

The goal here was to stick to the order of operations of copts as close as possible tho, where command line args come before `copts` on a target (where the runner itself having environment is a bit of an outlier, but I think it's closer to the target level than the command line flags)

PR #869, opened 2020-08-11
Add data attribute to apple_static_framework_import

keith:


apple/internal/apple_framework_import.bzl
```diff
@@ -353,6 +357,16 @@ target.
                 [apple_common.Objc, AppleFrameworkImportInfo],
             ],
         ),
+        "data": attr.label_list(
```
segiddins, line 360: I don't think you need to explicitly list the `data` attribute, since all rules have the attribute by default

keith: You do seem to need it, without this it fails today

segiddins: Weird, it's listed under https://docs.bazel.build/versions/master/be/common-definitions.html#common-attributes 🤷

keith: I noticed that too, but even rules_swift manually specifies it https://github.com/bazelbuild/rules_swift/blob/b255732244abc7e264e0b484271c10bd8999acab/swift/internal/attrs.bzl#L22-L31

keith:
Ended up reverting adding this to apple_dynamic_framework_import because it's a bit weird given that you end up bundling the framework that could also contain resources in itself. I figure this is the more likely use case

thii:
What happens if `data` includes resource `.bundle`s?

keith:
hrm I wonder what _should_ happen in that case. I bet with this change as is they would not be processed at all (which might cause issues) but would be placed in the same place in the bundle that they would if they were in the correct place.

thii:
If you want to add support for data, shouldn't everything in framework_imports that is not headers, modules or framework binary automatically become data?

andre-alves:
This attribute would be really helpful to us. I was about to start this discussion and propose a **[diff](https://github.com/andre-alves/rules_apple/commit/ac34c6a8ea641a9142cdf73f317863ff43740e86)**, but it's the same as @keith's.

I'm migrating a big repo from Buck to Bazel and have some third-party static xcframeworks with resources. These resources are outside the `.xcframework` directory and not using bundles. Buck supports this use-case by passing the resources in `deps`, but currently in Bazel I need to create intermediary libraries to act as "resource aggregators".

I like the `data` attribute implemented here because it's similar to `data` from `objc_library`/`swift_library` and gives the user control to choose the other apple resource rules as needed. If the framework requires a `.bundle` that is already built, we can use `apple_bundle_import`, otherwise, we can build one with `apple_resource_bundle`... the project I'm migrating it's mostly `apple_resource_group`.

Can we merge this feature? The only thing missing for me is that we also need the `data` attribute in `apple_static_xcframework_import` too.

Thanks!

keith:
yea i think merging this or yours is totally fine, i guess i just lost track of it myself. do you want to submit yours with some tests?

keith:
https://github.com/bazelbuild/rules_apple/pull/1841

PR #871, opened 2020-08-11
Make sure test setup is only called once

keith:


keith:
something relies on this, nbd

Issue #762, opened 2020-05-03
Support custom executable name


thii:
Currently the executable name of rules that create executables (`*_application`) always has the same value as the value of `bundle_name` attribute. Although it's possible to change the executable name by just changing the bundle name, it would also change the name of the artifacts depending on that (for instance, `.app` and `.app.dSYM`)—this is error-prone if you want to do something with them (for instance, `ipa_post_processor` does not accept any parameter).

I’d like to request to add a new attribute `executable_name` to rules that create executables to allow specify custom executable name in the final bundle. This could be an optional attribute that, if not specified, will get the value of `bundle_name` or `name` in this order. 

The motivations are:
* With Xcode, we have different executable names for each build configuration—that makes it possible to filter logs using the macOS' Console app using the process name, when you have multiple configurations of the same app installed on the same device.
* Being able to keep the conventions closer to what Xcode allows (same bundle name, but different executable name)

I can try to come up with a patch if this is accepted. Thanks!

thii:
@keith Would you accept a patch for this?

segiddins:
I think this makes sense to me, especially since it's easily configured in Xcode today.

PR #876, opened 2020-08-18
Support custom executable name for rules that create executables.

thii:
This adds a new attribute `executable_name` to rules that create
executables that allows specifying custom executable name in the final
bundle. This is an optional attribute that, if not specified, will get the
value of `bundle_name` or `name` in this order.

This new attribute can be useful, for example, when building a target
with multiple configurations, that it allows you to specify different
executable names for different configurations, while still keeping the
archive bundle name and dSYM bundle name unchanged.

Resolves https://github.com/bazelbuild/rules_apple/issues/762.

### DIFF
--- a/test/starlark_tests/targets_under_test/ios/BUILD
+++ b/test/starlark_tests/targets_under_test/ios/BUILD
@@ -117,6 +117,7 @@ ios_application(
 ios_application(
     name = "app_with_custom_executable_name",
     bundle_id = "com.google.example",
+    bundle_name = "custom_bundle_name",
     executable_name = "app.exe",
     families = ["iphone"],
     infoplists = [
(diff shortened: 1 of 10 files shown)


test/starlark_tests/targets_under_test/ios/BUILD
```diff
@@ -114,6 +114,24 @@ ios_application(
...
+    ],
+    minimum_os_version = "8.0",
+    tags = [
+        "manual",
+        "notap",
+    ],
+    deps = [
+        "//test/starlark_tests/resources:objc_main_lib",
+    ],
+)
```
keith, line 133: can you give this one a custom bundle_name too just to verify the order of operations prefers the custom `executable_name`

segiddins:
Thanks @thii !

Issue #852, opened 2020-07-28
[Feature Request] Support Asset Catalog compiler optimization


thii:
Xcode has a `ASSETCATALOG_COMPILER_OPTIMIZATION` build setting that:

> With no value, the compiler uses the default optimization. You can also specify `time` to optimize for speed of access or `space` to optimize for a smaller compiled asset catalogs.

I'd like to request to add an option specify the optimization value of `actool`. [Buck already does it](https://github.com/facebook/buck/pull/1016).

segiddins:
Related: https://github.com/bazelbuild/rules_apple/issues/732

Issue #732, opened 2020-03-16
Allow specifying copts for different tools invoked by xctoolrunner


segiddins:
Xcode allows specifying some options (for e.g. `momc` or `mapc` or `ibtool` or `actool`) via various build settings (e.g. we rely on setting `--no-inverse-relationship-warnings` to keep builds quiet via `MOMC_NO_INVERSE_RELATIONSHIP_WARNINGS` in Xcode). Currently, the resource rules do no have a way to specify copts that should be given to the resource compiling tools, but it would be really nice if there was.

keith:
Another useful case for this for `actool` would be to pass specific options if you want to produce thinned builds like `--filter-for-device-model` and `--filter-for-device-os-version`

segiddins:
cc @thomasvl, in case you have any thought on how best to implement this

PR #878, opened 2020-08-19
Add support for passing extra options to resource tools

thii:
This adds the following define names:

- `apple.actool_opts`
- `apple.ibtool_opts`
- `apple.momc_opts`
- `apple.mapc_opts`
- `apple.coreml_opts`

that will allow users to pass extra options to the relevant tools.

For example, to apply space optimization for asset catalog compilation,
pass the following option to the build command:

```
--define=apple.actool_opts="--optimization space"
```

Resolves https://github.com/bazelbuild/rules_apple/issues/732.
Resolves https://github.com/bazelbuild/rules_apple/issues/852.

### DIFF
--- a/apple/internal/resource_actions/actool.bzl
+++ b/apple/internal/resource_actions/actool.bzl
@@ -19,6 +19,10 @@ load(
     "xcode_support",
 )
 load(
+    "@build_bazel_rules_apple//apple/internal/utils:defines.bzl",
+    "defines",
+)
+load(
     "@build_bazel_rules_apple//apple/internal/utils:legacy_actions.bzl",
     "legacy_actions",
 )
@@ -205,6 +209,13 @@ def compile_asset_catalog(ctx, asset_files, output_dir, output_plist):
             "--output-partial-info-plist",
             xctoolrunner.prefixed_path(output_plist.path),
         ])
+
+    user_defined_args = defines.list_value(
+        ctx,
+        define_name = "apple.actool_opts",
+        default = [],
+    )
+    args.extend(user_defined_args)
 
     xcassets = group_files_by_directory(
         asset_files,
--- a/apple/internal/resource_actions/datamodel.bzl
+++ b/apple/internal/resource_actions/datamodel.bzl
@@ -15,6 +15,10 @@
 """Datamodel related actions."""
 
 load(
+    "@build_bazel_rules_apple//apple/internal/utils:defines.bzl",
+    "defines",
+)
+load(
     "@build_bazel_rules_apple//apple/internal/utils:legacy_actions.bzl",
     "legacy_actions",
 )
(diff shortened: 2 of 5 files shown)


segiddins:
Doing this via defines first feels weird to me -- it feels like these should primarily be rule attrs, just like `copts` is for `swift_library` and `objc_library`

thii:
I've considered that direction, but given that there aren't any rules specific for compiling resources, plus they can be added via many routes (`*_library`'s `data`; `*_application`, `ios_framework`, `*_extension`, `apple_resource_bundle`'s `resources`), doing this via defines seems to be the only feasible option right now. Any suggestion?

segiddins:
One idea that comes to mind (not saying it's a good one) is doing it via a new target that can be added to deps and propogates a new provider for these options. (Again, not advocating for that route).

I'd like for us to be certain that there's no better option before falling back on defines for configuration

keith:
idk what the state of starlark flags is at this point, but maybe there's an option there too

thii:
I forgot there's already the `apple_resource_group` rule. We can add a new attribute to that rule. But it'd require you to group your resource files every time just to add extra copts.

Starlark configuration is an option, too. Though, (in my limited understanding of it) I'm not very sure what it improves over defines in this specific case.

thii:
To summarize the above suggestions and findings, we have the following options:

- Add additional defines (this PR's current implementation).
  - Pros: No rule interface changes are needed.
  - Cons: Has to be set globally, can't change per target.
  - Usage example

        bazel build --define=apple.actool_opts="--optimization space" //your:target

- Add additional Starlark configurations
  - Pros: Can set per target.
  - Cons: Rule interface changes are (probably) needed.
  - Usage example

        bazel build --@build_bazel_rules_apple//apple:actool_opts="--optimization space" //your:target

- Add a new attribute to `apple_resource_group`
  - Pros: Can set per target.
  - Cons: Grouping resources into `apple_resource_group` targets is needed in order to add extra compiler options.
  - Usage example

    ```
    apple_resource_group(
      name = "MyAssets",
      resources = glob([
        "Assets.xcassets/**",
      ]),
      copts = {
        "actool": ["--optimization", "space"],
      },
    )
    ```

- Apply reasonable default. For example: with `actool`, apply `--optimization space` for `opt`, none for `dbg`, and `--optimization time` for `fastbuild`. A similar approach is being used for plist compilation.
  - Pros: No rule interface changes are needed.
  - Cons: Not customizable.

I'd like to know if there is any other approach or workaround that people are using to handle this.

segiddins:
I think saying 'if you want to customize resource compilation, you need to use `apple_resource_group` or `apple_resource_bundle`' makes sense to me ?

Additionally, I wonder if calling it `actool_copts`, `momc_copts`, etc makes more sense than the dictionary -- it'll lead to better docs, and more helpful error messages when trying to specify copts for a tool that doesn't support it

thii:
> I think saying 'if you want to customize resource compilation, you need to use apple_resource_group or apple_resource_bundle' makes sense to me ? 

Might be my personal preference — removed it from cons.

> Additionally, I wonder if calling it actool_copts, momc_copts, etc makes more sense than the dictionary -- it'll lead to better docs, and more helpful error messages when trying to specify copts for a tool that doesn't support it

SGTM (Originally I didn’t feel like to add a bunch of new attributes)

Also added a new option to the above summary: using Bazel features (as in `ctx.features`).

brentleyjones:
@thii What's the status on this? Do you still want to add it?

thii:
I have no plans to continue working on this any time soon.

PR #879, opened 2020-08-19
Allow App Extensions to be bundle loaders

brentleyjones:
This is needed when using SwiftUI Previews with a WidgetKit extension.


segiddins:
Would be nice to add a quick test that makes use of using the extension as a bundle loader

brentleyjones:
I expanded it to include tvOS and macOS app extensions as well, and have tests for the iOS and tvOS versions (there were no tests for macOS extensions, so I didn't add a test case here).

PR #880, opened 2020-08-19
Add bundle_only for dynamic framework imports

keith:


apple/internal/apple_framework_import.bzl
```diff
@@ -249,7 +249,7 @@ def _apple_dynamic_framework_import_impl(ctx):
         "dynamic_framework_file",
         header_imports,
         module_map_imports,
-        _all_framework_binaries(framework_groups),
+        [] if ctx.attr.avoid_link else _all_framework_binaries(framework_groups),
```
jerrymarino, line 252: Does this work with frameworks built from source and passed with `ios_framework` ? I think this would be a welcoming change 

My current ad-hoc implementation of framework integration has a few pieces:
- a "framework middle man" rule
- a "framework bundler" rule + aspect
The framework middle man is a way to attach frameworks to a rule and the framework bundler handles the business logic of how these frameworks are treated by the build: theres a few possibilites: embedding, embedding+linking, embedding+side-loaded-rpath. It'd be cool to upstream all of this and would probably be better if it wasn't add hoc 👍

keith: this is only handling it with apple_dynamic_framework_import

apple/internal/apple_framework_import.bzl
```diff
@@ -353,6 +353,13 @@ target.
                 [apple_common.Objc, AppleFrameworkImportInfo],
             ],
         ),
+        "avoid_link": attr.bool(
```
kastiglione, line 356: Is there precedent to the `avoid_link` name? If not, maybe a name that indicates it's for bundling?

keith: nope I totally made this up, any thoughts on a specific replacement?

kastiglione: at first I thought `bundle_only`, but then that has the problem of not mentioning linking. Then I thought `unlinked_bundle`. It seems not great but maybe ok.

brentleyjones:
We've run into a need for this as well (EarlyGrey in particular), though in our case we need it to work with `ios_framework` as well. Anything I can do to help get this over the hump?

apple/internal/apple_framework_import.bzl
```diff
@@ -353,6 +353,13 @@ target.
...
+            default = False,
+            doc = """
+Avoid linking the dynamic framework, but still include it in the app. This is useful when you want
+to manually dlopen the framework at runtime
```
brentleyjones, line 360:
```suggestion
to manually dlopen the framework at runtime.
```

apple/internal/apple_framework_import.bzl
brentleyjones, line 356: I like @kastiglione's suggestion: `bundle_only`. The doc can make it clear that it doesn't link.

apple/internal/apple_framework_import.bzl
brentleyjones, line 252: What would we have to change in `_ios_framework_impl` to get this to work for `ios_framework` as well?

keith: Looks like we'll have to change this instead https://github.com/bazelbuild/rules_apple/blob/652eb8ea1802c60d7745ec16f0110ca3cf1ccddb/apple/internal/partials/framework_provider.bzl

keith: I pushed something to that, I haven't tested it at all.

apple/internal/partials/framework_provider.bzl
```diff
@@ -31,7 +31,8 @@ def _framework_provider_partial_impl(
         bin_root_path,
         binary_provider,
         bundle_name,
-        rule_label):
+        rule_label,
+        bundle_only):
```
brentleyjones, line 35: The other actions seem to be in alphabetical order, can this one be as well?

brentleyjones:
I have a local integration need for `ios_framework`, so I'll try that early tomorrow.

brentleyjones:
I tested the `ios_framework` portion of the change locally and it worked perfectly :+1:.

keith:
Finally wrote some tests here

brentleyjones (approved):

thii (approved):

Issue #709, opened 2020-02-14
`--apple_generate_dsym` should also generate `.symbols` for upload to iTunes Connect


brentleyjones:
Xcode allows for exporting symbols to allow iTunes Connect to symbolize crash reports, energy usage reports, and the metrics reports. Bazel currently doesn't support generating these symbols. 

`xcrun symbols` should be called on each dSYM generated and packaged in a "Symbols" folder which is a sibling to the "Payload" folder:

```
xcrun symbols -noTextInSOD -noDaemon -arch all -symbolsPackageDir Symbols \
    bazel-bin/App/Sample.app.dSYM/Contents/Resources/DWARF/Sample_arm64 \
    bazel-bin/Extension/Extension.appx.dSYM/Contents/Resources/DWARF/Extension_arm64 \
    third_party/Framework.framework.dSYM/Contents/Resources/DWARF/Framework ...
```

It should also include the dSYMs for all embedded dynamic frameworks. This might require `apple_dynamic_framework_import` to take a new attribute that points to a dSYM.

Issue #545, opened 2019-08-15
No Symbols directory in IPA


FuegoFro:
When Xcode produces an IPA, at the top level of the archive (eg a sibling of the `Payload/` directory) is a `Symbols/` directory. It seems that this directory is what powers App Store Connect's "Includes Symbols" status as well as the "Download dSYM" button.

However, IPAs produced by `ios_application` don't have this directory and therefore App Store Connect doesn't find symbols for uploaded builds, nor can it (re)provide dSYM files. I'm using the `--apple_generate_dsym` flag and have the dSYM files produced by that, but for parity with Xcode it would be nice to be able generate the `Symbols/` directory.

sergiocampama:
We should be able to add a new flag to be used alongside `--apple_generate_dsym` to include the transitive dSYM bundles into the IPA.

liuliu:
Do we know how these `*.symbols` file generated? These seems to be some kind of processed dsym files, but I am not sure how. With that knowledge, we should be able to modify the ipa to include these.

thii:
```
xcrun symbols '-noTextInSOD' '-noDaemon' '-arch' 'all' '-symbolsPackageDir' 'Symbols' /path/to/dsym/binary
```

PR #882, opened 2020-08-24
Generate and include the symbols files in the IPA file

thii:
Xcode's export action for App Store has an option "Upload your app’s
symbols to receive symbolicated reports from Apple" that:

> Allows Apple to symbolicate crash logs and provide other diagnostic
information. The symbol information is extracted from the debugging
symbols (dSYM files) including in the archive, and is limited to
function and methods (including inline functions), names and paths of
source code files (for navigation), and line number information (for
navigation).

(from https://help.apple.com/xcode/mac/current/#/devde46df08a)

This patch adds a new define named `apple.package_symbols` that supports
the same functionality. When you build your app with
`--apple_generate_dsym` and
`--define=apple.package_symbols=(yes|true|1)`, `.symbols` files will be
generated and packaged within the final IPA file.

Note that this does not generate and package symbols from imported
frameworks yet, because there's currently no API that allows importing
dSYMs provided along with the imported frameworks.

RELNOTES: Added support for packaging symbols files in the produced
IPAs. Pass `--apple_generate_dsym` and
`--define=apple.package_symbols=(yes|true|1)` to your build to enable
this feature. Including symbols in your IPA allows Apple to symbolicate
crash logs and provide other diagnostic information in Xcode's
Organizer.

Fixes https://github.com/bazelbuild/rules_apple/issues/545.
Fixes https://github.com/bazelbuild/rules_apple/issues/709.

### DIFF
--- a/apple/internal/partials/debug_symbols.bzl
+++ b/apple/internal/partials/debug_symbols.bzl
@@ -19,8 +19,20 @@ load(
     "bundling_support",
 )
 load(
+    "@build_bazel_rules_apple//apple/internal:intermediates.bzl",
+    "intermediates",
+)
+load(
+    "@build_bazel_rules_apple//apple/internal:processor.bzl",
+    "processor",
+)
+load(
     "@build_bazel_rules_apple//apple/internal/utils:defines.bzl",
     "defines",
+)
+load(
+    "@build_bazel_rules_apple//apple/internal/utils:legacy_actions.bzl",
+    "legacy_actions",
 )
 load(
     "@bazel_skylib//lib:partial.bzl",
@@ -47,6 +59,10 @@ Depset of `File` references to dSYM files if requested in the build with --apple
 """,
         "linkmaps": """
 Depset of `File` references to linkmap files if requested in the build with --objc_generate_linkmap.
+""",
+        "symbols": """
+Depset of `File` references to symbols files if requested in the build with
+--define=apple.package_symbols=(yes|true|1).
 """,
     },
 )
@@ -145,7 +161,11 @@ def _bundle_dsym_files(ctx, debug_provider, bundle_name, bundle_extension = ""):
 
     return outputs
 
-def _debug_symbols_partial_impl(ctx, debug_dependencies = [], debug_outputs_provider = None):
+def _debug_symbols_partial_impl(
+    ctx,
+    debug_dependencies = [],
+    debug_outputs_provider = None,
+    package_symbols = False):
     """Implementation for the debug symbols processing partial."""
     deps_providers = [
         x[_AppleDebugInfo]
@@ -160,6 +180,9 @@ def _debug_symbols_partial_impl(ctx, debug_dependencies = [], debug_outputs_prov
 
     direct_linkmaps = []
     transitive_linkmaps = [x.linkmaps for x in deps_providers]
+
+    direct_symbols = []
+    transitive_symbols = [x.symbols for x in deps_providers]
 
     output_providers = []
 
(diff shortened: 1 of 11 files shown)


thii:
Thanks for the review. What do you think about enabling the new define flag by default? It's the default behavior of Xcode, so I think people migrating from Xcode would expect the same thing. 

Development-wise, it would allow us to clean up the shell tests for this, and just use Starlark tests.

brentleyjones:
I agree on it being enabled by default.

apple/internal/partials/debug_symbols.bzl
```diff
@@ -188,6 +211,20 @@ def _debug_symbols_partial_impl(ctx, debug_dependencies = [], debug_outputs_prov
                 transitive = [dsym_bundles],
             )
 
+            include_symbols = defines.bool_value(
```
segiddins, line 214: should we continue using defines for this purpose? It doesn't feel like they're really the right way to be doing configuration, vs explicit config settings or target attrs

apple/internal/partials/debug_symbols.bzl
```diff
@@ -225,7 +274,44 @@ def _debug_symbols_partial_impl(ctx, debug_dependencies = [], debug_outputs_prov
...
+    outputs = [symbols_dir]
+
+    commands = ["mkdir -p \"${OUTPUT_DIR}\""]
+
+    for (arch, arch_outputs) in debug_provider.outputs_map.items():
+        dsym_binary = arch_outputs["dsym_binary"]
+        dsym_binaries.append(dsym_binary)
+        commands.append(
+            ("/usr/bin/xcrun symbols -noTextInSOD -noDaemon -arch {arch} " +
+             "-symbolsPackageDir \"${{OUTPUT_DIR}}\" {dsym_binary}").format(
```
segiddins, line 294: should `{dsym_binary}` also be quoted?

apple/internal/partials/debug_symbols.bzl
brentleyjones: I agree we should look into what we should be doing going forward, but I wouldn't want to block this PR on something that might require a lot of design and thought, not tied to this change. For example, I wouldn't want to introduce configuration hash based output directories (which could happen depending on how we do this).

thii: It might not be the right way but I think "define" is how rules_apple is handling its "features" at the moment; for example:  use `--define=apple.package_swift_support` to disable packaging `SwiftSupport` in ipas. You can see more at https://github.com/bazelbuild/rules_apple/blob/master/doc/common_info.md.

apple/internal/partials/debug_symbols.bzl
thii: Good catch -- will fix.

test/starlark_tests/ios_application_tests.bzl
```diff
@@ -138,6 +138,20 @@ def ios_application_test_suite(name = "ios_application"):
...
+        name = "{}_package_symbols_test".format(name),
+        build_type = "simulator",
+        env = {
+            "BINARY_PATHS": ["Payload/app.app/app"],
+        },
+        target_under_test = "//test/starlark_tests/targets_under_test/ios:app",
+        verifier_script = "verifier_scripts/symbols_verifier.sh",
+        tags = [
+            name,
+            "manual",  # can't use Starlark transition on --define now
```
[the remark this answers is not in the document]
thii: No, it won't run at all, that's why I still had to add shell tests. If this feature is enabled by default, then we can allow this to run.

thii:
I think I'll keep this disabled by default since this is a breaking change. We can enable it after a few releases.

thii:
@segiddins Could you merge this too? (if there's no additional feedback)

PR #886, opened 2020-08-28
Enable linkmap tests

thii:
Previously, linkmap generation didn't work because it wasn't enabled in Apple CROSSTOOL (there wasn't a feature for it). With the change in https://github.com/bazelbuild/bazel/pull/12016, this now works correctly, so we can enable these tests now.

Consequently, this also uncomments the linkmap generation documentation.

### DIFF
--- a/doc/common_info.md
+++ b/doc/common_info.md
@@ -107,7 +107,6 @@ mutually exclusive, i.e. you can only specify one or the other for a particular
 build.
 
 -->
-<!-- Blocked on b/73547215
 
 ### linkmap Generation {#objc_generate_linkmap}
 
@@ -122,8 +121,6 @@ bazel build --objc_generate_linkmap //your/target
 By default, only the top level linkmap file is built when this flag is
 specified. If you require the linkmap file of the top level target dependencies,
 you'll need to specify the `--output_groups=+linkmaps` flag.
-
--->
 
 ### Debugging Entitlement Support {#apple.add_debugger_entitlement}
 
--- a/test/ios_application_test.sh
+++ b/test/ios_application_test.sh
@@ -214,6 +214,19 @@ EOF
   expect_log 'Target "//app:app" is missing CFBundleShortVersionString.'
 }
 
+# Tests that the linkmap outputs are produced when --objc_generate_linkmap is
+# present.
+function test_linkmaps_generated() {
+  create_common_files
+  create_minimal_ios_application
+  do_build ios --objc_generate_linkmap //app:app || fail "Should build"
+
+  declare -a archs=( $(current_archs ios) )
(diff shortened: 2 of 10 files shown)


keith:
Hrm why are they passing before your change has merged?

thii:
I think the [`analysis_test` rule’s documentation](https://github.com/bazelbuild/bazel-skylib/blob/2ec2e6d715e993d96ad6222770805b5bd25399ae/docs/analysis_test_doc.md#analysis_test) explains it:

> This rule essentially verifies that all targets under `targets` would generate no errors when analyzed with `bazel build [targets] --nobuild`. Action success/failure for the targets and their transitive dependencies
are not verified. An analysis test simply ensures that each target in the transitive dependencies propagate providers appropriately and register actions for their outputs appropriately.

keith:
but shouldn't they fail right now? otherwise what are these testing?

keith:
maybe we need a new assertion about  that file being produce?

thomasvl:
drive by - They were tagged `manual`, so `bazel test //...` wouldn't have included them.  https://docs.bazel.build/versions/master/test-encyclopedia.html#tag-conventions

thii:
@thomasvl I did remove those `manual` tags.

Added the integration tests back anyway.

thomasvl:
I was answering @keith's

> Hrm why are they passing before your change has merged?

The tests shouldn't have been running with the `manual` tag.

keith:
But since that's removed here and they're still passing without a bazel change, I'm still surprised

thomasvl:
Doh - I didn't look at the upstream change, i thought it was something already landed.

thomasvl:
I think the analysis test is just confirming that the rule did the work to get the file generated, but the integration tests is what is ensuring crosstool did it's job.  So it sorta comes down to if you want to just test the rule is honoring the contract or if you also want to always validate that the crosstool is doing the right thing also.  (validating everything in crosstool could be non trivial)

thomasvl:
For a little more detail - moving more to analysis tests was an attempt to get faster tests and less flaky tests.  In most cases, you don't want to ensure the tool finally invoked worked (hopefully it has testing of its own), instead you just want the rule's tests to ensure the rule is creating the right actions, expected outputs, etc.

The integration tests are much slower because they have to run the tools, some of the tools (ibtool, etc.) are known to flake; so they tend to actually slow down development.  Also, since the integration test requires making a whole mock workspace and invoking a sub build within them, they basically don't cache - any bzl change means the created workspace has _changed_ so they have to be re-run.  The analysis tests are more directly cached based on changes.

thii:
Rebased. The integration tests are passing now as the upstream change has been released. PTAL.

PR #888, opened 2020-08-31
Preserve the codesign load commands in the output binary when stripping bitcode from Swift standard libraries.

thii:
This is needed because the Swift standard libaries bundled in the
SwiftSupport directory in the .ipa file aren't resigned by the
developer's certificate --- they're originally signed by Apple and we're
not supposed to change it.

This change adds the `-keep_cs` flag to the `bitcode_strip` wrapper, and
only run it when building with Xcode 11.4 or above. This preserves the
codesign load commands when stripping bitcode from the Swift standard
libraries, thus ensures those `.dylib`s keep the original code signature
from Apple if we don't resign them.

Note that Swift standard libraries bundled under
`Payload/YourApp.app/Frameworks` are (re)signed with the developer's
certificate --- this change does not affect them.

### DIFF
--- a/apple/internal/partials/swift_dylibs.bzl
+++ b/apple/internal/partials/swift_dylibs.bzl
@@ -108,7 +108,12 @@ def _swift_dylib_action(ctx, platform_name, binary_files, output_dir):
             x.path,
         ])
 
-    if bitcode_support.bitcode_mode_string(ctx) == "none":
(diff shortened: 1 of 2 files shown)


thii:
Closing since this is incorrect—the code signature would be no longer valid when the binary is modified.

PR #889, opened 2020-08-31
Do not modify Swift standard libaries bundled in SwiftSupport

thii:
This is needed because the Swift standard libaries bundled in the
SwiftSupport directory in the .ipa file shouldn't be resigned by the
developer's certificate --- they're already signed by Apple and we're
not supposed to change it.

This change registers an additional action for copying Swift standard
libaries for bundling for SwiftSupport. This action does not touch those
dylibs regardless the current bitcode mode.

Note that Swift standard libraries bundled under
`Payload/YourApp.app/Frameworks` are still (re)signed with the
developer's certificate --- this change does not affect them.


apple/internal/partials/swift_dylibs.bzl
```diff
@@ -155,6 +160,11 @@ def _swift_dylibs_partial_impl(
         transitive = transitive_binary_sets,
     )
 
+    if bitcode_support.bitcode_mode_string(ctx) == "none":
+        strip_bitcode = True
+    else:
+        strip_bitcode = False
```
keith, line 166:
```suggestion
    strip_bitcode = bitcode_support.bitcode_mode_string(ctx) == "none"
```

test/starlark_tests/verifier_scripts/archive_contents_test.sh
```diff
@@ -133,6 +137,49 @@ if [[ -n "${BINARY_TEST_FILE-}" ]]; then
...
+            "$(cat "$codesign_output")"
+        fi
+      done
+    fi
+
+    if [[ -n "${CODESIGN_INFO_NOT_CONTAINS-}" ]]; then
+      for test_codesign_info in "${CODESIGN_INFO_NOT_CONTAINS[@]}"
+      do
+        codesign_info_found=false
+        if grep -sq "$test_codesign_info" "$codesign_output"; then
```
keith, line 163: in what case could this grep fail to find the output, but the loop below would find something?

test/starlark_tests/verifier_scripts/archive_contents_test.sh
```diff
@@ -133,6 +137,49 @@ if [[ -n "${BINARY_TEST_FILE-}" ]]; then
...
+          codesign_info_found=true
+          break
+        fi
+        for actual_codesign_info in "${codesign_output[@]}"
+        do
+          if [[ "$actual_codesign_info" == "$test_codesign_info" ]]; then
```
keith, line 169: should this be less strict, and potentially compare these to `*"$test_codesign_info"*`?

apple/internal/partials/swift_dylibs.bzl
```diff
@@ -171,11 +181,25 @@ def _swift_dylibs_partial_impl(
...
+            # We're not allowed to modify stdlibs that are used for Swift
+            # Support --- ignoring the bitcode mode here.
+            _swift_dylib_action(
+                ctx,
+                platform_name,
+                binaries_to_check,
+                swift_support_output_dir,
+            )
+            swift_support_file = (platform_name, swift_support_output_dir)
             transitive_swift_support_files.append(swift_support_file)
```
keith, line 203: can this logic move into the `if` below, since it shouldn't need to be done if we're not packaging swift support?

apple/internal/partials/swift_dylibs.bzl
```diff
@@ -171,11 +181,25 @@ def _swift_dylibs_partial_impl(
...
 
-            swift_support_file = (platform_name, output_dir)
+            swift_support_output_dir = intermediates.directory(
+                ctx.actions,
+                ctx.label.name,
+                "swiftlibs_for_swiftsupport",
+            )
+            # We're not allowed to modify stdlibs that are used for Swift
+            # Support --- ignoring the bitcode mode here.
+            _swift_dylib_action(
```
keith, line 196: what is the perf hit on running this multiple times? probably nbd since this should only be for "release" builds, but I'm curious if it's a problem

test/starlark_tests/verifier_scripts/archive_contents_test.sh
thii: oops, copypasta bug—the loop below isn't needed.

apple/internal/partials/swift_dylibs.bzl
thii: Didn't see any difference with our app. I'd guess each set of stdlibs would be cached twice and that's it.

And you're right — bundling Swift Support is only needed for App Store builds anyway.

thii:
(Fixed a typo (libaries -> libraries) and squashed all commits)

thii:
@keith Can you take a look? This fixes a regression in https://github.com/bazelbuild/rules_apple/pull/881.

thii:
Updated to only splitting the action when building without bitcode.

keith (approved):

PR #891, opened 2020-08-31
Support substitution of PRODUCT_BUNDLE_PACKAGE_TYPE in Info.plist

dflems:
This is a more complete version of #700 that adds a new member to `rule_support` for the bundle package type. I went through each bundle that `rules_apple` creates, created a corresponding target in an Xcode project, built it, and got the `CFBundlePackageType` value from its plist after building. They tend to be pretty consistent:

* `APPL` = applications for all platforms
* `FMWK` = frameworks for all platforms
* `XPC!` = extensions for all platforms
* `KEXT` = kernel extension

~I had to fix a bunch of existing tests that had the values wrong in either their plists or in the test assertions. A lot of these effectively already serve as a test for this new logic (as `plisttool` will and did fail when there's a conflict). I can add some more if anyone thinks it's necessary.~

Where possible, I updated the Info.plist files for integration tests to set `$(PRODUCT_BUNDLE_PACKAGE_TYPE)` for `CFBundlePackageType` so we can verify that the settings in `rule_support` were propagated down to the packaged bundles.

@allevato @keith @BalestraPatrick


examples/ios/HelloWorldSwift/Info.plist
```diff
@@ -13,7 +13,7 @@
     <key>CFBundleName</key>
     <string>$(PRODUCT_NAME)</string>
     <key>CFBundlePackageType</key>
-    <string>APPL</string>
+    <string>$(PRODUCT_BUNDLE_PACKAGE_TYPE)</string>
```
dflems, line 16: Xcode projects have generated plists with this default value for a while now

apple/internal/resource_actions/plist.bzl
```diff
@@ -240,6 +243,10 @@ def merge_root_infoplists(
         # been valid bugs, so this will still catch that.
         plists.append(struct(CFBundleIdentifier = bundle_id))
 
+    if bundle_package_type:
+        substitutions["PRODUCT_BUNDLE_PACKAGE_TYPE"] = bundle_package_type
+        plists.append(struct(CFBundlePackageType = bundle_package_type))
```
dflems, line 248: One thing that's notable here that wasn't happening before is that this line will validate that `CFBundlePackageType` is set correctly for the current bundle. Not sure if there are any recommendations on whether this should be rolled out slowly with a flag or merged in its current state (where it may fail builds that have the wrong type set in their plists)

keith: yea I thought about this too, one interesting thing is that with resource bundles (which I guess should be `BNDL`) we currently don't set `CFBundlePackageType` at all, and everything works in prod, so I guess it's not always validated. So for us we will start setting this based on this

dflems: As-is, if you don't have it set _at all_, this will set it for you. If you have it set incorrectly (or not specifically enough), the build will fail.

thii (approved):

thii:
There's also a few `infoplist_contents_test`s under `test/starlark_tests`. How about adding the new key to them to verify the produced Info.plist for each bundle type?

test/smart_resource_deduplication_test.sh
```diff
@@ -114,7 +124,7 @@ ios_framework(
     name = "framework",
     bundle_id = "com.framework",
     families = ["iphone"],
-    infoplists = ["Info.plist"],
+    infoplists = ["FrameworkInfo.plist"],
```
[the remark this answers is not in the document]
dflems: It was sharing `Info.plist` with the application (`CFBundlePackageType` = `APPL`)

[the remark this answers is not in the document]
thii: Oh, I didn't know you can do that.

apple/internal/partials/resources.bzl
```diff
@@ -451,6 +454,7 @@ def resources_partial(
         bundle_id: Optional bundle ID to use when processing resources. If no bundle ID is given,
             the bundle will not contain a root Info.plist and no embedded bundle verification will
             occur.
+        bundle_package_type: An optional four-character code representing the bundle type.
```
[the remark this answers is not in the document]
dflems: fixed

dflems:
@thii Updated the `infoplist_contents_test`s

I'm still slightly uneasy about this change as it may break people's builds if they have the incorrect bundle type in their plists. It's an easy fix (update plist values to the correct value or `$(PRODUCT_BUNDLE_PACKAGE_TYPE)`). How do we generally negotiate these types of changes? (cc @keith)

thomasvl:
fyi `forced_plists` likely would avoid the failure, but it means a bad value in the input just gets replaced without a developer knowing.

dflems:
@thomasvl Maybe we can add something like `base_plists` which are applied first and are allowed to be overridden by other plists? That would solve it too. That said, I think this is a good thing to validate - I just worry a bit about breaking people's builds. Is there a pattern that this repo uses for introducing potentially-breaking changes? Feature flags or something?

thomasvl:
There isn't a concept like that at the moment, just the `forced_plists` that can silently override things.  Entitlements validation has plumbing to control the behavior (skip, warn, error), so something like that could be done.  This could also be mapped to the version number in a release, i.e. - bump the major or minor to call out the change in behavior.

In this case, one could argue where things will error, there's a good chance the build was making something sorta invalid since the bundle type was wrong.  The things you "fixed" are tests that never fully tried to use the bundle so they are likely oversights more than anything else.  So, there might not be much chance for breaks in real targets.

Anyway, up to @keith et al for what to do.  Just wanted to point out `forced_plist` is an option; you can see how historically it is used to ensure some things are correct no matter what was in the input plists.

apple/internal/ios_rules.bzl
```diff
@@ -131,6 +131,7 @@ def _ios_application_impl(ctx):
         ),
         partials.resources_partial(
             bundle_id = bundle_id,
+            bundle_package_type = "APPL",
```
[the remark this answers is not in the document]
allevato: There's also a simplification opportunity here. The bundle package type should be associated with the _product type_ rather than the _rule type_. It's a subtle distinction, but since that's the way Xcode models it, it's best to retain that conceptual mapping.

To do that, you can add `bundle_package_type` as a new argument/field to the rule descriptors in rule_support.bzl and associated the appropriate 4-letter code with each of the product types listed there.

Then, these rule functions don't have to change at all. Instead of passing the bundle package type into the partials, have them call `rule_support.rule_descriptor(ctx)` (if they don't already) to retrieve the rule descriptor and pull the package type out of it that way.

apple/internal/resource_actions/plist.bzl
```diff
@@ -280,6 +280,10 @@ def merge_root_infoplists(
             struct(**rule_descriptor.additional_infoplist_values),
         )
 
+    # Replace PRODUCT_BUNDLE_PACKAGE_TYPE based on info in rule descriptor
+    if rule_descriptor.bundle_package_type:
+        substitutions["PRODUCT_BUNDLE_PACKAGE_TYPE"] = rule_descriptor.bundle_package_type
```
dflems, line 285: now it's just setting the `PRODUCT_BUNDLE_PACKAGE_TYPE` override

test/starlark_tests/resources/Info-nopackage.plist
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
dflems, line 1: `Info-nopackage.plist` was added for `macos_dylib` and `macos_command_line_application`. When using Info.plist embedding in binaries, Xcode will completely omit the `CFBundlePackageType` key, even if you include it in your plist (presumably because neither are bundles)

Some logic should be added to the plist tooling to support omitting keys for Info.plist in the embedded use case (this was a bit out-of-scope for this PR specifically)

dflems:
@allevato Updated with your recommendations. I added a new key to the rule descriptor in `rules_support`. `plist.bzl` picks this up automatically and only adds the substitution for `PRODUCT_BUNDLE_PACKAGE_TYPE` (with no validation)

apple/internal/rule_support.bzl
```diff
@@ -110,6 +115,7 @@ def _describe_rule_type(
             that should hold the icon sets (e.g. .xcassets or .xcstickers).
         archive_extension: Extension for the archive output of the rule.
         binary_infoplist: Whether the Info.plist output should be in binary form.
+        bundle_package_type: Four-character code representing the bundle type.
```
dflems, line 118: This is set for everything but `macos_command_line_application` and `macos_dylib` since they are not bundles

apple/internal/bundle_package_type.bzl
```diff
@@ -0,0 +1,27 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Support for bundle package type used by Apple bundling rules."""
+
+# Bundle package type identifiers
+# These correspond to the CFBundlePackageType key in Info.plist files.
+bundle_package_type = struct(
```
segiddins, line 19: for completeness, could add `DEXT` for `com.apple.package-type.wrapper.driver-extension` and `SYSX` for `com.apple.package-type.wrapper.system-extension` and `""` for `com.apple.package-type.in-app-purchase-content`, but not necessary

dflems: `rules_apple` doesn't build these yet. Figured I'd constrain it to just what it's aware of for now.

segiddins: yup totally makes sense, I just grepped through the xcspec and thought I'd share what else came up

dflems:
@allevato Would like if you gave this one last look if you can.

Please merge if it looks good (I don't have merge perms)

apple/internal/bundle_package_type.bzl
```diff
@@ -0,0 +1,27 @@
+# Copyright 2017 The Bazel Authors. All rights reserved.
```
allevato, line 1: Super-nit: 2020

apple/internal/bundle_package_type.bzl
```diff
@@ -0,0 +1,27 @@
...
+
+# Bundle package type identifiers
+# These correspond to the CFBundlePackageType key in Info.plist files.
+bundle_package_type = struct(
+    application = "APPL",
+    bundle = "BNDL",
+    framework = "FMWK",
+    extension = "XPC!",
+    kernel_extension = "KEXT",
+    test_bundle = "BNDL",
```
allevato, line 25: I wouldn't duplicate values here; a "test bundle" is a "(loadable) bundle" so listing it twice feels a bit redundant. Admittedly, "extension" vs. "xpc service" feel different enough that it's weird to pick just one and use it in both places... maybe "extension_or_xpc"?

I don't feel **super** strongly about this though.

apple/internal/bundle_package_type.bzl
dflems: Haha, blind copypasta

apple/internal/bundle_package_type.bzl
dflems: Fixed

PR #893, opened 2020-09-01
unittest.bash: support python 3

dflems:
If `python` on the user's `PATH` is python 3, running the tests fails locally. In python 3, `print()` must be used like a function, not a statement.

> edit: updated to just treat `print` like a function regardless. only very-old versions of python 2 don't support it


segiddins:
Iirc print is also a function in python 2 so I don't think the conditional is needed?

Issue #861, opened 2020-08-10
Actions that declare directories but don't create those directories fail to build with remote strategy


thii:
Currently, there are some resource actions that declare output directories but don't create those directories. Instead, they rely on Bazel's [unexpected behavior](https://github.com/bazelbuild/bazel/issues/6260#issuecomment-444549249) that creates those directories before executing an action. This makes the actions fail to execute with remote execution if the actions require the directories to exist before the execution.

The [documentation for `declare_directory`](https://docs.bazel.build/versions/master/skylark/lib/actions.html#declare_directory) says that
> You must create an action that generates the directory.

The [RBE API documentation](https://github.com/bazelbuild/remote-apis/blob/006a399149dd2dad4eedff247578e94bd957da9c/build/bazel/remote/execution/v2/remote_execution.proto#L512-L514) says that 
> Directories leading up to the output directories (_but not the output directories themselves_) are created by the worker prior to execution

Related:
- https://github.com/bazelbuild/bazel/issues/6260
- https://github.com/bazelbuild/bazel/issues/6393
- https://github.com/bazelbuild/remote-apis/issues/40

sergiocampama:
how does that work? how are rules expected to create an action to create the directory if the directory is an output of the action? this seems to result in either having to wrap the tools in a bash script in order to create the output before invoking the tool or having to modify the tool to do the expected thing, which in many cases is not possible... there is no reason why rbe should require this from the action and not have the executor create the output directory for you, since the worst case is that you have an empty directory which gets deduplicated in the CAS.

thii:
You've said the solutions :P

I imagine you'd have to change the output from a directory to a file, if that's possible. If the directory has to be an output of the action, it seems odd to me that the tool does not create the directory for you. I guess I'll file a radar :).

Having the executor create the output directory for you seems to be the best option, but it's against the REv2 specification (even though some of the RE implementations do that).

thii:
Reported to BuildGrid: https://gitlab.com/BuildGrid/buildgrid/-/issues/330.

edbaunton:
It's worth noting that in v2.1 of RE the distinction between output files and output directories is (dropped)[https://github.com/bazelbuild/remote-apis/blob/master/build/bazel/remote/execution/v2/remote_execution.proto#L493] and it simply becomes an `output_path`. How do you expect this to behave in that situation?

thii:
[...]

PR #896, opened 2020-09-02
Make sure output directory for actool exists before invoking actool

thii:
Previously, each "actool" compile action declared an output directory
but didn't create it. Instead, it relied on Bazel to create the
directory before invoking the tool. With a non-remote spawn strategy,
Bazel would implicitly create it for you; but with the remote spawn
strategy, it is not created by the remote worker.

Additionally, this was also against what was documented for
`declare_directory`, which says that "You must create an action that
generates the directory".

This patches the "actool" wrapper to create an empty directory if one
doesn't exist at the specified output path, to make it works with remote
execution.

Fixes https://github.com/bazelbuild/rules_apple/issues/861.


tools/xctoolrunner/xctoolrunner.py
```diff
@@ -213,6 +213,16 @@ def actool(_, toolargs):
 
   xcrunargs += toolargs
 
+  # The argument coming after "--compile" is the output directory. "actool"
+  # expects an directory to exist at that path. Create an empty directory there
+  # if one doesn't exist yet.
+  for idx, arg in enumerate(toolargs):
+    if arg == "--compile":
+      output_dir = toolargs[idx + 1]
+      if not os.path.exists(output_dir):
```
segiddins, line 222: could use `os.makedirs(output_dir, exist_ok=True)` instead of checking for existence

thii: Checked that but it's only available on Python ≥ 3.5.

segiddins: ugh i wish rules_python just bundled real python toolchains...

segiddins (approved):

PR #898, opened 2020-09-02
Add test for Swift based ios_static_framework

keith:
This just verifies we get the expected files in the output zip


segiddins (approved):

keith:
Great point, thanks for the push: https://github.com/bazelbuild/rules_apple/pull/902

PR #902, opened 2020-09-03
Replace sh test with bzl test

keith:


thii (approved):

keith:
I was looking at doing this for https://github.com/bazelbuild/rules_apple/commit/072a726baa9b5ec3ea92b86a3b8ee710ede265bf too, but this one is tougher:

1. it looks like right now there's no way to generate a framework from swift, only objc
2. symbol validation is direct, no globs, but i would need to validate for / against absolute paths which wouldn't work

segiddins (approved):

PR #903, opened 2020-09-03
Make AppleFrameworkImportInfo public

keith:
This provider is used by the framework import rules to propagate the
files they import. This is useful for users to write custom rules that
can be depended on by the framework import rules.

### DIFF
--- a/apple/providers.bzl
+++ b/apple/providers.bzl
@@ -136,6 +136,19 @@ to ensure that they are explicitly produced as outputs of the build.
     },
 )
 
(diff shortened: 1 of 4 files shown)


apple/providers.bzl
```diff
@@ -103,6 +103,19 @@ specific to any particular binary type.
     },
 )
 
+AppleFrameworkImportInfo = provider(
```
thii, line 106: Super nit: The provider list is sorted.

brentleyjones (approved):

PR #905, opened 2020-09-04
Add a feature to include imported frameworks' module maps in Obj-C compilation inputs

thii:
With the flip of `--incompatible_objc_framework_cleanup`, the
`objc_library` implementation in Bazel no longer passes module maps as
inputs to the compile actions, so that `@import` statements for
user-provided framework no longer work in a sandbox. This trap door
allows users to continue using `@import` statements for imported
framework by adding module map to header_imports so that they are
included in Obj-C compilation but they aren't processed in any way.

### DIFF
--- a/apple/internal/apple_framework_import.bzl
+++ b/apple/internal/apple_framework_import.bzl
@@ -35,6 +35,10 @@ load(
     "resources",
 )
 load(
+    "@build_bazel_rules_apple//apple/internal/utils:defines.bzl",
+    "defines",
+)
+load(
     "@build_bazel_rules_apple//apple:utils.bzl",
     "group_files_by_directory",
 )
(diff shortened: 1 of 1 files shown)


segiddins (approved):
I'm in favor of this -- it would save us needing a wrapper rule. Will give others a chance to weigh in before merging, since I know there's some controversy about using modulemaps.

keith:
what's the way out from this? or is this something we'd expect to support forever?

segiddins:
I think if the shape of `objc_library`'s usage of the `modulemap` field doesn't change, we're stuck with this forever. The alternative is to add a feature to `objc_library` to add modulemaps as inputs to compile actions

thii:
Yep. I can try to add this to `objc_library` but AFAIK it's not accepting new features anymore as it will be moved out to rules_cc. 

The flag is off by default, annotated with `incompatible`, so that people know it may break by default. We could make it more explicit by annotating it with something like `unsupported`.

keith:
yea seems fine to me

thii:
It seems no one else is against this change—maybe it's fine to merge?

keith (approved):

PR #908, opened 2020-09-08
Minimize the action and tweak where it runs.

keith:
- If there is nothing to do, just copy the archive over without requiring
  macOS/the Xcode enviornment.
  - There is a potential behavior change here, if the input archive
    was compressed, but the action wasn't going to compress, then the
    output is now compressed.
- If only doing the compression, also, don't require macOS.
- Add arguments to the script to allow log visibility into the work that is
  being done.

RELNOTES: None
PiperOrigin-RevId: 329546005

### DIFF
--- a/apple/internal/codesigning_support.bzl
+++ b/apple/internal/codesigning_support.bzl
@@ -378,13 +378,7 @@ def _post_process_and_sign_archive_action(
       entitlements: Optional file representing the entitlements to sign with.
     """
     input_files = [input_archive]
-
-    if entitlements:
-        input_files.append(entitlements)
-
-    provisioning_profile = getattr(ctx.file, "provisioning_profile", None)
-    if provisioning_profile:
-        input_files.append(provisioning_profile)
+    processing_tools = []
 
     signing_command_lines = _codesigning_command(
         ctx,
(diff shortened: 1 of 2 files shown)


apple/internal/codesigning_support.bzl
```diff
@@ -407,6 +406,26 @@ def _post_process_and_sign_archive_action(
...
+    # If there is no work to be done, skip the processing/signing action, just
+    # copy the file over.
+    has_work = any([signing_command_lines, ipa_post_processor_path, should_compress])
+    if not has_work:
+        ctx.actions.run_shell(
+            inputs = [input_archive],
+            outputs = [output_archive],
+            mnemonic = mnemonic,
+            progress_message = progress_message,
+            command = "cp -p '%s' '%s'" % (input_archive.path, output_archive.path),
```
segiddins, line 425: should this use `arguments = [input_archive, output_archive]` instead of embedding them into the command string?

keith: @thomasvl is there a reason you prefer one or the other?

thomasvl: Not really.  Something short like this is usually a little clearer to me in string form.  I thinks I also tend to use string form when I use `run_shell` just so help make it clear shell rules will apply to the command.  When it's a straight `run`, I think we tend to build up the args more because they are complex and the separate arg makes it easier.

segiddins (approved):

Issue #892, opened 2020-09-01
Allow apple_dynamic_framework_import to import dSYMs


thii:
Following-up https://github.com/bazelbuild/rules_apple/pull/882, there’s current no API that allows to import user-provided dSYMs that come along with prebuilt frameworks. This makes rules_apple unable to generate `.symbols` files from user-provided dSYMs. This suggests adding a new attribute (`dsym_imports` or `data`) to `apple_{dynamic,static}_framework_import` to make that possible.

keith:
Related https://github.com/bazelbuild/rules_apple/pull/869

PR #715, opened 2020-02-20
Fix output paths colliding for different target with same bundle name

AlessandroPatti:
Bundles are created in `bazel-bin/<package>/<bundle_name>.<app/appex>`. Two targets in the same package that use the same bundle name will see their outputs colliding and Bazel failing. A similar collision occurs with dSYM paths.
These changes replace the current paths with:
   - `bazel-bin/<package>/<target name>_archive_root/Payload/<bundle_name>.<app/appex>`
   - `bazel-bin/<package>/<target name>_dsyms/<bundle_name>.<app/appex>.dSYMS`

PR #910, opened 2020-09-09
Add ability to import user-provided dSYM to apple_dynamic_framework_import

thii:
This adds a new `dsym_imports` attribute to
`apple_dynamic_framework_import` rule, that will allow users to import
the dSYM bundle of the imported framework. This is currently only used
for generating and packaging `.symbols` files into the `.ipa` file for
submitting to the App Store.

Closes https://github.com/bazelbuild/rules_apple/issues/892.

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -12,6 +12,16 @@ x_defaults:
     - "test/..."
     - "examples/..."
 
+  # Tests that need extra flags
+  # 1. Test that .symbols files are generated and bundled in .ipa file
+  package_symbols: &package_symbols
+    platform: macos
+    test_flags:
+    - "--apple_generate_dsym"
+    - "--define=apple.package_symbols=true"
+    test_targets:
+    - "test/starlark_tests:ios_application_package_symbols_test"
+
 # NOTE: To avoid listing the same things for build_flags/test_flags for each
 # of these tasks, they are listed in the .bazelrc instead.
 tasks:
@@ -33,6 +43,11 @@ tasks:
     # has landed on them breaking this project.
     - .bazelci/update_workspace_to_deps_heads.sh
     <<: *common
+
+  macos_latest_package_symbols:
+    name: "Symbols Tests with Latest Bazel"
+    bazel: latest
+    <<: *package_symbols
 
 buildifier:
   version: latest
--- a/apple/internal/partials/framework_import.bzl
+++ b/apple/internal/partials/framework_import.bzl
@@ -92,13 +92,12 @@ def _framework_import_partial_impl(
             files_to_bundle = [x for x in files_to_bundle if x not in avoid_files]
 
     # Collect the architectures that we are using for the build.
-    build_archs_found_dict = dict()
-    for x in targets:
-        if not AppleFrameworkImportInfo in x:
-            continue
-        for build_arch in x[AppleFrameworkImportInfo].build_archs.to_list():
-            build_archs_found_dict[build_arch] = None
-    build_archs_found = build_archs_found_dict.keys()
+    build_archs_found = [
+        build_arch
+        for x in targets
+        if AppleFrameworkImportInfo in x
+        for build_arch in x[AppleFrameworkImportInfo].build_archs.to_list()
+    ]
 
     # Start assembling our partial's outputs.
     bundle_zips = []
--- a/apple/internal/partials/framework_import.bzl
+++ b/apple/internal/partials/framework_import.bzl
@@ -54,6 +54,10 @@ load(
     "@bazel_skylib//lib:paths.bzl",
     "paths",
 )
+load(
+    "@bazel_skylib//lib:sets.bzl",
+    "sets",
+)
 
 # TODO(b/161370390): Remove ctx from the args when ctx is removed from all partials.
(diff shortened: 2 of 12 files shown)


.bazelci/presubmit.yml
```diff
@@ -12,6 +12,16 @@ x_defaults:
     - "test/..."
     - "examples/..."
 
+  # Tests that need extra flags
+  # 1. Test that .symbols files are generated and bundled in .ipa file
+  package_symbols: &package_symbols
+    platform: macos
+    test_flags:
+    - "--apple_generate_dsym"
+    - "--define=apple.package_symbols=true"
```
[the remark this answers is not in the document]
thii: I hesitated to do that since it would unnecessarily slow down the builds of targets that don't need dSYMs and symbols generated.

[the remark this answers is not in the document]
thii: I can give it a test run.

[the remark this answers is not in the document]
thii: The build fails if we pass `--apple_generate_dsym` to everything, because of output paths collision. (There's a patch that attempts to fix this: https://github.com/bazelbuild/rules_apple/pull/715)

thii:
Updated to also extracting debug symbols from frameworks that don't provide dSYMs. Thanks @brentleyjones for letting me know this.

brentleyjones:
@thii There is a test failiure

thii:
It was because of an incompatible flag being flipped in https://github.com/bazelbuild/bazel/commit/213043a1b9fa5053fae720220f28ebc604283079. Fixed here: https://github.com/bazelbuild/rules_apple/pull/965.

brentleyjones:
Merged the fix, can you rebase? Thanks!

brentleyjones (approved):
This is nice, thanks @thii for including the non-dSYM case.

I can't test `master` yet locally, so I'll wait for someone else to approve as well before we merge it.

brentleyjones (approved):
I hadn't realized that `apple.package_symbols` defaulted to `no` (our version had that flipped).

This generates all of our symbols now (though more than we had before, because we only did `arm64` before, and I see your comment on not knowing the arch yet. Is there a way to narrow it down, similar to show we strip architectures of imported frameworks? And if not, have you verified that submitting more symbols than needed doesn't cause an upload error in Appstore Connect?

apple/internal/partials/framework_import.bzl
```diff
@@ -83,12 +92,13 @@ def _framework_import_partial_impl(
...
-        if AppleFrameworkImportInfo in x
-        for build_arch in x[AppleFrameworkImportInfo].build_archs.to_list()
-    ]
+    build_archs_found_dict = dict()
+    for x in targets:
+        if not AppleFrameworkImportInfo in x:
+            continue
+        for build_arch in x[AppleFrameworkImportInfo].build_archs.to_list():
+            build_archs_found_dict[build_arch] = None
+    build_archs_found = build_archs_found_dict.keys()
```
thii, line 101: When debugging this PR, I've found that this may contain duplicate values, so I changed the implementation of it. Not sure if I should submit it here.

apple/internal/partials/framework_import.bzl
```diff
@@ -193,15 +203,112 @@ def _framework_import_partial_impl(
...
+    )
+    outputs = [symbols_dir]
+
+    commands = ["mkdir -p \"${OUTPUT_DIR}\""]
+
+    for binary in all_binaries:
+        # If dSYMs are bundled with multiple non-fat binaries, the 'symbols'
+        # command may try to extract symbols from a binary that doesn't have a
+        # slice for an architecture, but it's fine since it won't return a
+        # non-zero code in that case.
```
thii, line 286: Not sure if we should optimize for this case since it's very uncommon. If we do, then it might be better to extract these commands into a tool, than adding more ad-hoc bash here.

apple/internal/partials/framework_import.bzl
brentleyjones: > Not sure if I should submit it here.

Is the change you have here what you aren't sure of? Without the change, what impact does the duplicates have?

apple/internal/partials/framework_import.bzl
brentleyjones: I'm fine with not optimizing for now, that is something we can improve later. Thoughts @keith?

apple/internal/partials/framework_import.bzl
thii: e.g. `[x86_64, x86_64, x86_64]`

thii: The change is kind of unrelated to this PR. I can submit it separately if this PR is going to take some more time. 

Without this, some commands may end up being executed more than once.

brentleyjones: I would submit it as a followup PR, since it's unrelated to the core of the change.

thii:
The failure with the last green Bazel seems to be because of https://github.com/bazelbuild/bazel/commit/d3fc253a49a00c34408bbaf5378376cbcea1c5c9.

brentleyjones (approved):

brentleyjones:
Thanks again @thii!

PR #916, opened 2020-09-16
Add support for App Clips to rules.

keith:
This change adds a new ios_app_clip rule, and enables embedding app clips in ios_application rules through a new app_clips property.

Additional changes are made to codesigning to support multiple code signed archives as outputs for any given target, to support an app clip having both a signed .ipa and a signed embeddable archive for use when embedding in an application target.

Unit tests changes include verification of embedding and verification that both embedded and standalone app clips are correctly signed and possess app clip entitlement.

Based on: PR #838 by brentleyjones

RELNOTES: None
PiperOrigin-RevId: 332074165


keith:
looks like some tests fixes are required here

segiddins (approved):

PR #917, opened 2020-09-21
Converted two partials to the new ctx-less format. Addressed run_support to remove ctx in the process.

keith:
PiperOrigin-RevId: 332890021

### DIFF
--- a/apple/internal/BUILD
+++ b/apple/internal/BUILD
@@ -213,10 +213,12 @@ bzl_library(
     ],
     deps = [
         ":apple_product_type",
+        ":bundling_support",
         ":codesigning_support",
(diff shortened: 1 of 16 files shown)


keith:
made some more changes for the custom executable_name, still lots more to do i think, not going to do it now

thii:
I can help on that if you don't mind, since I added that attribute.

keith:
sure, that would be great!

keith:
you can submit a PR into this branch, or take this cherry pick and apply fixes yourself, up to you!

thii:
https://github.com/bazelbuild/rules_apple/pull/918

PR #918, opened 2020-09-22
Converted two partials to the new ctx-less format. Addressed run_support to remove ctx in the process.

thii:
Add executable_name parameters where applicable to fix compatibility
with the open-source branch

Also remove binary_partial's bundle_name parameter, since it was only
used to declare the executable filename, which is now handled by
the executable_name parameter.


keith (commented):
awesome, thanks for owning this!

PR #920, opened 2020-09-24
Enable configuration transition for iOS rules.

keith:
This is an NFC for macOS/tvOS/watchOS since they already had the transition applied.

RELNOTES: None.
PiperOrigin-RevId: 333561201


segiddins (approved):

sergiocampama:
finally! 🎉

PR #921, opened 2020-09-24
Enable apple_rule_transition

keith:
Historically this was disabled because it used to require a flag we
didn't want to force on everyone. Now that is on by default.


segiddins (approved):
I'm a tad concerned about this, since it's a big change (conceptually). We should make sure it gets mentioned in RELNOTES, even if we think it should be NFC

keith:
had to apply with https://github.com/bazelbuild/rules_apple/pull/920

PR #923, opened 2020-09-26
Use swift-stdlib-tool's own --strip-bitcode option instead of invoking bitcode_strip to remove embedded bitcode from Swift libraries

thii:


thii:
The failing app clip test seems unrelated, it has been randomly failing on master too.

sergiocampama:
did you profile this change? last time I tried, IIRC, it was slower to do that for some reason, or something was not great about it, can't remember well, (so long ago, tho)

thii:
Are you asking about stripping bitcode vs not stripping it? (Because this change is about switching to `swift-stdlib-tool`'s `--strip-bitcode` option instead of invoking `bitcode_strip` directly.)

I profiled a test app that embeds 36 dylibs but didn't see any significant difference:

```
(times in seconds)
with bitcode_strip
(56.498+52.931+51.357+53.342+52.617+61.679+56.253+57.147+52.399+54.653)/10 = 54.8876

with swift-stdlib-tool --strip-bitcode
(58.643+53.106+57.880+54.701+55.701+54.965+57.089+57.292+58.474+57.285)/10 = 56.5136

no stripping bitcode
(56.598+62.710+55.816+53.562+56.597+55.798+53.957+54.272+56.888+54.444)/10 = 56.0642
```

thomasvl:
I think what @sergiocampama might be recalling is originally we didn't strip bitcode in here because one can end up stripping multiple times when you have extensions.  I think the flow was, we have to run the tool for each extension that uses Swift (and the parent app), then required support libs are all collected and deduplicated into the final copies.  So by doing a strip at the end, you only do it once per support library, but if you strip while collecting, since you might collect something multiple times, you strip it multiple times.

I know we also wanted to see about doing one collection actions for all extensions and the parent app, at which point then it would be better to strip while collecting.  But the catch was always build vs. packaging, and if someone just builds an extension, do we still bundle the Swift support libs or not?  And with the build graph normally works, there wasn't a good way to know that case.

brentleyjones:
@thii Do you still wish to land this change?

thii:
I'm dropping this.

PR #925, opened 2020-09-28
Migrate the remaining `ios_*` rules (except `ios_static_framework`) to `apple_common.link_multi_arch_binary`.

keith:
After this change, no Apple rules use `apple_binary` directly. Instead, they call its functionality through the Starlark API that Bazel core exposes. The `{ios,tvos}_static_framework` rules still use `apple_static_library` since there is no Starlark-callable API equivalent of that yet.

PiperOrigin-RevId: 334189971

### DIFF
--- a/apple/internal/binary_support.bzl
+++ b/apple/internal/binary_support.bzl
@@ -15,10 +15,6 @@
 """Binary creation support functions."""
 
 load(
-    "@build_bazel_rules_apple//apple/internal:apple_product_type.bzl",
-    "apple_product_type",
-)
-load(
     "@build_bazel_rules_apple//apple/internal:entitlement_rules.bzl",
     "entitlements",
 )
@@ -27,16 +23,8 @@ load(
     "exported_symbols_lists",
 )
 load(
-    "@build_bazel_rules_apple//apple/internal:rule_support.bzl",
-    "rule_support",
-)
-load(
     "@build_bazel_rules_apple//apple/internal:swift_support.bzl",
     "swift_runtime_linkopts",
-)
-load(
-    "@bazel_skylib//lib:collections.bzl",
-    "collections",
 )
 
 def _create_swift_runtime_linkopts_target(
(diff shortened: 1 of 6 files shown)


apple/internal/binary_support.bzl
```diff
@@ -193,141 +181,7 @@ def _add_exported_symbols_lists(name, exported_symbols_lists_value):
...
-    """
-    bundling_args = dict(kwargs)
-
-    rule_descriptor = rule_support.rule_descriptor_no_ctx(platform_type, product_type)
-
-    linkopts = kwargs.pop("linkopts", [])
-    linkopts = linkopts + collections.before_each("-rpath", rule_descriptor.rpaths)
-    linkopts = linkopts + rule_descriptor.extra_linkopts
-    if extension_safe:
-        linkopts = linkopts + ["-application_extension"]
```
keith, line 250: Note this change is wiped out, we probably need to reapply it elsewhere

segiddins: or is it OK because we pass `-fapplication-extension` when needed?

segiddins (approved):

apple/internal/binary_support.bzl
keith: I think for that to matter we also have to have the linkopt

segiddins: https://github.com/bazelbuild/rules_apple/pull/925/files#diff-eebcb98abfda3b6dca6c52be04dd76b3R271 it's there?

keith: that's only for extensions, this also applied to  ios_framework targets, pushed a commit that should restore this

segiddins: nice, though ideally we'd have a test 🙃

keith: it doesn't seem like there's a way to use starlark tests to validate flags atm, is that what you've seen too?

segiddins (approved):

Issue #911, opened 2020-09-11
AppleBinaryInfo providers should be listed in OutputGroupInfo


nikhilm:
I have a use case where I have

```
macos_dylib(
  name = "foo",
  ...
)

pkg_tar(
  ...
  srcs = [":foo"],
)
```
to package a dylib as part of a tar. When dsyms are enabled, this does the wrong thing as the dSYM also ends up in the tar, which i don't want.

I thought of using `filegroup()` with `output_group` to request a specific provider (the `AppleBinaryInfo` returned from macos_dylib), but since the AppleBinaryInfo is not advertised in the returned OutputGroupInfo, it cannot be retrieved in this way.

I'm working around this for now by having a custom rule that accesses the `AppleBinaryInfo` and returns a `DefaultInfo`, but it would be nice to fix this for all users and all of the Apple rules.

What would such a PR look like given the partials used? I'm happy to submit one with some guidance.

keith:
Hrm, I guess it might be arguable whether or not you want the dSYMs in the tar, depending on your use case. Can you exclude those from your build, or with `--output_groups=-dsyms` instead (I haven't verified these would exclude them myself but I would assume so)?

nikhilm:
That is an excellent point relevant to this. I realized after that using my custom rule won't really work.

The larger problem I'm trying to solve is:
- I do want the dsym to be generated, so it is important my overall build graph use the `DefaultInfo` so Bazel will actually generate the dsym.
- I don't want to put the dsym in the tar, since the tar effectively represents a bunch of files that I'm going to package into something to ship to customers.

This means `--output_groups=-dsyms`  isn't what I want either.

There is some potential for using something like https://github.com/bazelbuild/rules_pkg/issues/85, otherwise, for now, I've taken to just removing the file from the tar outside of Bazel, in a post-build step.

I'm not sure if rules_apple is the right place to solve something like this. Feel free to close this.

keith:
Yea that all makes sense. As a short term workaround I would probably write my own pkg_tar rule (or use a genrule) and pick and choose the files you want from the final output (emulating the issue you linked).

I can definitely understand the use case here, but I don't have any good ideas on how this would fit in to rules_apple, it seems like the most likely medium term solution is for that issue to be resolved, but I'll leave this open in case others have ideas!

nikhilm:
[...]

PR #929, opened 2020-09-30
Put macos_dylib into its own output group

nikhilm:
This is what I was thinking for #911.

The default outputs remain the same (dylib + dsyms).
Now there is an additional output group `dylib` that has only the dylib, which can be retrieved by a downstream filegroup, similar to how `dsyms` already exists.
Meanwhile, the AppleBinaryInfo provider is still maintained.

What do you think?

### DIFF
--- a/apple/internal/macos_rules.bzl
+++ b/apple/internal/macos_rules.bzl
@@ -1277,6 +1277,7 @@ def _macos_dylib_impl(ctx):
             **outputs.merge_output_groups(
                 link_result.output_groups,
                 processor_result.output_groups,
+                {"dylib": depset(direct = [output_file])},
             )
         ),
         link_result.binary_provider,
--- a/test/starlark_tests/macos_dylib_tests.bzl
+++ b/test/starlark_tests/macos_dylib_tests.bzl
@@ -26,6 +26,10 @@ load(
     ":rules/dsyms_test.bzl",
     "dsyms_test",
 )
+load(
+    ":rules/output_group_test.bzl",
+    "output_group_test",
+)
 
 # buildifier: disable=unnamed-macro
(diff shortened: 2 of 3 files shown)


nikhilm:
bump @keith

nikhilm:
This is ready to go. Rebased on latest master.

apple/internal/macos_rules.bzl
```diff
@@ -1253,24 +1274,17 @@ def _macos_dylib_impl(ctx):
...
     return [
-        AppleBinaryInfo(
-            binary = output_file,
-            product_type = rule_descriptor.product_type,
-        ),
         DefaultInfo(files = depset(transitive = [
-            depset([output_file]),
             processor_result.output_files,
         ])),
         OutputGroupInfo(
```
keith, line 1290: I think you could inline the dylib into this, if you had to you could do the merging with a dictionary of `{"dylib": output_file}` or something here? If this worked this would be a less invasive change

nikhilm:
That simplifies it. How would I write tests for htis?

nikhilm:
bump

brentleyjones (approved):

brentleyjones:
You should be able to write star lark tests for this. See `dsyms_test` for an example of testing for the existence of the dsyms in the `OutputGroupInfo`, you could do the same thing for testing the dylib existence.

nikhilm:
Sorry for the slow update. There is a test now. I think this can be merged.

brentleyjones:
There is an unused `load` @ test/starlark_tests/rules/output_group_test.bzl:19, after that is fixed we can merge it.

keith (approved):

PR #932, opened 2020-10-05
Pass the correct archive root path when codesigning for the embedding archive

segiddins:
This happened to work due to lack of sandboxing, and racing with the non-embedded case

cc @klmcarthur who landed this code upstream


keith (approved):

brentleyjones (approved):

klmcarthur:
LGTM - Thanks for fix

PR #934, opened 2020-10-07
dSYM binary should be named the same with app executable

thii:
This mirrors Xcode's behavior when specifying custom executable name for
the app bundle. Without this, certain crash analytics service wouldn't
be able to correctly symbolicate crash logs.


segiddins (approved):

thii:
Rebased and fixed conflicts. Would appreciate a merge.

PR #941, opened 2020-10-12
Use -Wl so linkopts only use a single argument

segiddins:
This ensures they wont be incorrectly de-duplicated if put into a NestedSet

### DIFF
--- a/apple/internal/linking_support.bzl
+++ b/apple/internal/linking_support.bzl
@@ -18,10 +18,6 @@ load(
     "@build_bazel_rules_apple//apple/internal:rule_support.bzl",
     "rule_support",
 )
-load(
-    "@bazel_skylib//lib:collections.bzl",
-    "collections",
-)
 
 def _sectcreate_objc_provider(segname, sectname, file):
     """Returns an objc provider that propagates a section in a linked binary.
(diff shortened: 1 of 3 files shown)


apple/internal/linking_support.bzl
```diff
@@ -89,7 +85,7 @@ def _register_linking_action(ctx, extra_linkopts = []):
     rpaths = rule_descriptor.rpaths
     linkopts = []
     if rpaths:
-        linkopts.extend(collections.before_each("-rpath", rpaths))
+        linkopts.extend(["-Wl,-rpath,{}".format(rpath) for rpath in rpaths])
```
keith, line 88:
```suggestion
        linkopts.append("-Wl,-rpath,{}".format(rpath) for rpath in rpaths)
```

keith (commented):
this seems legit but I wonder if there's ever a difference by doing this vs letting clang forward them

apple/internal/linking_support.bzl
segiddins: how would this work? don't we need extend since we're adding multiple elements?

keith: oh sorry you're right, you still need extend, but at least in py you wouldn't need to form the list normally

keith: turns out starlark doesn't support this mode

keith, line 88: you can remove this `if` now

keith (approved):

keith:
https://github.com/bazelbuild/rules_apple/pull/973

PR #942, opened 2020-10-12
Add codesignopts argument for adding custom codesign flags

michaeleisel:
This would be the equivalent of `linkopts` but for codesign. I know of two cases where additional flags that aren't currently supported by rules_apple would be helpful, each of which speeds up signing: `--digest-algorithm=sha1` and `--resource-rules=<path to resource rules>`. I don't know if we can support the latter with just this change, since it wants a path to a resource rules file (perhaps it warrants its own `--define=...`?). However, for the digest algorithm case, and potentially other flags, this would be nice to have. Finding useful code signing flags is basically an open problem, and these two cases I mention are just what I've personally discovered, not to mention other documented flags. As `man codesign` says, `codesign has several operations and options that are purposely left undocumented in this manual page because they are either experimental (and subject to change at any time), or unadvised to the unwary`. It could be nice to have a flexible way to add flags.


apple/internal/rule_factory.bzl
```diff
@@ -267,6 +267,11 @@ Do not change its value.
             doc = """
 This attribute is public as an implementation detail while we migrate the architecture of the rules.
 Do not change its value.
+    """,
+        ),
+        "signopts": attr.string_list(
```
keith, line 272: I wonder if this should be:

```suggestion
        "codesignopts": attr.string_list(
```

?

tools/codesigningtool/codesigningtool.py
```diff
@@ -244,10 +245,20 @@ def generate_arg_parser():
       "--disable_timestamp", action="store_true", help="disables the use of "
       "timestamp services"
   )
+  parser.add_argument("extra", nargs = argparse.REMAINDER, help="additional "
+      "arguments that go directly to codesign, starting with a '--' argument")
```
keith, line 249: What's the need for the extra `--`? just a sanity check?

michaeleisel: according to argparse docs, "If you have positional arguments that must begin with - and don’t look like negative numbers, you can insert the pseudo-argument '--' which tells parse_args() that everything after that is a positional argument:"

apple/internal/rule_factory.bzl
michaeleisel: done

michaeleisel:
so is there support of this and i should polish it with tests?

tools/codesigningtool/codesigningtool.py
michaeleisel: i also like that it mirrors `bazel run` and its use of `--` in a way

keith:
tests would be great, I'm supportive

test/starlark_tests/targets_under_test/ios/BUILD
```diff
@@ -1023,6 +1023,20 @@ ios_extension(
 
 # ---------------------------------------------------------------------------------------
 
+ios_application(
```
michaeleisel, line 1026: i see some other rules have tags "notap" and "manual", not sure if they should be here as well

michaeleisel:
@keith ptal

michaeleisel:
as for documentation, is that auto-generated or does it have to be updated manually?

script.sh
```diff
@@ -0,0 +1,2 @@
+#!/bin/bash
```
segiddins: accidentally committed?

test/starlark_tests/verifier_scripts/codesignopts_verifier.sh
```diff
@@ -0,0 +1,22 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -eu
+
+# Assert that the bundle itself is signed.
+assert_is_codesigned "$BUNDLE_ROOT"
```
segiddins: this file appears to be unused?

michaeleisel: your pr comments are on outdated commits

test/starlark_tests/targets_under_test/ios/BUILD
keith: Everything should have `manual`, `notap` is a google thing so probably doesn't matter anymore, but also wouldn't hurt to include.

keith (approved):

test/starlark_tests/targets_under_test/ios/BUILD
michaeleisel: done

keith:
guess we should wait to merge this until we figure out if this supports the resource rules case as discussed in slack

michaeleisel:
@thii the toolchain stuff you discussed for zld, would it work for here as well to have a file available? if so, then this would support `--resource-rules=$(RESOURCE_RULES)` without more attrs needed

thii:
Generally, it only works with the local spawn strategy. You still need to add the files to the action's inputs.

For this `ProcessAndSign` action though, since it already has an execution requirement of `no-sandbox`, I think it will work.

michaeleisel:
@thi when i tried the toolchain stuff with `linkopts`, it works, but with `codesignopts`, it gets replaced with an empty string rather than the path. i've tried `bazel build --spawn_strategy=local` to make sure it wasn't using the sandbox. any ideas what's wrong?

apple/internal/codesigning_support.bzl
```diff
@@ -149,6 +149,9 @@ def _codesign_args_for_path(
                 maybe_double_quote(signed_framework),
             ])
 
+    extra_opts = getattr(ctx.attr, "codesignopts", [])
```
thii, line 152: It seems this is missing the expanding of Make variables.
```starlark
    extra_opts = [ctx.expand_make_variables("codesignopts", f, {}) for f in extra_opts]
```

michaeleisel:
nice, it works now. note that i used `opt.format(**ctx.var)` instead of `ctx.expand_make_variables("codesignopts", opt, {})` since it appears the latter is deprecated: https://github.com/bazelbuild/bazel/issues/5859

michaeleisel:
this also means one needs to use `{RULES}` and not `$(RULES)`

thii:
Can we make it works with `$(RULES)`? Since that's how Bazel's "Make" variables are supposed to work. It seems odd if `codesignopts` alone use a different format than other attributes.

michaeleisel:
done

thii (approved):

michaeleisel:
@keith thoughts on merging, now that we have it working for both the digest-sha and resource-rules use cases?

test/starlark_tests/targets_under_test/ios/BUILD
```diff
@@ -1061,6 +1061,24 @@ ios_extension(
...
+    name = "app_with_codesignopts",
+    bundle_id = "com.google.example",
+    codesignopts = ["--digest-algorithm=sha1", "--digest-algorithm=sha384"],
```
keith, line 1067: should we add the resource rules case as a test here to validate the expansion behavior works?

keith (approved):
looks good to me! one small question about testing

keith:
looks like the buildifier issue here is related to this change, i'll have to try and fixup the other issue

michaeleisel:
some other failure now?

brentleyjones:
It's a linter warning:

```
test/testdata/rules/substitution.bzl:1: module-docstring: The file has no module docstring.
A module docstring is a string literal
```

PR #945, opened 2020-10-14
Allow overriding the linker executable

michaeleisel:
since many large orgs use a custom linker via `-fuse-ld=`, and it sounds like the alternative to this is either to break hermeticity or change the toolchain, it seems like a nice option to have

### DIFF
--- a/apple/internal/linking_support.bzl
+++ b/apple/internal/linking_support.bzl
@@ -92,6 +92,8 @@ def _register_linking_action(ctx, extra_linkopts = []):
         linkopts.extend(collections.before_each("-rpath", rpaths))
 
     linkopts.extend(rule_descriptor.extra_linkopts + extra_linkopts)
+    if hasattr(ctx.executable, "linker"):
+        linkopts.append("-fuse-ld=" + ctx.executable.linker.path)
 
     binary_provider_struct = apple_common.link_multi_arch_binary(
         ctx = ctx,
--- a/apple/internal/rule_factory.bzl
+++ b/apple/internal/rule_factory.bzl
@@ -269,6 +269,14 @@ This attribute is public as an implementation detail while we migrate the archit
 Do not change its value.
     """,
         ),
+        "linker": attr.label(
+            executable = True,
+            cfg = "host",
+            allow_single_file = True,
+            doc = """
+An executable to use for linking instead of the default linker
+    """,
+        ),
         "linkopts": attr.string_list(
             doc = """
 A list of strings representing extra flags that should be passed to the linker.


michaeleisel:
@keith thoughts on this? also, would you happen to know the preferred way of specifying it as an input? currently, this rule doesn't work because it can't find the linker executable at link-time

segiddins:
This feels like something that's better done in the cc crosstool?

michaeleisel:
i've always found it confusing what the differences are between crosstools, toolchains, and platforms, and i was also under the impression that i had to override a whole bunch of stuff just to override one thing, meaning mimicking a lot of stuff in the default setup. but maybe i'm wrong. is there a simple and concise way to do this without changing the rules?

thii:
Do you plan to finish this? This would be nice to have at least in the short term. The workaround of placing zld at a global path works, but it would require you to provision your remote execution worker the same way, which is what we want to avoid.

michaeleisel:
so, after learning about that trick of using .bazelrc to get linkopts not to add `-Wl,`, i'd be inclined to be to close this and just add a rule like swift_library's `data` attr, then use linkopts

thii:
How about using `toolchains`? Something like this:

```starlark
# defs.bzl
def _tool_impl(ctx):
    return [
        platform_common.TemplateVariableInfo({
            ctx.attr.var_name: ctx.file.zld.path,
        }),
    ]

tool = rule(
    implementation = _tool_impl,
    attrs = {
        "var_name": attr.string(),
        "zld": attr.label(
            allow_single_file = True,
        ),
    },
)
```

```starlark
# BUILD
load(":defs.bzl", "tool")

tool(
    name = "zld_bin",
    var_name = "ZLD",
    zld = "zld",
)

ios_application(
    name = "HelloWorld",
    app_icons = ["//examples/resources:PhoneAppIcon.xcassets"],
    bundle_id = "com.example.hello-world",
    families = [
        "iphone",
        "ipad",
    ],
    infoplists = [":Info.plist"],
    launch_storyboard = "//examples/resources:Launch.storyboard",
    linkopts = ["-fuse-ld=$(ZLD)"],
    minimum_os_version = "8.0",
    toolchains = [":zld_bin"],
    version = ":HelloWorldVersion",
    deps = [":Sources"],
)
```

We'd still need to fix rules_apple to tell it not to add an extra `-Wl,` here, though.

michaeleisel:
interesting, where is this toolchains attr documented

thii:
https://docs.bazel.build/versions/master/be/common-definitions.html#common.toolchains

michaeleisel:
@keith is this different from the crosstool stuff you were talking about? this looks relatively simple

keith:
🤔 yes this is a new use case to me. if that works that seems awesome!

thii:
The absolute path issue can be fixed by https://github.com/bazelbuild/bazel/pull/12337.

apple/internal/linking_support.bzl
```diff
@@ -92,6 +92,8 @@ def _register_linking_action(ctx, extra_linkopts = []):
         linkopts.extend(collections.before_each("-rpath", rpaths))
 
     linkopts.extend(rule_descriptor.extra_linkopts + extra_linkopts)
+    if hasattr(ctx.executable, "linker"):
+        linkopts.append("-fuse-ld=" + ctx.executable.linker.path)
 
     binary_provider_struct = apple_common.link_multi_arch_binary(
```
thii, line 98: `link_multi_arch_binary` has an `extra_link_inputs` parameter. Let's use that instead.

PR #946, opened 2020-10-15
Workaround actool error: "failed to open liblaunch_sim.dylib."

jerrymarino:
`actool` is flaky on Xcode 11 betas, all Xcode 12 builds and it hasn't been fixed as of 12.1.  In testing a large build with multiple failures, this has fixed the issue. The error happens on ~1% of builds and it seems better to retry _inside_ of Bazel than to retry the entire build. Putting this up as RFC to get everyones appetite on fixing this in Bazel.

Note: this issue has been persisting in all of Xcode 12 builds including 12.1.

### DIFF
--- a/tools/xctoolrunner/xctoolrunner.py
+++ b/tools/xctoolrunner/xctoolrunner.py
@@ -235,10 +235,10 @@ def actool(_, toolargs):
   # based on the same codebase.
   #
   # Note: `actool` is problematic on all Xcode 12 builds including to 12.1. 25%
-  # of the time, it fails with the error: "failed to open # liblaunch_sim.dylib`.
+  # of the time, it fails with the error:
+  # "failed to open # liblaunch_sim.dylib"
   #
-  # This workaround adds a retry, and this works due to logic in `actool`.
-  #
+  # This workaround adds a retry it works due to logic in `actool`:
   # The first time `actool` runs, it spawns a dependent service as the current
   # user. After a failure, `actool` spawns it in a way that subsequent
   # invocations will not have the error. It only needs 1 retry.
@@ -254,7 +254,7 @@ def actool(_, toolargs):
       sys.stdout.write("%s" % stdout)
     if stderr:
       sys.stderr.write("%s" % stderr)
-      return return_code
+    return return_code
 
   return_code, _, _ = execute.execute_and_filter_output(
       xcrunargs,
--- a/tools/xctoolrunner/xctoolrunner.py
+++ b/tools/xctoolrunner/xctoolrunner.py
@@ -238,6 +238,16 @@ def actool(_, toolargs):
       trim_paths=True,
       filtering=actool_filtering,
       print_output=True)
+
+  if return_code == 0:
+      return return_code
+
+  # `actool` is problematic on Xcode 12 so retry
+  return_code, _, _ = execute.execute_and_filter_output(
+      xcrunargs,
+      trim_paths=True,
+      filtering=actool_filtering,
+      print_output=True)
   return return_code
 
 
--- a/tools/xctoolrunner/xctoolrunner.py
+++ b/tools/xctoolrunner/xctoolrunner.py
@@ -233,16 +233,29 @@ def actool(_, toolargs):
   # helps.
   # Yes, IBTOOL appears to be correct here due to "actool" and "ibtool" being
   # based on the same codebase.
-  return_code, _, _ = execute.execute_and_filter_output(
+  #
+  # Note: `actool` is problematic on all Xcode 12 builds including to 12.1. 25%
+  # of the time, it fails with the error: "failed to open # liblaunch_sim.dylib`.
+  #
+  # This workaround adds a retry, and this works due to logic in `actool`.
+  #
+  # The first time `actool` runs, it spawns a dependent service as the current
+  # user. After a failure, `actool` spawns it in a way that subsequent
+  # invocations will not have the error. It only needs 1 retry.
+  return_code, stdout, stderr = execute.execute_and_filter_output(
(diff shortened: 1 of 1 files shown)


keith:
Out of curiosity what types of flakes do you see?

This is a tough one to me overall. This feels pretty against bazel's desire to be hermetic, but OTOH if the underlying tools aren't, what can you do.

jerrymarino:
@keith I agree and ideally the `actool` would be 100% reproducible. Until it is, I think the best experience is that the bazel rules have an affordance for the flake if it's prominent enough. I've heard from a handful of people had this so wanted to get a fix up.

I'm mainly targeting the `failed to open liblaunch_sim.dylib` error which is most prominent with many concurrent `actool` invocations for large catalogs.

```
INFO: From AssetCatalogCompile Some/xcassets:
--
  | /* com.apple.actool.errors */
  | /Some/Assets.xcassets: error: Failed to find newest available Simulator runtime
  | Underlying Errors:
  | Description: No available runtime found from: [
  | watchOS 7.0 (7.0 - 18R382) - com.apple.CoreSimulator.SimRuntime.watchOS-7-0 (unavailable, failed to open liblaunch_sim.dylib) ==> not available: Error Domain=com.apple.CoreSimulator.SimError Code=401 "The watchOS 7.0 simulator runtime is not available." UserInfo={NSLocalizedDescription=The watchOS 7.0 simulator runtime is not available., NSUnderlyingError=0x7fd1d6b10250 {Error Domain=NSPOSIXErrorDomain Code=53 "Software caused connection abort" UserInfo={NSLocalizedFailureReason=XPC error talking to SimLaunchHostService: <error: 0x7fff955749a0> { count = 1, transaction: 0, voucher = 0x0, contents =
  | "XPCErrorDescription" => <string: 0x7fff95574b08> { length = 18, contents = "Connection invalid" }
  | }}}, NSLocalizedRecoverySuggestion=Unable to open liblaunch_sim.dylib.  Try reinstalling Xcode or the simulator runtime., NSLocalizedFailureReason=failed to open liblaunch_sim.dylib},
  | tvOS 14.0 (14.0 - 18J383) - com.apple.CoreSimulator.SimRuntime.tvOS-14-0 ==> iPhone does not support runtime,
  | iOS 14.0 (14.0 - 18A372) - com.apple.CoreSimulator.SimRuntime.iOS-14-0 ==> not available: Error Domain=com.apple.CoreSimulator.SimError Code=401 "The iOS 14.0 simulator runtime is not available." UserInfo={NSLocalizedDescription=The iOS 14.0 simulator runtime is not available., NSUnderlyingError=0x7fd1d6b08650 {Error Domain=NSPOSIXErrorDomain Code=53 "Software caused connection abort" UserInfo={NSLocalizedFailureReason=XPC error talking to SimLaunchHostService: <error: 0x7fff955749a0> { count = 1, transaction: 0, voucher = 0x0, contents =
  | "XPCErrorDescription" => <string: 0x7fff95574b08> { length = 18, contents = "Connection invalid" }
  | }}}, NSLocalizedRecoverySuggestion=Unable to open liblaunch_sim.dylib.  Try reinstalling Xcode or the simulator runtime., NSLocalizedFailureReason=failed to open liblaunch_sim.dylib},
  | ] when matching for <IBCocoaTouchPlatformToolDescription: 0x7fd1b03318e0> System content for IBCocoaTouchFramework-fourteenAndLater <IBScaleFactorDeviceTypeDescription: 0x7fd1b0330cb0> scaleFactor=2x, renderMode.identifier=(null), idiom=<IBIPhoneIdiom: 0x7fd1a5688d10> runtime=<IBCocoaTouchTargetRuntime: 0x7fd1b03163f0>
```


PS: Another orthogonal idea I've been floating is to override Apple's actool with a reproducible one. There is an [open source effort](https://github.com/facebook/xcbuild/blob/master/Libraries/acdriver/Sources/Driver.cpp) to create a compatible tool which might be worth looking into at a later point.

thii:
It looks more like a `CoreSimulator` issue rather than an `actool` issue. We've seen errors like "Unable to resolve device type..." from `AssetCatalogCompile` actions, when someone started a build after downloading a new Xcode version but hadn't opened it, which have been solved by just opening it once.

jerrymarino:
Today I tested this out 50 times on Xcode 12.1 on fresh a macOS env each build w/o a build cache. I was able to repro `failed to open liblaunch_sim.dylib` in 23% of builds but it won't repro with this patch. I've added more info on the bug, why the retry mechanism works, and made a small improvement to output printing.

@thii agree - this seems like the same error. Does the `NSLocalizedRecoverySuggestion` above seem related? For me it happens locally and more on CI.

It seems like the broken code is within the dependent core sim services that `actool` uses. My only repro is with `actool` despite many other uses of simulators. I'm not an expert in core siumator but it seems related to `SimLaunchHostService` and `actool`. I haven't dug into it much more as retrying `actool` works 100% of the time on the latest Xcode releases for me in the latest builds of Xcode 12.

I still think working around it here is the best option overall. We could want better retry detection e.g. parsing the XML but it'd add complexity. Otherwise we can leave the issue open as documentation of how to fix it and wait for it to be fixed.

thii:
Found the error log. It was a different error, but might be related.

actool-error

jerrymarino:
@thii do you happen to remember if retrying fixed it? Maybe if there's many different errors it's better to just retry in the case of a non zero exit as it does, vs checking all possible error codes.

thii:
Ah, in our case, retrying did _not_ fix it, but opening Xcode once did. So it was a completely different error. Sorry for the confusion.

segiddins (approved):

brentleyjones (approved):

PR #956, opened 2020-10-21
Ignore missing executable_names

keith:
For some plists like those in resource bundles executable names should
not be included. In this case this can be called with
`include_executable_name = True` since that's the default, but
`executable_name = None`. In this case we can ignore this since that's
not a valid value.


brentleyjones (approved):

segiddins (approved):
Would be cool to add a test for the particular case this is trying to cover, if we have the infrastructure in place

keith:
no good place to add something like this atm

Issue #710, opened 2020-02-14
`skip_simulator_signing_allowed` should be `True` for `ios_unit_test`


brentleyjones:
In the [Codesign Bundles for the Simulator](https://github.com/bazelbuild/rules_apple/blob/master/doc/common_info.md#codesign-bundles-for-the-simulator-applecodesign_simulator_bundles) section of Common Info it states:

>One exception is XCTest bundles, those do need to be signed for the simulators to load them. The above --define does not change the behavior around signing of these bundles as a result.

This is incorrect. I changed `skip_simulator_signing_allowed` to `True` for `ios_unit_test` and all of our tests continue to pass, including our App Host test. I didn't have any `ios_ui_test` to check, but I believe that statement might continue to hold true for it.

keith:
Any harm in flipping this bit for everyone?

thomasvl:
It comes down to what the test uses as it if it needs to be signed.  There are some api even in the simulator that need to have things signed because they need to have entitlements.  When I did the initial work, it took a like of individual investigations to figure out when something failed because of the removal of the signature.  The APIs don't all error, they might return `nil`, so by the time the test fails, you can be well removed from the code path that made the call.

So changing the default means you likely put everyone into that case of having to go digging see why something started failing.  If you were starting from zero, then, yes, a different default might be easier as folks would hit the error when the added the tests that need to be signed.

I'll also note that tests tend not to be very granular, so any one test method could need it, making it harder to tell as the tests evolve if the flag state could be revisited.

Anyway, that's the background on why I originally stuck with off by default.

keith:
thanks for the context! I wonder if that means we should have an override in the opposite direction that allows opting into no singing instead of opting out.

brentleyjones:
Yeah, my main ask is the ability to disable code signing for XCTest bundles, whatever the method.

keith:
Here's a possible change that could replace the existing flag and be a bit more flexible around which targets you want to opt into this or not https://github.com/bazelbuild/rules_apple/pull/958

PR #958, opened 2020-10-21
Add apple.skip_codesign_simulator_bundles

keith:
This is a replacement for `apple.codesign_simulator_bundles` that uses
the newer `features` system. This has the benefit of being able to set
it on specific targets, and disable it for others. This way if you have
some targets that require signing, you can use:

```
features = ["-apple.skip_codesign_simulator_bundles"],
```

On those will still passing
`--features=apple.skip_codesign_simulator_bundles` if that works for
you.

Once the old option is removed there will be no filtering for when to
use this, so you can use it on whatever target types you want, and like
before it's up to you to manage the tradeoff between performance and
correctness.

Fixes https://github.com/bazelbuild/rules_apple/issues/710

### DIFF
--- a/apple/internal/codesigning_support.bzl
+++ b/apple/internal/codesigning_support.bzl
@@ -255,6 +255,7 @@ def _signing_command_lines(
 def _should_sign_simulator_bundles(
         *,
         config_vars,
+        features,
         rule_descriptor):
     """Check if a main bundle should be codesigned.
 
@@ -267,6 +268,13 @@ def _should_sign_simulator_bundles(
       True/False for if the bundle should be signed.
 
     """
+    if "apple.codesign_simulator_bundles" in config_vars:
+        # buildifier: disable=print
+        print("warning: --define apple.codesign_simulator_bundles is deprecated, please switch to --features=apple.skip_codesign_simulator_bundles")
+
+    if "apple.skip_codesign_simulator_bundles" in features:
+        return False
+
     if not rule_descriptor.skip_simulator_signing_allowed:
         return True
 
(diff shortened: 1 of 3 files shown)


keith:
Example cleanup after deprecation has baked https://github.com/bazelbuild/rules_apple/pull/959

brentleyjones:
Nice! Thanks for this!

apple/internal/codesigning_support.bzl
```diff
@@ -239,6 +239,12 @@ def _should_sign_simulator_bundles(ctx):
       True/False for if the bundle should be signed.
 
     """
+    if "apple.codesign_simulator_bundles" in ctx.var:
+        print("warning: --define apple.codesign_simulator_bundles is deprecated, please switch to --features apple.skip_codesign_simulator_bundles")
+
+    if "apple.skip_codesign_simulator_bundles" in ctx.features:
```
segiddins, line 245: we want to use proper feature configuration, so not_features is also taken into account

keith: ha yep nice call, fixed I believe

brentleyjones (approved):

steeve:
@keith why not sure starlark flags?

keith:
in this case features allows you to disable only for some targets as shown in the description

steeve:
Which is where I'm unsure where to draw the line between `define`, `features` and starlark flags...

brentleyjones:
@keith Tested this locally and it worked for us.

brentleyjones:
I think this is still a good change. Think we can land it?

brentleyjones:
@googlebot I consent.

brentleyjones (approved):

apple/internal/codesigning_support.bzl
```diff
@@ -325,8 +350,11 @@ def _codesigning_args(
     is_device = platform_prerequisites.platform.is_device
     should_sign_sim_bundles = _should_sign_simulator_bundles(
         config_vars = platform_prerequisites.config_vars,
+        features = platform_prerequisites.features,
         rule_descriptor = rule_descriptor,
     )
+
+    # We need to re-sign imported frameworks
```
brentleyjones, line 357: I think we need to keep signing for imported frameworks, in case of incorrect signing on the existing framework?

thii (approved):

PR #965, opened 2020-10-30
Disable --incompatible_run_shell_command_string for now to fix incompatibility with Bazel at HEAD.

thii:
Disable this for now because apple_support currently relies on a
workaround that needs to be able to pass a sequence of strings to
actions.run_shell's 'command'.


brentleyjones (approved):
:+1:

PR #967, opened 2020-10-30
Properly handle Unicode characters that come out of `ibtool`.

keith:
In some cases, `ibtool` emits messages with curly quotes:

```
warning: “SomeViewController“ is unreachable because it has no entry points, and no identifier for runtime access via -[UIStoryboard instantiateViewControllerWithIdentifier:].
```

If `sys.{stdout,stderr}` have a default encoding of `ascii`, the later operation to write the Unicode strings to these streams will fail unless we reopen the stream with `utf8` encoding.

PiperOrigin-RevId: 339937656
(cherry picked from commit 14bfbaabda19ca31492157db0c05b508f9993a75)

### DIFF
--- a/tools/wrapper_common/execute.py
+++ b/tools/wrapper_common/execute.py
@@ -16,6 +16,7 @@
 """Common functionality for tool wrappers to execute jobs.
 """
 
+import io
 import os
 import re
 import subprocess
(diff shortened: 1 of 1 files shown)


segiddins (approved):

thii:
The build error here can be fixed by https://github.com/bazelbuild/rules_apple/pull/968.

PR #968, opened 2020-10-31
Set bitcode_strip wrapper's print_output to False

thii:
As we're not doing any post-processing of its output here. This fixes
the incompatibility with the change in
https://github.com/bazelbuild/rules_apple/pull/967.


keith:
Thanks!

PR #979, opened 2020-11-05
Add apple.strip_swift_symbols feature

thii:
This is equivalent to Xcode's `STRIP_SWIFT_SYMBOLS` setting, if enabled
all Swift symbols from Swift standard libraries and imported dynamic
frameworks will be removed.


apple/internal/partials/framework_import.bzl
```diff
@@ -306,6 +310,7 @@ def _generate_symbols(
 def framework_import_partial(
         *,
         actions,
+        features = [],
```
thii, line 313: I don’t think this should have a default value here but this reduces the patch amount a bit, as we don’t have a need for this feature for non-iOS platforms.

thii:
@brentleyjones Do you have any ideas how to test this?

brentleyjones:
Could we use `binary_contents_test` for both sides of the flag to check for the symbols? Resulting in at least 4 things tested:
- dylibs not stripped
- dylibs stripped
- imported framework not stripped
- imported framework stripped

keith:
How does this interplay with bazel's native strip logic?

thii:
Bazel strips the built target's binary (built by Bazel), where this is for the Swift standard libraries and imported dynamic frameworks. And it looks like it doesn't strip Swift symbols (there's no `-T` flag): https://cs.opensource.google/bazel/bazel/+/c282526c071389cd6f88cb77565283b257316267:src/main/java/com/google/devtools/build/lib/rules/objc/CompilationSupport.java;l=1574-1597;drc=b38680a0ee3822e32adb4a63a29106bf64643718

Looking again at Xcode's documentation and our build logs, it seems the `STRIP_SWIFT_SYMBOLS` build setting and the "Strip Swift Symbols" export option are completely different things. The former is used for stripping the product when archiving, while the latter is used _only_ for stripping the Swift standard libraries when exporting an ipa.

Symbols striping of imported frameworks in Xcode is controlled by another build setting called `COPY_PHASE_STRIP`.

> I think in general I would prefer that we don't add options until they have a solid use case. Also that reduces the shear number of options people have to know about, and reduces the diff from upstream. So in general if we can do the "right" thing for everyone I would prefer that until it's no longer the case.

Agreed. I think it's fine to make this the default (same with Xcode), but it'd be better to have an option to allow users to opt-out, since it might cause [problems](https://developer.apple.com/library/archive/releasenotes/DeveloperTools/RN-Xcode/Chapters/Introduction.html):

> The new Strip Swift Symbols (STRIP_SWIFT_SYMBOLS) build setting is enabled by default. It adjusts the level of symbol stripping so that when the linked product of the build is stripped, all Swift symbols are removed. This significantly reduces the size of Swift frameworks. If the lack of Swift symbols causes problems, such as when using dladdr(), this setting can be disabled. To view the exported symbols from file that has been stripped, use xcrun dyldinfo -export instead of nm. (31306055)

A side issue is that, since `rules_apple` doesn't yet have a full-fledge feature configuration support like `rules_swift`, so in order to make this enabled by default, we would have to name it a negative naming like `apple.no_strip_swift_symbols` 😢?

thii:
Update: Implemented the default features support.

rsahara:
@thii Can this PR be merged or does it need some other fixes? (Wondering if I should create a patch or wait for this to be merged.)

thii:
I'll try to get back to this and finish it soon, hopefully within next week.

apple/internal/partials/swift_dylibs.bzl
```diff
@@ -191,6 +202,7 @@ def _swift_dylibs_partial_impl(
                 platform_prerequisites = platform_prerequisites,
                 resolved_swift_stdlib_tool = apple_toolchain_info.resolved_swift_stdlib_tool,
                 strip_bitcode = strip_bitcode,
+                strip_swift_symbols = strip_swift_symbols,
```
rsahara, line 205: JFYI, I tested this on our TestFlight builds.
Enabling `strip_swift_symbols` in the `swift_dylibs` made the TestFlight validation fail, but enabling it just in `framework_import` was OK.

```
ITMS-90428: Invalid Swift Support - The files libswiftos.dylib, libswiftARKit.dylib, ... (long list)
```

keith: What was the end of the error message?

rsahara: Oops sorry.

```
ITMS-90428: Invalid Swift Support - The files libswiftos.dylib, (long list), don’t match /Payload/(app name).app/Frameworks/libswiftos.dylib, (long list).
```

PR #986, opened 2020-11-15
Add bundle_package_type bzl library

BalestraPatrick:
I encountered a failure related to the `bundle_package_type.bzl` import not being found when trying to run stardoc in some of our rules. I tested this locally and it seems to fix. I assume this was forgot in #891.


keith:
Thanks!

PR #987, opened 2020-11-16
Add ios_dylib rule

brentleyjones:
I need to add tests, but we use this internally to create SwiftUI Preview thunks, and others have expressed interest in having this rule, so I thought I would open the PR earlier.

### DIFF
--- a/apple/internal/ios_rules.bzl
+++ b/apple/internal/ios_rules.bzl
@@ -23,6 +23,10 @@ load(
     "bundling_support",
 )
 load(
+    "@build_bazel_rules_apple//apple/internal:codesigning_support.bzl",
+    "codesigning_support",
+)
+load(
     "@build_bazel_rules_apple//apple/internal:entitlements_support.bzl",
     "entitlements_support",
 )
@@ -72,6 +76,7 @@ load(
 )
 load(
     "@build_bazel_rules_apple//apple:providers.bzl",
+    "AppleBinaryInfo",
     "IosAppClipBundleInfo",
     "IosApplicationBundleInfo",
     "IosExtensionBundleInfo",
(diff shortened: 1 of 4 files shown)


apple/ios.bzl
```diff
@@ -79,6 +80,37 @@ def ios_app_clip(name, **kwargs):
...
+        product_type = apple_product_type.dylib,
+        link_swift_statically = True,
+        include_entitlements = False,
+        exported_symbols_lists = binary_args.pop("exported_symbols_lists", None),
+        **binary_args
+    )
+
+    _ios_dylib(
+        name = name,
+        dylibs = binary_args.pop("frameworks", None),
```
segiddins, line 110: why would this be passed in via `frameworks`?

brentleyjones: This allows specifying `ios_framework` deps, the same as `ios_application`.

apple/internal/ios_rules.bzl
```diff
@@ -542,6 +547,75 @@ def _ios_app_clip_impl(ctx):
...
+    output_file = ctx.actions.declare_file(ctx.label.name + ".dylib")
+    codesigning_support.sign_binary_action(ctx, binary_artifact, output_file)
+
+    return [
+        AppleBinaryInfo(
+            binary = output_file,
+            product_type = ctx.attr._product_type,
+        ),
+        DefaultInfo(files = depset(transitive = [
+            depset([output_file]),
```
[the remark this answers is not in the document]
brentleyjones: This would diverge from `_macos_dylib_impl` then. If we fix it there first I can here. I just want them to match as much as possible to allow for a shared implementation later.

apple/ios.bzl
keith: like the dylib can have frameworks?

apple/internal/ios_rules.bzl
```diff
@@ -542,6 +547,75 @@ def _ios_app_clip_impl(ctx):
...
+        executable_name = executable_name,
+        partials = [debug_outputs_partial],
+        platform_prerequisites = platform_prerequisites,
+        predeclared_outputs = predeclared_outputs,
+        provisioning_profile = getattr(ctx.file, "provisioning_profile", None),
+        rule_descriptor = rule_descriptor,
+        rule_executables = rule_executables,
+        rule_label = label,
+    )
+    output_file = ctx.actions.declare_file(ctx.label.name + ".dylib")
```
keith, line 598:
```suggestion
    output_file = actions.declare_file(ctx.label.name + ".dylib")
```

this is what macos_dylib does, seems like with this and the change below there can be 0 diff between them

apple/internal/ios_rules.bzl
```diff
@@ -542,6 +547,75 @@ def _ios_app_clip_impl(ctx):
...
+    codesigning_support.sign_binary_action(ctx, binary_artifact, output_file)
+
+    return [
+        AppleBinaryInfo(
+            binary = output_file,
+            product_type = ctx.attr._product_type,
```
keith, line 604:
```suggestion
            product_type = rule_descriptor.product_type,
```

this is what macos_dylib does, seems like with this and the change above there can be 0 diff between them

keith:
Are the dylibs produced from this valid? Looks to me like their install names are incorrect:

```
bazel-bin/Modules/Logger/libLogger.dylib:
        bazel-out/ios-x86_64-min11.0-applebin_ios-ios_x86_64-fastbuild-ST-7786d55448f4/bin/Modules/Logger/libLogger_bin (compatibility version 0.0.0, current version 0.0.0)
```

keith:
huh looks like the macos_dylib rule produces one with a similar name, but I don't think that's right? Doesn't that need to be `@loader_path/Frameworks/libFoo.dylib` or something similar?

keith:
It kinda seems like right now you can't use this how you might want by bundling a tree of dylibs into a ios_unit_test, in these cases it seems that you're fighting between what attr to put them in first, such as `dylibs`, and then you end up statically linking the transitive dependencies still.

brentleyjones:
Yeah, it works for our specific use case of making a single dylib for use with SwiftUI Previews (and an XCBBuildService proxy). It mirrors how the macOS one works. My use case is taking the resulting binary directly, so I didn't look into bundling inside another app yet, but I can look into that and make sure it works. As for the transitive dependencies, we don't hit that in our use case so far, so I don't have many opinions about it yet.

brentleyjones:
> As for the transitive dependencies, we don't hit that in our use case so far, so I don't have many opinions about it yet.

I assume we should be able to make it work though, and ideally just through `deps` instead of `frameworks`.

apple/ios.bzl
brentleyjones: It can depend on them, yes.

keith:
I can't think of another example where a packaging rule can be directly depended on via `deps` like this, and I assume this is why apple_binary has a `dylibs` attr 🤔

Issue #993, opened 2020-11-20
rules_apple failing on Bazel CI due to the removal of --incompatible_objc_compile_info_migration


meteorcloudy:
https://buildkite.com/bazel/rules-apple-darwin/builds/3560#c1d136b3-df02-4c04-804b-7b70c550749a
```
-- Test log: -----------------------------------------------------------
Executing: bazel build --announce_rc --symlink_prefix=test- --verbose_failures --define=RULES_SWIFT_BUILD_DUMMY_WORKER=1 --use_top_level_targets_for_symlinks --incompatible_merge_genfiles_directory --incompatible_objc_compile_info_migration --xcode_version=11.7 --compilation_mode opt --ios_multi_cpus=i386,x86_64 --strategy=SwiftCompile=local --objccopt=-Werror --objccopt=-Wunused-command-line-argument --objccopt=-Wno-unused-function --objccopt=-Wno-format --objccopt=-Wno-unused-variable //app:app
$TEST_TMPDIR defined: output root default is '/private/var/tmp/_bazel_buildkite/d5733b9d189ba727a8f29b2bad8cbac0/execroot/build_bazel_rules_apple/_tmp/2bd68464f88dc7938f566aba37879387' and max_idle_secs default is '15'.
ERROR: --incompatible_objc_compile_info_migration :: Unrecognized option: --incompatible_objc_compile_info_migration
```

Caused by: https://github.com/bazelbuild/bazel/commit/bcdd55d8956e22758fe3d866427e2dc3ebc4e31b

Please remove the deleted incompatible flag from the test.

keith:
https://github.com/bazelbuild/rules_apple/pull/994

keith:
ha removing that breaks our build for the most recent release, I'll think about a solution

meteorcloudy:
So removing an incompatible flag is technically an incompatible change 😂
Thanks for looking into it!

PR #994, opened 2020-11-20
Remove incompatible flag from test

keith:
Fixes https://github.com/bazelbuild/rules_apple/issues/993

### DIFF
--- a/test/apple_shell_testutils.sh
+++ b/test/apple_shell_testutils.sh
@@ -451,6 +451,12 @@ function do_action() {
       "--incompatible_merge_genfiles_directory"
   )
 
+  local bazel_version="$(bazel --version)"
+  local bazel_major_version="$(echo "${bazel_version#bazel }" | cut -f1 -d.)"
+  if [[ ( "${bazel_major_version}" != "no_version" ) && ( "${bazel_major_version}" -lt 4 ) ]]; then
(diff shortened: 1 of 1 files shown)


segiddins:
Do we have any way to check the running bazel version in those shell tests?

segiddins:
Added a commit that will hopefully smooth things over for bazel 3, @keith / @brentleyjones if y'all want to take another look

segiddins:
@googlebot I consent

test/apple_shell_testutils.sh
```diff
@@ -449,9 +449,14 @@ function do_action() {
       # Explicitly pass these flags to ensure the external testing infrastructure
       # matches the internal one.
       "--incompatible_merge_genfiles_directory"
-      "--incompatible_objc_compile_info_migration"
   )
 
+  local bazel_version="$(bazel --version)"
```
brentleyjones, line 454: As part of the LTS support, didn't they say there was a way to know which ball version was being used?

As for this, if I run with a non-path bazel, this check will be wrong. Is there a way to get the path to the bazel that is running this?

segiddins: further down, we already invoke bazel directly via `bazel`

segiddins (approved):

brentleyjones (approved):

PR #999, opened 2020-12-01
Test CI

keith:


keith:
depends on https://github.com/bazelbuild/rules_apple/pull/996 to pass

PR #1007, opened 2020-12-16
Fix duplicate module error when the intermediates directory for ios_static_framework accidentally goes into the header search paths, when building with local strategy.

thii:
This patch changes the path of the intermediate module map by placing it
under a subdirectory of the intermediate directory. Since there's no way
one could add the intermediates directory into a compile action's header
search paths, this prevents the module map from being implicitly
imported when it's produced by a previous invocation.


thii:
The build with last green Bazel failed because it can't be downloaded now.

thii:
https://github.com/bazelbuild/continuous-integration/issues/1074

brentleyjones (approved):

PR #1017, opened 2021-01-05
Fix simulator running with tree artifacts

keith:
The tree artifacts bundler correctly sets the permissions of the app and
its nested directories to be writable but then bazel resets them to not
be writable. This doesn't apply for the ipa codepath since the
permissions are retained in the zip. This fixes this by copying the
output .app from bazel into a tempdir and chmoding all the nested
directories. Without this the `simctl install` fails.

### DIFF
--- a/apple/internal/templates/ios_sim.template.py
+++ b/apple/internal/templates/ios_sim.template.py
@@ -51,6 +51,7 @@ import os
 import os.path
 import platform
 import plistlib
+import shutil
 import subprocess
 import tempfile
 import time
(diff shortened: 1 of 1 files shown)


apple/internal/templates/ios_sim.template.py
```diff
@@ -417,7 +418,14 @@ def extracted_app(ios_application_output_path, app_name):
   """
   if os.path.isdir(ios_application_output_path):
     logger.debug("Found app directory: %s", ios_application_output_path)
-    yield os.path.realpath(ios_application_output_path)
+    with tempfile.TemporaryDirectory(prefix="bazel_temp") as temp_dir:
+      temp_app_path = os.path.join(temp_dir, app_name + ".app")
+      shutil.copytree(ios_application_output_path, temp_app_path)
+      for root, dirs, _ in os.walk(temp_app_path):
+        for directory in dirs:
+          os.chmod(os.path.join(root, directory), 0o777)
```
brentleyjones, line 426: Is it faster to just blanket apply 0x777 to all entities recursively? Does it break anything?

keith: I assume there will be more files than directories so that would end up doing a lot more `chmod` calls. I imagine it _could_ break something but I'm not sure how risky it actually would be.

brentleyjones: Ahh, there isn't a `chmod -R` version, never mind.

brentleyjones (approved):

PR #1022, opened 2021-01-07
Loosen the restriction when making a static framework with Swift.

keith:
PiperOrigin-RevId: 349566308
(cherry picked from commit 0bd7b9db49c3588bc83260b35baccb81a6cfbbc8)


keith:
diff best viewed w/o whitespace

PR #1027, opened 2021-01-07
Remove plist and strings handling at the rule implementation level for settings bundles.

keith:
PiperOrigin-RevId: 350244630
(cherry picked from commit 83d81ea39b90f8a76c7e0f9f1131697f5cab51d5)


keith:
to be safe I think we should merge https://github.com/bazelbuild/rules_apple/pull/988 first

Issue #1019, opened 2021-01-06
codesign error when using apple_resource_group


joprice:
I am getting the error below when building an app that uses apple_resource_group:

```nd-sign-2048736916.sh failed: error executing command bazel-out/applebin_ios-ios_x86_64-fastbuild-ST-c4cf2ec641b4/bin/ios-app/ios-app-intermediates/process-and-sign-2048736916.sh should_sign
Traceback (most recent call last):
  File "/private/var/tmp/_bazel_josephprice/33e38ece9d552fdf1983387d9ac53815/execroot/__main__/bazel-out/host/bin/external/build_bazel_rules_apple/tools/codesigningtool/codesigningtool.runfiles/build_bazel_rules_apple/tools/codesigningtool/codesigningtool.py", line 301, in <module>
    sys.exit(main(generate_arg_parser().parse_args()))
  File "/private/var/tmp/_bazel_josephprice/33e38ece9d552fdf1983387d9ac53815/execroot/__main__/bazel-out/host/bin/external/build_bazel_rules_apple/tools/codesigningtool/codesigningtool.runfiles/build_bazel_rules_apple/tools/codesigningtool/codesigningtool.py", line 296, in main
    _invoke_codesign(args.codesign, identity, args.entitlements, args.force,
  File "/private/var/tmp/_bazel_josephprice/33e38ece9d552fdf1983387d9ac53815/execroot/__main__/bazel-out/host/bin/external/build_bazel_rules_apple/tools/codesigningtool/codesigningtool.runfiles/build_bazel_rules_apple/tools/codesigningtool/codesigningtool.py", line 74, in _invoke_codesign
    _, stdout, stderr = execute.execute_and_filter_output(cmd,
  File "/private/var/tmp/_bazel_josephprice/33e38ece9d552fdf1983387d9ac53815/execroot/__main__/bazel-out/host/bin/external/build_bazel_rules_apple/tools/codesigningtool/codesigningtool.runfiles/build_bazel_rules_apple/tools/wrapper_common/execute.py", line 106, in execute_and_filter_output
    raise subprocess.CalledProcessError(proc.returncode, cmd_args)
[...]

PR #1029, opened 2021-01-07
Validate for all case-variants of invalid top-level directories in bundles

thii:
Previously, we only checked for the non-existent top level `Resources`
folder, but not other variants like `resources`. If the developer uses
any of those names, the validation would pass and then fail in the
codesign action.

Since the build may be running on a case-insensitive file-system, which
is the default file-system for macOS, the validation should check for
all the case variants of invalid directory names.

Fixes https://github.com/bazelbuild/rules_apple/issues/1019.


brentleyjones (approved):

brentleyjones:
Thanks @thii!

PR #1030, opened 2021-01-08
Thread codesignopts through without ctx

keith:
Previously we were reading this only in the place it was used directly
from the ctx. There's an ongoing refactor to remove ctx from all these
rules (which is almost done). Because of this we have to pass this value
down instead.


apple/internal/ios_rules.bzl
```diff
@@ -290,6 +290,7 @@ def _ios_application_impl(ctx):
         rule_descriptor = rule_descriptor,
         rule_executables = rule_executables,
         rule_label = label,
+        codesignopts = [ctx.expand_make_variables("codesignopts", opt, {}) for opt in ctx.attr.codesignopts],
```
keith, line 293: the variable expansion here is pretty annoying, im kinda surprised we're not doing this anywhere else? maybe it isn't necessary?

apple/internal/rule_factory.bzl
```diff
@@ -637,6 +632,11 @@ fashion, such as a Cocoapod.
                 doc = """
 A list of iOS app clips to include in the final application bundle.
 """,
+            ),
+            "codesignopts": attr.string_list(
```
keith, line 636: I moved this in this RFC to restrict this attr to only ios_application. Not saying that's good but otherwise it would require much more threading

keith:
Best viewing just the first commit, still debugging why the second was needed for tests to pass locally for me

apple/internal/codesigning_support.bzl
```diff
@@ -592,6 +602,7 @@ def _sign_binary_action(
         paths_to_sign = [path_to_sign],
         platform_prerequisites = platform_prerequisites,
         provisioning_profile = provisioning_profile,
+        codesignopts = [],
```
segiddins, line 605: does it also need to be threaded through to this action?

apple/internal/rule_factory.bzl
segiddins: is this technically a breaking change?

apple/internal/codesigning_support.bzl
keith: this one doesn't come from a codepath that can have one atm, so i just ignored it

apple/internal/rule_factory.bzl
keith: def. not particularly suggesting we do this, but when I made it I was thinking about the balance of realistic use cases vs maintainability

apple/internal/ios_rules.bzl
thii: This was a workaround because there's nothing like `codesign_inputs`, so `expand_locations` doesn't work. This works because codesigning is always done outside of a sandbox.

keith: nice. i guess we'd prob need to abstract this if we did this on every possible rule

apple/internal/rule_factory.bzl
brentleyjones: If something is needed by ios_application, it minimally is needed by ios_app_clip as well, and most likely the app extensions too. I think we need to keep it where it was 😕.

apple/internal/processor.bzl
```diff
@@ -440,7 +440,8 @@ def _bundle_post_process_and_sign(
         provisioning_profile,
         rule_descriptor,
         rule_executables,
-        rule_label):
+        rule_label,
+        codesignopts):
```
brentleyjones, line 444: Nit: can this get moved up to below `bundle_name`?

apple/internal/processor.bzl
```diff
@@ -632,7 +637,8 @@ def _process(
         provisioning_profile,
         rule_descriptor,
         rule_executables,
-        rule_label):
+        rule_label,
+        codesignopts = []):
```
brentleyjones, line 641: Can this get moved up, and can/should we remove the default?

apple/internal/codesigning_support.bzl
```diff
@@ -60,6 +60,7 @@ def _codesign_args_for_path(
         path_to_sign,
         platform_prerequisites,
         provisioning_profile,
+        codesignopts,
```
brentleyjones, line 63: Nit: can this be moved up?

brentleyjones (commented):
I think this threading is needed, sadly. I don't think it's that much of an additional maintenance burden, once it's in place.

apple/internal/codesigning_support.bzl
keith: What ordering where you hoping for? I felt like there wasn't a super clear order to go for so I didn't try for one

keith: ah this group is alphabetical, I'll move

apple/internal/ios_rules.bzl
keith: Note for remote exec to be compatible with this we might eventually need codesign_inputs as well, a problem for another day

thii (approved):

brentleyjones (approved):

PR #1031, opened 2021-01-08
Exclude Swift modules in the transitive closure of `avoid_deps` when verifying that an `{ios,tvos}_static_framework` contains only a single Swift module.

keith:
This allows users to build multiple static frameworks that have Swift modules that depend on each other, as long as the end user includes all of the frameworks in their search paths when they build.

PiperOrigin-RevId: 350785087
(cherry picked from commit 8c2494e616bc6ee2fbd2676ef9b0d47a11e7fe7e)

### DIFF
--- a/apple/internal/aspects/BUILD
+++ b/apple/internal/aspects/BUILD
@@ -44,6 +44,7 @@ bzl_library(
         "//apple/internal:__pkg__",
     ],
     deps = [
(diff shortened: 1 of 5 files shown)


keith:
Closes https://github.com/bazelbuild/rules_apple/pull/660

Issue #895, opened 2020-09-02
Support for CFBundleIcons.CFBundleAlternateIcons


steeve:
This `Info.plist` key is used to implement alternate icons.

Unfortunately at this moment, trying to set that key will result in `plisttool.py` complaining when merging `CFBundleIcons`. So it seems not possible to use dynamic icons with `rules_apple` at this moment.

Example:
```xml
<key>CFBundleIcons</key>
<dict>
	<key>CFBundleAlternateIcons</key>
	<dict>
		<key>MyOtherIcon</key>
		<dict>
			<key>CFBundleIconFiles</key>
			<array>
				<string>MyOtherAppIcon</string>
			</array>
			<key>UIPrerenderedIcon</key>
			<false/>
		</dict>
	</dict>
	<key>CFBundlePrimaryIcon</key>
	<dict>
		<key>CFBundleIconFiles</key>
		<array>
			<string>MyAppIcon</string>
		</array>
		<key>CFBundleIconName</key>
		<string>MyIcon</string>
	</dict>
</dict>
```

I was thinking about something like this:
```python
app_icon = glob(["Resources/Assets/AppIcon.xcassets/**"]),
alternate_icons = {
    "MyOtherIcon": glob(["Resources/Assets/OtherAppIcon.xcassets/**"]),
    "MyThirdIcon": glob(["Resources/Assets/ThirdAppIcon.xcassets/**"]),
},
```

keith:
A current workaround is to add generate this plist with a custom rule and provide the icons as data. I wonder if making a separate rule like this would make more sense vs adding it to *application 🤔🤔

steeve:
Another idea would be make `ios_application` a macro and make `app_icon` accept a dict when one wants alternate icons, with `""` or `None` as primary.

steeve:
Actually, I just tried skipping the `app_icon` parameter, passing my own `Info.plist` and adding the icons as `resources`, `plisttool` still complains. Will look deeper a bit later.

Another short term solution is to do it during the post process phase.

steeve:
Another weirdness, if our case, the `plist` was generated by Xcode and already contained icon definition like so (`AppIconEnterprise` is an `appiconset`):
```json
{
  "CFBundlePrimaryIcon": {
    "CFBundleIconFiles": [
      "AppIconEnterprise"
    ],
    "CFBundleIconName": "AppIconEnterprise"
  }
}
```

But the generated one by rules_apple looks like so:
```json
{
[...]

PR #1032, opened 2021-01-08
Support for Alternate Icons

steeve:
This is a draft PR for alternate icons support. There are no tests since I want to validate the approach first.

Because alternate icons unfortunately can't be packaged as `appiconset`, alternate icons must be directories in the form of `<alticon_id>.alticon`, such as `AppIcon-Dark.alticon` if `AppIcon-Dark` is the id. The `PNGs` inside can be named freely.

For instance, with an alticon id as `Dark`:
```
MyApp/Resources/Dark.alticon/
├── AppIcon-Dark@2x.png
├── AppIcon-Dark@2x~ipad.png
├── AppIcon-Dark@2x~ipadpro.png
└── AppIcon-Dark@3x.png
```

They can be added via the `ios_application.alternate_icons` attribute. The icons are processed like any other `png` file and will be embedded as a top level resource.

Some implementation notes:
Since there is no `string_keyed_label_list_dict`, the `alticon` trick is a way to both group the icons together and explicit the icon id.

Also, since all intermediate plists are merged at the top level, `CFBundleIcons.CFBundleAlternateIcons` needs to be generated in the same partial as `CFBundleIcons.CFBundlePrimaryIcon`. This is done by detecting if there are alternate icons, and if so, reinject the `plist` into `alticonstool`, which will modify it accordingly.

Refs #895


thii:
> Since there is no string_keyed_label_list_dict, the alticon trick is a way to both group the icons together and explicit the icon id.

You can use `label_keyed_string_dict` and swap their keys and values, and require each icon to be grouped to a label. I recall `rules_nixpkgs` was doing something similar in their implementation.

steeve:
True, although I think defining a directory is easier for organising files and globing them all in one go.
Also, the way it works is they have to be aggregated into resources for later parsing/dispatching on their nature. Thus, the filename actually has a meaning, not so much the rule's attribute.

tinder-maxwellelliott:
FWIW, this branch is working great for me locally and is able to load all alternate icons seamlessly

steeve:
Thank you all for testing. I'm busy with work at the moment, I'll rebase and proper this PR in the next week or two (since we need it too).

steeve:
I have rebased the PR on master

steeve:
Rebased, and CI is finally green! Thus, gentle ping for you guys @thii and @keith.

segiddins (approved):

tools/alticonstool/alticonstool.py
```diff
@@ -0,0 +1,70 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+import argparse
+import os
+import sys
+import plistlib
```
thii, line 20: I think this wouldn’t run with PY2 as the host Python, but we can revisit this later if there are requests for PY2 support.

tools/alticonstool/BUILD
```diff
@@ -0,0 +1,21 @@
+licenses(["notice"])
+
+py_binary(
+    name = "alticonstool",
+    srcs = ["alticonstool.py"],
+    srcs_version = "PY2AND3",
```
thii, line 6:
```suggestion
    srcs_version = "PY3",
```

I've just tried this branch with `--host_force_python=PY2` and looks like setting this to `PY3` gives us a slightly better error message:

> ERROR: /Users/user/rules_apple/test/starlark_tests/targets_under_test/ios/BUILD:290:16 Bundling app_with_alternate_icons failed: //tools/alticonstool:alticonstool: This target is being built for Python 2 but (transitively) includes Python 3-only sources. You can get diagnostic information about which dependencies introduce this version requirement by running the `find_requirements` aspect. If this is used in a genrule, you may need to migrate from tools to exec_tools. For more info see the documentation for the `srcs_version` attribute: https://docs.bazel.build/versions/master/be/python.html#py_binary.srcs_version caused by //tools/alticonstool:alticonstool

With `PY2AND3`, it would fail at `xctoolrunner`:

> ERROR: /Users/user/rules_apple/test/starlark_tests/targets_under_test/ios/BUILD:290:16: AssetCatalogCompile test/starlark_tests/targets_under_test/ios/app_with_alternate_icons-intermediates/xcassets failed: (Exit 1): xctoolrunner failed: error executing command bazel-out/host/bin/tools/xctoolrunner/xctoolrunner actool --compile ... (remaining 20 argument(s) skipped)
  File "/private/var/tmp/_bazel_user/c481fd6bdd7c10e4da1c8594f7656dc0/execroot/build_bazel_rules_apple/bazel-out/host/bin/tools/xctoolrunner/xctoolrunner.runfiles/build_bazel_rules_apple/tools/xctoolrunner/xctoolrunner.py", line 44, in <module>
    from build_bazel_rules_apple.tools.wrapper_common import execute
ImportError: No module named build_bazel_rules_apple.tools.wrapper_common
Note: The failure of target //tools/xctoolrunner:xctoolrunner (with exit code 1) may have been caused by the fact that it is a Python 3 program that was built in the host configuration, which uses Python 2. You can change the host configuration (for the entire build) to instead use Python 3 by setting --host_force_python=PY3.

> If this error started occurring in Bazel 0.27 and later, it may be because the Python toolchain now enforces that targets analyzed as PY2 and PY3 run under a Python 2 and Python 3 interpreter, respectively. See https://github.com/bazelbuild/bazel/issues/7899 for more information.
Target //test/starlark_tests/targets_under_test/ios:app_with_alternate_icons failed to build

steeve: I guess the issue isn't in alticontool but rather in xctoolrunner. It's may be fine to make it `PY3` anyway, after all most tools are already.

steeve: Ah well
```
py_binary(
    name = "xctoolrunner",
    srcs = ["xctoolrunner.py"],
    python_version = "PY3",
    visibility = [
        "//apple/internal:__pkg__",
    ],
    deps = [":xctoolrunner_lib"],
)
```

steeve:
Great feedback @thii, thank you, updated the PR

apple/internal/resource_actions/actool.bzl
```diff
@@ -208,12 +239,24 @@ def compile_asset_catalog(
...
 
-    outputs = [output_dir]
+    alticons_outputs = []
+    actool_output_plist = None
+    actool_outputs = [output_dir]
     if output_plist:
-        outputs.append(output_plist)
+        if alternate_icons:
+            alticons_outputs = [output_plist]
+            actool_output_plist= intermediates.file(
```
thii, line 248: Formatting (not sure why buildifier can’t catch this).
```suggestion
            actool_output_plist = intermediates.file(
```

steeve: Not my proudest moment

thii (approved):

steeve:
Thank you @thii !

PR #1036, opened 2021-01-13
Fix swift_stdlib_tool remote-exec failure

brentleyjones:
The `destination_path` might not exist on the worker. It seems that `swift_stdlib_tool` currently works because bazel implicitly creates the directory locally when `declare_directory()` is used. To fix this we need to ensure the output directory is created in `swift_stdlib_tool`.


keith (approved):

brentleyjones:
Similar to https://github.com/bazelbuild/rules_apple/pull/896. After re-reading that, I'll update this to work with Python 2 as well.

PR #1042, opened 2021-01-22
Remove dynamic framework requirement for generated_header

brentleyjones:
A swift framework doesn't need to have a generated header (and can be disabled with `--features=swift.no_generated_header`). Without this header the framework can't be consumed by Objective-C, but otherwise it's fine.

### DIFF
--- a/apple/internal/aspects/swift_dynamic_framework_aspect.bzl
+++ b/apple/internal/aspects/swift_dynamic_framework_aspect.bzl
@@ -92,9 +92,15 @@ def _swift_dynamic_framework_aspect_impl(target, ctx):
     if not swiftdeps:
         return []
 
+    # Get the generated_header using the CcInfo provider
+    generated_header = None
+    for dep in ccinfos:
+        headers = dep[CcInfo].compilation_context.headers.to_list()
+        if headers:
+            generated_header = headers.pop(0)
+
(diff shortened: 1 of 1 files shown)


brentleyjones:
@mccorkill1 Hopefully this doesn't break your use case for `ios_dynamic_framework`. This is needed for `ios_framework` though.

mccorkill1:
This looks good. I don't think it'll break anything for the ios_dynamic_framework rule.

apple/internal/aspects/swift_dynamic_framework_aspect.bzl
```diff
@@ -133,7 +133,7 @@ single swift_library dependency.\
             generated_header = headers.pop(0)
 
     # Make sure that all dictionaries contain at least one module before returning the provider.
-    if all([module_name, generated_header, swiftdocs, swiftmodules, modulemap_file]):
```
keith, line 136: Doesn't the modulemap not end up existing in this case too?

brentleyjones: It still existed in for us.

brentleyjones: Ahh, it shouldn't be created. I'll update that.

keith (approved):

PR #1052, opened 2021-01-28
Make apple_verification_test publicly consumable

keith:
These rules allow you to write tests against the final product of a
build. This can be useful to verify that certain things about your
release application are true. This makes a small number of changes to
make this usable from outside rules_apple.

### DIFF
--- a/test/BUILD
+++ b/test/BUILD
@@ -61,7 +61,7 @@ filegroup(
         "apple_shell_testutils.sh",
         "unittest.bash",
     ],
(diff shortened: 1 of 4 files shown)


keith:
Note: there are a few more scripts that might also need to be exported if we wanted

thii (approved):

brentleyjones (approved):
I really like these tests, and think they are generally useful

segiddins (approved):

PR #1053, opened 2021-01-28
Refactor `.symbols` files generation code path to allow filtering of the symbols tool's output

thii:


tools/symbols_tool/symbols_tool.py
```diff
@@ -0,0 +1,99 @@
...
+           "-noTextInSOD",
+           "-noDaemon",
+           "-arch",
+           arch,
+           "-symbolsPackageDir",
+           output_dir,
+           binaries[i]]
+    execute.execute_and_filter_output(
+        cmd,
+        filtering=_filter_symbols_tool_output,
```
brentleyjones, line 48: Is there a way to have a test for this tool, to ensure that the output is what we want?

apple/internal/apple_support_toolchain.bzl
```diff
@@ -152,6 +156,13 @@ A `File` referencing a dynamic library used to redirect stdout and stderr when n
             executable = True,
             doc = """
 A `File` referencing a tool that copies and lipos Swift stdlibs required for the target to run.
+""",
+        ),
+        "symbols_tool": attr.label(
+            cfg = "host",
```
thii, line 162: This should be `exec`, though all other darwin tools here use the `host` configuration, so I just keep it the same.

apple/internal/partials/framework_import.bzl
```diff
@@ -284,6 +286,10 @@ def _generate_symbols(
             if not sets.contains(has_dsym_framework_basenames, framework_basename):
                 all_binaries.append(framework_binary)
 
+    # Skip the action if there're no binaries that can provide symbols
+    if not all_binaries:
+        return []
```
thii, line 291: This fixes a bug - previously the action would still be executed (with just `mkdir -p /path/to/symbols_files`) even if the binary list was empty.

tools/symbols_tool/symbols_tool.py
thii: Added a test.

tools/symbols_tool/symbols_tool.py
```diff
@@ -0,0 +1,100 @@
...
+  symbols_cmds = []
+  for i, arch in enumerate(archs):
+    symbols_cmds += ["xcrun symbols -noTextInSOD -noDaemon -arch {arch} \
+        -symbolsPackageDir {output_dir} {binary}".format(
+            arch=arch,
+            output_dir=output_dir,
+            binary=binaries[i],
+        )]
+  cmd.append("\n".join(symbols_cmds))
+  execute.execute_and_filter_output(
```
segiddins, line 47: instead of executing the list of commands via bash, I think it'd be preferable to execute each of them separately? If not, we'll need to add a `set -e` and potentially handle quoting in the `xcrun` invocation to ensure that (a) failure of any command will exit non-0 and (b) spaces wont break things

tools/symbols_tool/symbols_tool.py
```diff
@@ -0,0 +1,100 @@
...
+from build_bazel_rules_apple.tools.wrapper_common import execute
+
+_TIMING_INFO_PATTERN = r"(.*\s\[.*,\s.*\sseconds\]:)"
+
+
+def _generate_symbols(args):
+  """Generate argument parser for tool."""
+  output_dir = args.output_dir
+  archs = args.arch
+  binaries = args.binary
```
segiddins, line 32: might be nice to assert that `archs` and `binaries` are the same length?

thii:
`//test:apple_intent_library_test` seems flaky (or maybe it just takes too long sometimes that makes it aborted by CI). /cc @steeve 

https://storage.googleapis.com/bazel-untrusted-buildkite-artifacts/db7ee4f9-b5c2-41c8-9499-d756a37bf3cd/test/apple_intent_library_test/attempt_1.log

steeve:
Thanks. I see that the timeout is 15s. I think it should be less than that though.
I'll try adding a `size = medium` to it.

apple/internal/partials/framework_import.bzl
```diff
@@ -292,31 +298,28 @@ def _generate_symbols(
     )
     outputs = [symbols_dir]
 
-    commands = ["mkdir -p \"${OUTPUT_DIR}\""]
+    args = actions.args()
+    args.add("--output_dir", symbols_dir.path)
```
segiddins, line 302:
```suggestion
    args.add("--output_dir", symbols_dir)
```

we can let the `args` expand the file path at execution time, instead of doing so during analysis

thii: Does that work with directories?

thii: > Error in add: Cannot add directories to Args#add since they may expand to multiple values. Either use Args#add_all (if you want expansion) or args.add(directory.path) (if you do not).

thii: seems to work with a little trick.

segiddins (approved):

thii:
Closing since we're taking the upstream implementation for this feature.

PR #1055, opened 2021-01-28
Allow other types of verifier scripts

keith:


keith:
not sure we actually want to do this, also not sure if any existing scripts here rely on being sourced for other functions, i bet they do

PR #1056, opened 2021-01-29
Test failing test

keith:


keith:
pretty sure this won't fail but it should

keith:
something weird about bash with trap and set -u

PR #1087, opened 2021-02-22
Add support for inferring the codesigning identity during signing.

thii:
This change add support for inferring the codesigning identity during signing as well as passing it explicitly during manifest creation. Use of inferred signing identity requires specification of a provisioning profile.

Additionally correct small typo in codesigning_support.bzl for embedded manifests, and a bug in execute.py wherein subsequent commands requesting output be printed would fail after first command executed due to re-mapping stdout and stderr multiple times.

PiperOrigin-RevId: 356754914
(cherry picked from commit 366d71c93d4b857d7d0efbbb545bec7bc5886b73)

### DIFF
--- a/apple/internal/codesigning_support.bzl
+++ b/apple/internal/codesigning_support.bzl
@@ -449,7 +449,7 @@ def _generate_codesigning_dossier_action(
       provisioning_profile: The provisioning profile file. May be `None`.
       resolved_codesigning_dossier_tool: The `struct` from resolve_tools representing the code signing tool.
     """
-    input_files = [x.dossier_path for x in embedded_dossiers]
+    input_files = [x.dossier_file for x in embedded_dossiers]
 
     mnemonic = "GenerateCodesigningDossier"
     progress_message = "Generating codesigning dossier for %s" % label_name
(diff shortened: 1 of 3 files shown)


brentleyjones (approved):

steeve:
Thanks @thii, what does this bring vs that codesigning tool already does? AFAIK its role is to automatically infer the codesigning identity too.

brentleyjones:
This is just bringing in google changes. I haven't determined what the use of the dossier tool is yet.

apple/internal/codesigning_support.bzl
```diff
@@ -422,6 +426,91 @@ def _codesigning_command(
...
+        dossier_arguments.extend(["--entitlements_file", entitlements.path])
+    if provisioning_profile:
+        input_files.append(provisioning_profile)
+        dossier_arguments.extend(["--provisioning_profile", provisioning_profile.path])
+
+    for embedded_dossier in embedded_dossiers:
+        input_files.append(embedded_dossier.dossier_file)
+        dossier_arguments.extend(["--embedded_dossier", embedded_dossier.relative_bundle_path, embedded_dossier.dossier_file.path])
+
+    args_file = intermediates.file(
```
segiddins, line 477: this feels like it should be doable with an `actions.args()`

thii: This would be a nice improvement, but since the open-source version of the rules is not using this code path, I'm leaving it as-is to help keep things in sync.

PR #1092, opened 2021-02-27
Fix module name in example/multi_platform/Buttons

S-Shimotori:
I found this target got error because its `module_name` was wrong.

```sh
bazel build //examples/multi_platform/Buttons:ButtonsStaticFramework
```

```
Error in fail:
error: Found swift_library with module name 'Buttons' but expected 'ButtonsStaticFramework'. Swift static frameworks expect a single swift_library dependency with `module_name` set to the same `bundle_name` as the static framework target.
```

This PR will

- Fix `bundle_name` attribute (Use the same name as `ButtonsLib`)
- Add test case to `ExamplesBuildTest`

### DIFF
--- a/examples/multi_platform/Buttons/BUILD
+++ b/examples/multi_platform/Buttons/BUILD
@@ -111,7 +111,7 @@ ios_extension(
 
 ios_static_framework(
     name = "ButtonsStaticFramework",
-    bundle_name = "ButtonsStaticFramework",
+    bundle_name = "Buttons",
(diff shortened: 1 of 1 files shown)


thii (approved):
LGTM. Can you fix the `ButtonsTVStaticFramework` too while you're at it?

S-Shimotori:
sure

thii:
@S-Shimotori Thanks for the fix!

PR #1093, opened 2021-03-03
Add objc_provider to dynamic framework rules

mccorkill1:
In its current state, the dynamic framework rules are only packaging rules. They take one swift_library as a dependency and bundle it into a dynamic framework that xcode can consume. Because swift_library outputs a static library, if it has any swift_library dependencies, they are linked into the library that gets bundled into the dynamic framework. This can cause a runtime error if multiple dynamic frameworks imported in Xcode have the same swift_library linked.

To avoid the runtime error, I am adding the objc_provider to the dynamic framework rules. This allows them to be used as a dependency of a swift_library and the framework will be linked dynamically instead of statically. So, in the above example, instead of packaging a swift_library that depends on a swift_library, you would package a swift_library that depends on a dynamic framework. Then, in Xcode, when you import two dynamic frameworks that depend on the same swift_library, you won’t get two copies at runtime since they are dynamically linked instead of statically linked.

### DIFF
--- a/apple/internal/ios_rules.bzl
+++ b/apple/internal/ios_rules.bzl
@@ -1120,7 +1120,9 @@ def _ios_dynamic_framework_impl(ctx):
             # Make the ObjC provider using the framework_files depset found
             # in the AppleDynamicFramework provider. This is to make the
             # ios_dynamic_framework usable as a dependency in swift_library
-            objc_provider = apple_common.new_objc_provider(dynamic_framework_file = provider.framework_files)
+            objc_provider = apple_common.new_objc_provider(
+                dynamic_framework_file = provider.framework_files
+            )
             new_providers.append(objc_provider)
 
     return [
--- a/apple/internal/ios_rules.bzl
+++ b/apple/internal/ios_rules.bzl
@@ -1120,7 +1120,7 @@ def _ios_dynamic_framework_impl(ctx):
             # Make the ObjC provider using the framework_files depset found
             # in the AppleDynamicFramework provider. This is to make the
             # ios_dynamic_framework usable as a dependency in swift_library
-            objc_provider = apple_common.new_objc_provider(**{"dynamic_framework_file": provider.framework_files})
+            objc_provider = apple_common.new_objc_provider(dynamic_framework_file = provider.framework_files)
             new_providers.append(objc_provider)
 
     return [
--- a/apple/internal/ios_rules.bzl
+++ b/apple/internal/ios_rules.bzl
@@ -1113,6 +1113,17 @@ def _ios_dynamic_framework_impl(ctx):
 
     providers = processor_result.providers
 
+    new_providers = []
+    for provider in providers:
+        new_providers.append(provider)
+        if type(provider) == "AppleDynamicFramework":
+            # Make the ObjC provider using the framework_files depset found in the AppleDynamicFramework provider
+            # This is to make the ios_dynamic_framework usable as a dependency in swift_library
+            objc_provider_fields = {}
+            objc_provider_fields["dynamic_framework_file"] = provider.framework_files
+            objc_provider = apple_common.new_objc_provider(**objc_provider_fields)
+            new_providers.append(objc_provider)
+
     return [
         DefaultInfo(files = processor_result.output_files),
         IosFrameworkBundleInfo(),
@@ -1122,7 +1133,7 @@ def _ios_dynamic_framework_impl(ctx):
                 processor_result.output_groups,
             )
         ),
-    ] + providers
+    ] + new_providers
 
 def _ios_static_framework_impl(ctx):
     """Experimental implementation of ios_static_framework."""
--- a/apple/internal/ios_rules.bzl
+++ b/apple/internal/ios_rules.bzl
@@ -1120,9 +1120,7 @@ def _ios_dynamic_framework_impl(ctx):
             # Make the ObjC provider using the framework_files depset found
             # in the AppleDynamicFramework provider. This is to make the
             # ios_dynamic_framework usable as a dependency in swift_library
-            objc_provider_fields = {}
-            objc_provider_fields["dynamic_framework_file"] = provider.framework_files
-            objc_provider = apple_common.new_objc_provider(**objc_provider_fields)
+            objc_provider = apple_common.new_objc_provider(**{"dynamic_framework_file": provider.framework_files})
             new_providers.append(objc_provider)
 
     return [
--- a/apple/internal/ios_rules.bzl
+++ b/apple/internal/ios_rules.bzl
@@ -1113,9 +1113,8 @@ def _ios_dynamic_framework_impl(ctx):
 
     providers = processor_result.providers
 
-    new_providers = []
+    additional_providers = []
     for provider in providers:
-        new_providers.append(provider)
         if type(provider) == "AppleDynamicFramework":
             # Make the ObjC provider using the framework_files depset found
             # in the AppleDynamicFramework provider. This is to make the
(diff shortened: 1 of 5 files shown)


thii:
Since `*_dynamic_framework` is a packaging rule (for third-party distribution), it doesn't make sense when an `*_dynamic_framework` target is a dependency of a `swift_library` target.

Instead of allowing `*_dynamic_framework`s to be a dependency of another `*_dynamic_framework`, what about adding an `avoid_deps` attribute to allow passing a list of `{objc,swift}_library`s on which this framework depends in order to compile, but the transitive closure of which will not be statically linked into the framework's binary?

brentleyjones:
I've always wondered, _why_ do we have "packaging" rules. Why can't they both be for packaging and deps?

mccorkill1:
I just added a few tests to demonstrate how the ios_dynamic_framework rule can be used as a dependency for a swift_library. I think I agree with @brentleyjones. The dynamic framework rules are still packaging rules, but by adding the objc_provider they can _also_ be used as deps. If that is not the preferred way forward I can look into the avoid_deps approach, but I thought this was a good way to add a useful piece of functionality with just a few lines of code.

thii (approved):
I don't know the original story either, but from what I heard, Bazel tends to separate packaging from building. If we could do the same thing with the `*_library` rules, I think we'd better use those.

Another minor concern is that, if the rule consumers haven't seen this, it might be confusing at first if they find that their dependency graph has to be structured in a specific way in order to "exclude" some deps from the final binary.

Since this can solve the above use-case with very little change, I think it's fine to go with this direction now, and defer the avoid_deps implementation later.

apple/internal/ios_rules.bzl
```diff
@@ -1113,6 +1113,18 @@ def _ios_dynamic_framework_impl(ctx):
...
 
+    new_providers = []
+    for provider in providers:
+        new_providers.append(provider)
+        if type(provider) == "AppleDynamicFramework":
+            # Make the ObjC provider using the framework_files depset found
+            # in the AppleDynamicFramework provider. This is to make the
+            # ios_dynamic_framework usable as a dependency in swift_library
+            objc_provider_fields = {}
+            objc_provider_fields["dynamic_framework_file"] = provider.framework_files
```
brentleyjones, line 1124: Instead of creating an empty dict, then filling one field, can you create the dict with that field?

apple/internal/ios_rules.bzl
```diff
@@ -1113,6 +1113,18 @@ def _ios_dynamic_framework_impl(ctx):
...
+            objc_provider = apple_common.new_objc_provider(**objc_provider_fields)
```
brentleyjones, line 1125: Actually, since you are only passing one field, always, can you instead call `new_objc_provider` with the one field?

apple/internal/ios_rules.bzl
```diff
@@ -1113,6 +1113,18 @@ def _ios_dynamic_framework_impl(ctx):
 
     providers = processor_result.providers
 
+    new_providers = []
+    for provider in providers:
+        new_providers.append(provider)
+        if type(provider) == "AppleDynamicFramework":
+            # Make the ObjC provider using the framework_files depset found
+            # in the AppleDynamicFramework provider. This is to make the
+            # ios_dynamic_framework usable as a dependency in swift_library
```
brentleyjones, line 1122: Does this allow it to be a usable dependency in objc_library as well?

apple/internal/tvos_rules.bzl
```diff
@@ -471,6 +471,18 @@ def _tvos_dynamic_framework_impl(ctx):
...
+    providers = processor_result.providers
+    new_providers = []
+    for provider in providers:
+        new_providers.append(provider)
+        if type(provider) == "AppleDynamicFramework":
+            # Make the ObjC provider using the framework_files depset found in the AppleDynamicFramework provider
+            # This is to make the tvos_dynamic_framework usable as a dependency in swift_library
+            objc_provider_fields = {}
+            objc_provider_fields["dynamic_framework_file"] = provider.framework_files
+            objc_provider = apple_common.new_objc_provider(**objc_provider_fields)
```
brentleyjones, line 483: Same comment about calling this directly.

apple/internal/watchos_rules.bzl
```diff
@@ -273,6 +273,17 @@ def _watchos_dynamic_framework_impl(ctx):
...
 
+    new_providers = []
+    for provider in providers:
+        new_providers.append(provider)
+        if type(provider) == "AppleDynamicFramework":
+            # Make the ObjC provider using the framework_files depset found in the AppleDynamicFramework provider
+            # This is to make the watchos_dynamic_framework usable as a dependency in swift_library
+            objc_provider_fields = {}
+            objc_provider_fields["dynamic_framework_file"] = provider.framework_files
+            objc_provider = apple_common.new_objc_provider(**objc_provider_fields)
```
brentleyjones, line 284: Same comment about calling this directly.

apple/internal/ios_rules.bzl
```diff
@@ -1113,6 +1113,16 @@ def _ios_dynamic_framework_impl(ctx):
...
+            objc_provider = apple_common.new_objc_provider(**{"dynamic_framework_file": provider.framework_files})
```
brentleyjones, line 1123: I think you can just call `apple_common.new_objc_provider(dynamic_framework_file
 = provider.framework_files)`?

apple/internal/ios_rules.bzl
```diff
@@ -1113,6 +1113,16 @@ def _ios_dynamic_framework_impl(ctx):
...
     providers = processor_result.providers
 
+    new_providers = []
+    for provider in providers:
+        new_providers.append(provider)
+        if type(provider) == "AppleDynamicFramework":
+            # Make the ObjC provider using the framework_files depset found
+            # in the AppleDynamicFramework provider. This is to make the
+            # ios_dynamic_framework usable as a dependency in swift_library
+            objc_provider = apple_common.new_objc_provider(dynamic_framework_file = provider.framework_files)
```
brentleyjones, line 1123: Now can you wrap it to 80? Thanks!
```suggestion
            objc_provider = apple_common.new_objc_provider(
                dynamic_framework_file = provider.framework_files
            )
```

apple/internal/ios_rules.bzl
```diff
@@ -1113,6 +1113,18 @@ def _ios_dynamic_framework_impl(ctx):
...
+        new_providers.append(provider)
+        if type(provider) == "AppleDynamicFramework":
+            # Make the ObjC provider using the framework_files depset found
+            # in the AppleDynamicFramework provider. This is to make the
+            # ios_dynamic_framework usable as a dependency in swift_library
+            objc_provider = apple_common.new_objc_provider(
+                dynamic_framework_file = provider.framework_files
+            )
+            new_providers.append(objc_provider)
+
```
brentleyjones, line 1127: Can build up `additional_providers`, add them to `providers`, then use `providers` down on 1137 instead of `new_providers`:

```suggestion
    additional_providers = []
    for provider in providers:
        if type(provider) == "AppleDynamicFramework":
            # Make the ObjC provider using the framework_files depset found
            # in the AppleDynamicFramework provider. This is to make the
            # ios_dynamic_framework usable as a dependency in swift_library
            objc_provider = apple_common.new_objc_provider(
                dynamic_framework_file = provider.framework_files
            )
            additional_providers.append(objc_provider)
    providers.extend(additional_providers)
```

Let's do this in the three locations.

brentleyjones (commented):
I would like to get the `avoid_deps` alternative in as well (this can be a followup PR). Also, ideally this should work with more than just `swift_library`.

apple/internal/ios_rules.bzl
mccorkill1: Yep! I like that better. I'll make the update today and ping you when it's done.

brentleyjones (approved):

brentleyjones:
@mccorkill1 I'll merge this now, to get it in and unblock you, but can you look into the `objc_library` use as well? Ideally this wouldn't only work for swift_library.

mccorkill1:
I'll add it to the backlog. I did a bit of research on this a while ago but since it wasn't our primary use case it got put aside. I agree that it should work for both types of library.

PR #1099, opened 2021-03-11
Introduce minimum_deployment_os_version

steeve:
This a draft, I'll update as we move along.

`minimum_deployment_os_version` allows to decorelate the bundle `MinimumOSVersion`
from the `minimum_os_version` which also controls the compilation flags.

This allows to build most modules only once (provided new features are
guarded, also with weak frameworks). However, when trying to
install such bundle on iOS < 14, it will fail with:

    This app contains an app extension that specifies an extension point
    identifier that is not supported on this version of iOS for the value of
    the NSExtensionPointIdentifier key in its Info.plist.

In our case, when using WidgetKit, this was because the `NSExtensionPointIdentifier=com.apple.widgetkit-extension` wasn't not recognized.

It is possible, however, to still compile with one version, but declare a
different `MinimumOSVersion` in the bundle's plist.

This is what this PR does.

Not only this reduces the number of actions, it also fixes code completion breaking from different iOS minimum versions in Tulsi, for which there is no fix AFAICT.


steeve:
I am unable to update the doc for the same reason: https://github.com/bazelbuild/rules_apple/pull/1182#issuecomment-857762996

brentleyjones:
Can you add some tests for this?

steeve:
Ah right, I figure I'd wait for some feedback first.

brentleyjones (approved):

steeve:
I'm not sure what to do with this error which seems unrelated, and I can't reproduce:
```
(22:36:43) ERROR: /Users/buildkite/builds/bk-imacpro-15/bazel/rules-apple-darwin/test/starlark_tests/targets_under_test/ios/BUILD:457:16: in swift_runtime_linkopts rule //test/starlark_tests/targets_under_test/ios:app_with_swift_dep.swift_runtime_linkopts:
--
&nbsp; | Traceback (most recent call last):
&nbsp; | File "/Users/buildkite/builds/bk-imacpro-15/bazel/rules-apple-darwin/apple/internal/swift_support.bzl", line 77, column 60, in _swift_runtime_linkopts_impl
&nbsp; | linkopts.extend(swift_common.swift_runtime_linkopts(
&nbsp; | File "/private/var/tmp/_bazel_buildkite/e12247f29a0bb4c84aab3cca347d1d97/external/build_bazel_rules_swift/swift/internal/linking.bzl", line 312, column 18, in swift_runtime_linkopts
&nbsp; | toolchain.linker_opts_producer,
&nbsp; | Error: 'SwiftToolchainInfo' value has no field or method 'linker_opts_producer'
&nbsp; | Available attributes: action_configs, all_files, cc_toolchain_info, clang_implicit_deps_providers, cpu, feature_allowlists, generated_header_module_implicit_deps_providers, implicit_deps_providers, linker_supports_filelist, object_format, requested_features, supports_objc_interop, swift_worker, system_name, test_configuration, tool_configs, unsupported_features
```

steeve:
Ah I think this is because of https://github.com/bazelbuild/rules_swift/commit/446d8f4a1531357e089795ef597af88d87e9ba9d

segiddins (approved):

brentleyjones:
Thanks @steeve!

steeve:
Thank you for being so fast !

jerrymarino:
Nice PR! I was wondering if you had thought about respecting this value inside of the info.plist: currently `rules_apple` seems to clobber it. On the bright side it'd prevent us from having the extra flag!

PR #1115, opened 2021-04-02
Fix incorrect platform type

dbettermann:
Without this change I was getting device entitlements applied to simulator builds.

### DIFF
--- a/apple/internal/entitlement_rules.bzl
+++ b/apple/internal/entitlement_rules.bzl
@@ -282,7 +282,7 @@ def _entitlements_impl(ctx):
         explicit_minimum_os = None,
         features = ctx.features,
         objc_fragment = ctx.fragments.objc,
-        platform_type_string = str(ctx.fragments.apple.single_arch_platform.platform_type),
+        platform_type_string = ctx.attr.platform_type,
         uses_swift = uses_swift,
         xcode_path_wrapper = None,
         xcode_version_config = ctx.attr._xcode_config[apple_common.XcodeVersionConfig],


apple/internal/entitlement_rules.bzl
```diff
@@ -282,7 +282,7 @@ def _entitlements_impl(ctx):
         explicit_minimum_os = None,
         features = ctx.features,
         objc_fragment = ctx.fragments.objc,
-        platform_type_string = str(ctx.fragments.apple.single_arch_platform.platform_type),
+        platform_type_string = ctx.attr.platform_type,
```
brentleyjones, line 285: Offline we talked, but I think this bug might have been hidden by the transition, which you have disabled.

dbettermann: Successfully reverted the transition patch.

Getting an error about a missing attribute when trying to use it within the app. Mentioning because it appears to be an issue with the `ios_dylib` rule that you have here: https://github.com/bazelbuild/rules_apple/pull/987

`no such attribute 'families' in 'ios_dylib' rule`

Will post again after successfully testing the transition revert.

dbettermann: Nvm on the `families` attribute.

Was missing this: https://github.com/bazelbuild/rules_apple/pull/987/files#diff-4a7ae7309bfed96760d671ad00e5726a6d0c466112275402d9944abd295d7328R1109-R1115

dbettermann: Not having signing issues on the sim without this change + reverting the transition. Will close today unless I hear that this is need for some reason.

brentleyjones: I think it might still be worth fixing, just that now we know the transition is probably covering up the bug.

PR #1132, opened 2021-04-19
Use arm64e when compiling arm64 macos_kernel_extensions.

keith:
RELNOTES: Apple Silicon kernel extensions must be compiled as arm64e even if other userspace code is compiled for arm64.
PiperOrigin-RevId: 367280391
(cherry picked from commit 162721c2ba8cb7f9ee4019f194b79e3444d435c8)

### DIFF
--- a/apple/internal/BUILD
+++ b/apple/internal/BUILD
@@ -287,6 +287,7 @@ bzl_library(
         ":rule_factory",
         ":rule_support",
         ":run_support",
+        ":transition_support",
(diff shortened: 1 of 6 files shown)


keith:
requires bazel 4.1.0 to pass AFAICT locally, probably some of the Apple Silicon support in bazel itself

keith:
going to merge with tests disabled because the transition changes reduce conflicts with other upstream fixes we need

Issue #867, opened 2020-08-11
Add docs generation


keith:
https://github.com/bazelbuild/stardoc

alexeagle:
Need some help with this one too? I recently added stardoc for rules_docker

keith:
that would be awesome!

PR #1135, opened 2021-04-29
Introduce first stardoc API docgen

alexeagle:
This replaces some hand-maintained documentation with the output of the documentation generator.
I merged the contents by porting some of the hand-edits back to the API docstrings.

This gives us a pattern to follow to incrementally replace the docs with generated equivalents.

Note that the documentation has to be organized into .md files that are 1:1 with .bzl files we generate from (a starlark 'module')

See #867


keith (approved):

keith:
Thanks!

PR #1136, opened 2021-04-30
docs: docgen for dtrace.bzl

alexeagle:


keith (approved):

keith:
something incompat landed upstream, I will look to fix it monday

PR #1138, opened 2021-05-01
Stardoc rules to generate API docs for watchos

alexeagle:
Changes the migration pattern to match existing filenames. This means we don't need so much link redirection, and causes less churn on the git history.


apple/watchos.doc.bzl
```diff
@@ -0,0 +1,22 @@
+"""Bazel rules for creating watchOS applications and bundles."""
```
thii, line 1: Can this be an H1 header?

thii (approved):

apple/watchos.doc.bzl
alexeagle: Sure, pushed another commit changing all the module docstrings to start with H1 to keep the generated pages consistent.

https://github.com/bazelbuild/rules_apple/blob/b6e4595f4eb7be613e3a267d395d808f85009b40/doc/rules-watchos.md
is the preview

PR #1141, opened 2021-05-04
Use cpp fragment, instead of objc fragment, to determine whether to generate dsym

keith:
The information is identical, and we would like to migrate all uses to
cpp fragment so that we can delete the info in objc fragment.

PiperOrigin-RevId: 370450547
(cherry picked from commit 3a981f465334985b7a1d29ec805a1d1299081707)


keith:
requires rules_swift release

PR #1146, opened 2021-05-04
Revert "Add support for old dSYMs API for now"

keith:
This reverts commit 4834b24285fcab6284c35ec872a2ce90f6252811.


thii:
There's one more. Forgot to add a TODO in the code, so just note it here so we won't forget:

https://github.com/bazelbuild/rules_apple/blob/029baeadd43ca88e1bb0dfcb3ebd31825c2afda3/apple/internal/partials/apple_symbols_file.bzl#L57

keith:
thanks pushed that here

PR #1149, opened 2021-05-06
Remove link_cocoa feature

keith:
All this did was pass `-framework Cocoa` which should be handled
automatically as needed.


segiddins (approved):

keith:
I'm surprised this actually breaks stuff, I guess it's just a nicety given that every app needs Cocoa. We can leave it for now

PR #1151, opened 2021-05-08
Ignore duplicate class warnings when generating plists

keith:
Otherwise on a M1 machine you see this:

```
objc[79978]: Class AMSupportURLConnectionDelegate is implemented in both /usr/lib/libauthinstall.dylib (0x207db1160) and /System/Library/PrivateFrameworks/MobileDevice.framework/Versions/A/MobileDevice (0x116ea42b8). One of the two will be used. Which one is undefined.
objc[79978]: Class AMSupportURLSession is implemented in both /usr/lib/libauthinstall.dylib (0x207db11b0) and /System/Library/PrivateFrameworks/MobileDevice.framework/Versions/A/MobileDevice (0x116ea4308). One of the two will be used. Which one is undefined.
```


segiddins (approved):

keith:
done in https://github.com/bazelbuild/rules_apple/commit/b8b3cc52d1c4c395a33aec2b4eb6015ebd6a6e94

PR #1153, opened 2021-05-10
[Apple/Native Symbols] New option to include .symbols files in .ipa

thii:
This reverts 4287aa277e14c693a9c036a8382521964c6951b3 (our previous implementation of this feature) - and cherry-picks dcbe0383634f510c3fa2af021a99d34d577d0c94 (the upstream change).

Also refactor https://github.com/bazelbuild/rules_apple/pull/910 to add support for processing user-provided dSYMs via `apple_symbols_file_partial`.

---

Apple uses the `.symbols` file format for uploading native debug symbols inside an `.ipa`:

http://faq.sealedabstract.com/xcodeCI/#create-symbol-files-with-this-one-weird-trick

Similar to Bitcode symbol uploading, this allows Apple's infrastructure to decode
iOS apps' native symbols, which allows them to report actionable performance
issues and crashes.

This CL follows the example from the `bitcode_symbols.bzl` partial and does the
same thing for `.symbols` files, which are generated using the `symbols`
command-line tool inside Xcode.

This feature is opt-in at the `ios_application()` level. Set `include_symbols_in_bundle = True` to enable it.


doc/common_info.md
```diff
@@ -173,14 +173,6 @@ can pass `--define=apple.propagate_embedded_extra_outputs=(yes|true|1)` to
...
 ```
 
-### Include symbols in ipas
-
-Including symbols in the final ipa is necessary if you want to allow Apple to
-symbolicate crash logs from App Store Connect and provide to you directly in
-the Xcode's Organizer window. If you want to include symbols in your ipa for
-App Store builds, you can pass `--apple_generate_dsym` and
-`--define=apple.package_symbols=yes` to `bazel build`.
-
```
brentleyjones, line 183: Should we have some form of docs still?

brentleyjones (approved):

doc/common_info.md
thii: I’ll add docs back. This PR isn’t finished yet. The upstream change doesn’t handle the case with prebuilt dSYMs (https://github.com/bazelbuild/rules_apple/pull/910), so we’ll have to reimplement that.

thii: This is now an attribute, so the docs will be generated later by Stardoc.

brentleyjones (approved):

keith (approved):

PR #1154, opened 2021-05-10
Switch default CPU to use host

keith:
When building on an Apple Silicon machine, these CPUs would still have
been for x86_64. This flips the default to match the host machine instead.
Users can still pass the platform specific cpu flags to override them.

Closes https://github.com/bazelbuild/bazel/issues/14284.

### DIFF
--- a/apple/internal/transition_support.bzl
+++ b/apple/internal/transition_support.bzl
@@ -47,7 +47,7 @@ def _cpu_string(*, cpu, platform_type, settings):
         macos_cpus = settings["//command_line_option:macos_cpus"]
         if macos_cpus:
             return "darwin_{}".format(macos_cpus[0])
-        return "darwin_x86_64"
+        return settings["//command_line_option:host_cpu"]
     if platform_type == "tvos":
         if cpu:
             return "tvos_{}".format(cpu)
@@ -190,6 +190,7 @@ _apple_rule_common_transition_inputs = [
 ]
 _apple_rule_base_transition_inputs = _apple_rule_common_transition_inputs + [
     "//command_line_option:cpu",
+    "//command_line_option:host_cpu",
     "//command_line_option:ios_multi_cpus",
     "//command_line_option:macos_cpus",
     "//command_line_option:tvos_cpus",
--- a/apple/internal/transition_support.bzl
+++ b/apple/internal/transition_support.bzl
@@ -40,6 +40,8 @@ def _cpu_string(*, cpu, platform_type, settings):
         cpu_value = settings["//command_line_option:cpu"]
         if cpu_value.startswith("ios_"):
             return cpu_value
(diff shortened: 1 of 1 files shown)


keith:
so there's still weird behavior here, and we need this alongside https://github.com/bazelbuild/bazel/pull/13440

With just this change, the path to the output has the right arch in it, but the binary has the wrong arch:

```
% file bazel-out/applebin_macos-darwin_arm64-fastbuild-ST-fb66f80159e9/bin/examples/macos/CommandLineSwift/CommandLineSwift
bazel-out/applebin_macos-darwin_arm64-fastbuild-ST-fb66f80159e9/bin/examples/macos/CommandLineSwift/CommandLineSwift: Mach-O 64-bit executable x86_64
```

and without this change, but with that bazel default change you get the right arch but the wrong path:

```
% file bazel-out/applebin_macos-darwin_x86_64-fastbuild-ST-3790a2650836/bin/examples/macos/CommandLineSwift/CommandLineSwift
bazel-out/applebin_macos-darwin_x86_64-fastbuild-ST-3790a2650836/bin/examples/macos/CommandLineSwift/CommandLineSwift: Mach-O 64-bit executable arm64
```

segiddins (approved):

keith:
I no longer see the issue mentioned above with 5.0.0rc2

brentleyjones (approved):

apple/internal/transition_support.bzl
```diff
@@ -40,13 +40,18 @@ def _cpu_string(*, cpu, platform_type, settings):
...
+            return "ios_sim_arm64"
         return "ios_x86_64"
     if platform_type == "macos":
         if cpu:
             return "darwin_{}".format(cpu)
         macos_cpus = settings["//command_line_option:macos_cpus"]
         if macos_cpus:
             return "darwin_{}".format(macos_cpus[0])
+        cpu_value = settings["//command_line_option:cpu"]
+        if cpu_value.startswith("darwin_"):
```
thii, line 53: On Intel Macs, unless `--cpu=darwin_x86_64` is passed, this defaults to `darwin`.

```suggestion
        if cpu_value.startswith("darwin"):
```

brentleyjones: But would we want to return `darwin`, or fall through to the `darwin_x86_64` case? I think we want to preserve old behavior for Intel.

keith: hrm yea that's kinda weird either way. I guess in that case we can continue letting it fallthrough as it was before, but we should try to eliminate that ambiguous arch identifier anyways

brentleyjones:
I tested this an it works. I think we should merge it in order to fix https://github.com/bazelbuild/bazel/issues/14284.

PR #1155, opened 2021-05-10
Add override_archs to macos_command_line_application

keith:
This allows you to specify that the rule be built for multiple
architectures and produce a fat binary. This is useful to force binaries
to be share-able between Apple Silicon and Intel machines which is
useful for making sure host tools don't invalidate caches.

### DIFF
--- a/apple/internal/macos_rules.bzl
+++ b/apple/internal/macos_rules.bzl
@@ -1416,6 +1416,17 @@ macos_command_line_application = rule_factory.create_apple_binary_rule(
     platform_type = "macos",
     product_type = apple_product_type.tool,
     doc = "Builds a macOS Command Line Application binary.",
+    additional_attrs = {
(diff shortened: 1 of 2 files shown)


keith:
Submitting as an example alternative to https://github.com/bazelbuild/bazel/pull/13461

brentleyjones:
I can see a need for both this and https://github.com/bazelbuild/bazel/pull/13461. This can be used to undo a global setting. I could see the global setting being a very normal thing (enough that I think the default should be both x86 and arm, and you would use the flag to undo it if you didn't need it).

keith:
Looks like https://github.com/bazelbuild/rules_apple/commit/73c041db161f5058f2e4cc124b7591300618b5a6 will help this use case

keith:
There is now the universal rule for this instead

PR #1161, opened 2021-05-18
Pass command line arguments through launch options to xctestrunner

ra1028:
This allows CommandLineArguments to be passed via xcestrunner.
For example, `-AppleLanguages (en)` option to change the language settings of the test execution environment can be passed now.

The options passed are passed through [here](https://github.com/google/xctestrunner/blob/64a9be0b6fa833b4b2371729c5c8cdd2c6f7775b/test_runner/xctestrun.py#L78-L87) to the `CommandLineArguments`, which are consumed by the simulator.
The flags passed in `test_arg` are passed directly to xctestrunner by default, so I defined `--command_line_args` to make them distinguishable.

I've added unit tests, but the tests `//test:ios_test_runner_unit_test` itself don't seem to be passed for several reasons at the moment. Once this PR is merged, I will fix it so that the tests will pass.

### DIFF
--- a/test/ios_test_runner_unit_test.sh
+++ b/test/ios_test_runner_unit_test.sh
@@ -433,6 +433,25 @@ function test_ios_unit_simulator_id() {
   expect_log "Executed 3 tests, with 0 failures"
 }
 
+function test_ios_unit_test_dot_separated_command_line_args() {
+  create_sim_runners
+  create_ios_unit_tests
+  do_ios_test //ios:PassingUnitTest \
+    --test_arg="--command_line_args=arg1,arg2,arg3" || fail "should pass"
+
+  expect_log "Test Suite 'PassingUnitTest' passed"
+}
+
+function test_ios_unit_test_multiple_command_line_args() {
+  create_sim_runners
+  create_ios_unit_tests
+  do_ios_test //ios:PassingUnitTest \
+    --test_arg="--command_line_args=arg1" \
+    --test_arg="--command_line_args=arg2" || fail "should pass"
+
+  expect_log "Test Suite 'PassingUnitTest' passed"
+}
+
 function test_ios_unit_other_arg() {
   create_sim_runners
   create_ios_unit_tests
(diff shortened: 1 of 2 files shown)


ra1028:
@googlebot I signed it

test/ios_test_runner_unit_test.sh
```diff
@@ -433,6 +433,25 @@ function test_ios_unit_simulator_id() {
...
+}
+
+function test_ios_unit_test_multiple_command_line_args() {
+  create_sim_runners
+  create_ios_unit_tests
+  do_ios_test //ios:PassingUnitTest \
+    --test_arg="--command_line_args=arg1" \
+    --test_arg="--command_line_args=arg2" || fail "should pass"
+
+  expect_log "Test Suite 'PassingUnitTest' passed"
```
keith, line 452: the other tests here actually verify these were passed, can you do that? i think you can check in NSUserDefaults to get these to see them

keith (commented):
looks good, one comment on the test, thanks!

test/ios_test_runner_unit_test.sh
ra1028: @keith 
Thanks for the comment. I've added tests to check if the arguments have been passed to the app runtime using ProcessInfo.
`bazel test //test:ios_test_runner_unit_test` still doesn't pass because other test fails. I'll work on that after this PR.

keith: everything is passing on CI it looks like, so im not sure which one you mean?

keith (approved):

test/ios_test_runner_unit_test.sh
thii: This test is tagged "manual" so it's disabled on CI.
https://github.com/bazelbuild/rules_apple/blob/d15beb478c05e493a87620af0ee67ab0ece41632/test/BUILD#L407-L420

keith: hrm we should probably enable now and toss 32 bit support

PR #1162, opened 2021-05-20
Fix `test:ios_test_runner_unit_test` fail and enable to run it on CI

ra1028:
`//test:ios_test_runner_unit_test` doesn't actually pass and it's not exposed because it's tagged `manual` to prevent running on CI.
But now, probably we can drop support for testing on 32-bit simulator and re-enable it.

Here is a PR that added `manual` tag to this test: https://github.com/bazelbuild/rules_apple/pull/769

### DIFF
--- a/test/BUILD
+++ b/test/BUILD
@@ -412,11 +412,6 @@ apple_shell_test(
         "--ios_multi_cpus=i386,x86_64",
     ],
     shard_count = 9,
-    tags = [
-        # Disable these tests on BuildKite because we can't guarantee that it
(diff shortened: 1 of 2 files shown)


test/BUILD
```diff
@@ -412,11 +412,6 @@ apple_shell_test(
         "--ios_multi_cpus=i386,x86_64",
```
ra1028, line 412: @keith 
Should we drop support for `i386` cpu here?

thii: Yes it's fine to drop it.

thii (approved):

PR #1164, opened 2021-05-21
Add additional_linker_inputs to linking attributes

segiddins:
This is to support linkopts with input files, and with $(location)
expansion.  Much of the support is done by native code that
automatically does the right thing if the additional_linker_inputs
attribute is on the rule.  This is similar to how the linkopts
attribute is supported now.

PiperOrigin-RevId: 375142378
(cherry picked from commit 48d3dcb6955c21cb77134975646281ce59fee64a)


keith (approved):
nice, makes zld integration easier

brentleyjones:
Is there anything that we added to support zld, that this replaces, and should now be removed?

keith:
not in rules_apple AFAIK but in individual configs probably

brentleyjones:
Ahh, yes, we did it through a trick to get the link opts added instead :+1:.

keith:
looks like this requires a bazel release

PR #1167, opened 2021-05-24
Generate ios-rules.md from sources

alexeagle:
Replaces hand-maintained docs with stardoc automation


alexeagle:
note, the broken CI is also present on https://buildkite.com/bazel/rules-apple-darwin/builds?branch=master

apple/ios.bzl
```diff
@@ -259,21 +259,45 @@ def ios_ui_test(name, **kwargs):
...
+
+Args:
+    runners: a list of runner targets
+    **kwargs: passed to the [ios_ui_test]
+"""
     apple_test_assembler.assemble(
         name = name,
         bundle_rule = _ios_internal_ui_test_bundle,
         test_rule = _ios_ui_test,
+        runners = runners,
```
keith, line 300: 🤔

apple/macos.doc.bzl
```diff
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-"""Bazel rules for creating macOS applications and bundles."""
+"""# Bazel rules for creating macOS applications and bundles."""
```
segiddins, line 15: intentional?

segiddins (approved):

apple/macos.doc.bzl
alexeagle: yes, that's just correcting macos to look like tvos and watchos, was a mistake in the previous PR

apple/ios.bzl
alexeagle: yeah this is needed or else stardoc drops the documentation for this attribute on the floor. Note that I defaulted it to `None` which isn't the best default, but I'm being careful not to change any semantics.

More broadly, I had started with a different approach: show the stardoc for the underlying rule `_ios_internal_ui_test_bundle` - that's most similar to what's on the docsite today. However it documents everything *except* the `runners` attribute which I think is actually pretty confusing today.

thii (approved):
Nice, iOS rules docs is what we want the most :)
Thanks.

PR #1181, opened 2021-06-01
Adds support for smartcard identities

### LINKED ISSUE #1180: Support for smartcard signing identities
If you're trying to sign builds when using CryptoTokenKit based solutions to store signing identities (e.g. [Signing Manager](https://twocanoes.com/products/mac/signing-manager/)) you're going to hit this error:
```sh
Unable to find an identity on the system matching the ones in /path/to/mpp
```

When storing identities that way `security find-identity` won't be able to find them. So it'd be great if the current [codesigningtool](https://github.com/bazelbuild/rules_apple/blob/master/tools/codesigningtool/codesigningtool.py#L134-L159) was able to also find smartcard identities, e.g. by parsing the output of:
```sh
security export-smartcard -t identities
```

**EDIT:**

After testing this with a team member (Tim Sutton) we realized that there's probably a better way to tackle this. Essentially we want to be able to:

1. Find identities by `Common Name (CN)` and return their `SHA1` fingerprints or find all available identities
2. Ignore expired certs in the process

We explored three possible ways of doing that:

1. Parsing `security export-smartcard -t identities` directly
2. Use `security export-smartcard -e /some/path` and process the generated `pem` certificates
3. Use `system_profiler SPSmartCardsDataType -xml` and parse the XML directly

From our investigation option `3.` seems to be more reliable and future proof than the others because:

1. There's no easy way to get the expiration date from the output of `security export-smartcard -t identities`, it would be in a `CFData` object contained in the output ([see example here](https://gist.github.com/thiagohmcruz/bf340ee92aadf93d81c2f1639218bd06#file-security_export_smartcard_identities_output-L5)). Also that output is not a known format so the chances of future `security` updates breaking code signing are high.
2. We've found an issue with `security export-smartcard -e /some/path`. If two certificates have the same `CN` that tool will use that as an identifier while exporting the `pem` files. The problem with that is that it will end up exporting only one `pem` (export the first one then export the second one with the same name overriding the previous file on disk).
3. So it seems that `system_profiler SPSmartCardsDataType -xml` is a good choice here. It's possible to parse a known format (`XML`), it exports all identities regarding of the value of `CN` and its [output contains the expiry date too](https://gist.github.com/thiagohmcruz/d94b747217cd4e25deeaa5f77f47561f#file-system_profiler_output-L60-L66)

thiagohmcruz:
Resolves #1180

So signed builds are possible when using CryptoTokenKit based solutions to store signing identities (e.g. [Signing Manager](https://twocanoes.com/products/mac/signing-manager/)). When storing identities that way `security find-identity` won't be able to find them. 

The proposal in this PR is to append to the current list of `SHA1` the values for the `smartcard` identities found on the system. Please see detailed explanation in the linked issue #1180 for why `system_profiler SPSmartCardsDataType -xml` is possibly a good solution here.

I'm also 100% open to changing the approach if you spot any issues with this implementation!

Note that the behaviour triggered by setting the `--ios_signing_cert_name` flag continues to exist: try to find a signing identity by name and return it's `SHA1` or return all `SHA1` found if the flag is not provided.

### DIFF
--- a/tools/codesigningtool/codesigningtool.py
+++ b/tools/codesigningtool/codesigningtool.py
@@ -13,6 +13,8 @@
 # limitations under the License.
 
 import argparse
+import base64
+import datetime
 import os
 import plistlib
 import re
@@ -120,16 +122,122 @@ def _certificate_fingerprint(identity):
   fingerprint = fingerprint.replace(":", "")
   return fingerprint
 
+def _certificate_common_name(cert):
+  _, subject, _ = execute.execute_and_filter_output([
+    "openssl",
+    "x509",
+    "-noout",
+    "-inform",
+    "DER",
+    "-subject"
+  ], inputstr=cert, raise_on_failure=True)
+  subject = subject.strip().split('/')
+  cert_cn = [f for f in subject if "CN=" in f]
+
+  if len(cert_cn) == 0:
+    return None
+
+  cert_cn = cert_cn[0]
+  cert_cn = cert_cn.replace("CN=", "")
+
+  return cert_cn
 
 def _get_identities_from_provisioning_profile(mpf):
   """Iterates through all the identities in a provisioning profile, lazily."""
   for identity in mpf["DeveloperCertificates"]:
-    if not isinstance(identity, bytes):
+    cert = _certificate_data(identity)
+    yield _certificate_fingerprint(cert)
+
+def _certificate_data(cert):
+  if not isinstance(cert, bytes):
       # Old versions of plistlib return the deprecated plistlib.Data type
       # instead of bytes.
-      identity = identity.data
-    yield _certificate_fingerprint(identity)
+      cert = cert.data
 
+  return cert
+
+def _get_smartcard_tokens(xml):
+  """Get available tokens from the output of 'system_profiler SPSmartCardsDataType -xml'"""
+  tokens = [x for x in xml if x.get("_name", None) == "AVAIL_SMARTCARDS_TOKEN"]
+
+  if len(tokens) == 0:
+    return []
+
+  tokens = tokens[0].get("_items", None)
+  tokens = [x.get("_name", None) for x in tokens]
+
+  return tokens
+
+def _get_smartcard_keychain(xml):
+  """Get keychain items from the output of 'system_profiler SPSmartCardsDataType -xml'"""
+  keychain = [x for x in xml if x.get("_name", None) == "AVAIL_SMARTCARDS_KEYCHAIN"]
+
+  if len(keychain) == 0:
+    return []
+
+  keychain = keychain[0].get("_items", None)
+  return keychain
+
+def _find_smartcard_identities(identity=None):
+  """Finds smartcard identitites on the current system."""
+  ids = []
+  _, xml, _ = execute.execute_and_filter_output([
+      "/usr/sbin/system_profiler",
+      "SPSmartCardsDataType",
+      "-xml"
+  ], raise_on_failure=True)
+  xml = plistlib.loads(str.encode(xml))
+  if len(xml) == 0:
+    return []
+  xml = xml[0].get("_items", None)
+  if not xml:
+    return []
+
+  tokens = _get_smartcard_tokens(xml)
+  keychain = _get_smartcard_keychain(xml)
+
+  # For each 'token' finds non-expired certs and:
+  #
+  # 1. Check if 'identity' was provided and if it matches a 'CN', in that case stop the loop
+  #    and return the respective fingerprint (SHA1)
+  # 2. Otherwise append fingerprints found to 'ids' to be returned at the end
+  #
+  # ps: note that if 'identity' is provided and it does not match any existing item in the
+  # smartcard keychain 'ids' will be empty, making this function's behaviour consistent with
+  # '_find_codesign_identities' where it's being called
+  for token in tokens:
+    token_data = [x for x in keychain if x.get("_name", None) == token]
+    if len(token_data) == 0:
+      continue
+    token_data = token_data[0]
+
+    for (k, data) in token_data.items():
+      if k == "_name":
+        continue
+      # Extract expiry date and ignore expired certs. The row being processed looks like this:
+      #
+      # Valid from: 2021-02-12 21:35:04 +0000 to: 2022-02-12 21:35:05 +0000, SSL trust: NO, X509 trust: YES
+      #
+      expiry_date = re.search(r"(?<=to:)(.*?)(?=,)", data, re.DOTALL).group().strip()
+      expiry_date = datetime.datetime.strptime(expiry_date, "%Y-%m-%d %H:%M:%S %z")
+      now = datetime.datetime.now(expiry_date.tzinfo)
+      if now > expiry_date:
+        continue
+
+      # This is a valid identity, decode the certificate, extract
+      # Common Name and Fingerprint and handle their values accordingly
+      # as described above
+      cert = re.search(r"(?<=-----BEGIN CERTIFICATE-----)(.*?)(?=-----END CERTIFICATE-----)", data, re.DOTALL).group().strip()
+      cert = base64.b64decode(cert)
+      cert = _certificate_data(cert)
+      common_name = _certificate_common_name(cert)
+      fingerprint = _certificate_fingerprint(cert)
+      if identity == common_name:
+        return [fingerprint]
+      if not identity:
+        ids.append(fingerprint)
+
+  return ids
 
 def _find_codesign_identities(identity=None):
   """Finds code signing identities on the current system."""
@@ -156,6 +264,10 @@ def _find_codesign_identities(identity=None):
       groups = m.groupdict()
       id = groups["hash"]
       ids.append(id)
+
+  # Finds smartcard identities if present
+  ids += _find_smartcard_identities(identity)
+
   return ids
 
 


brentleyjones:
Just an FYI, we can't really review the change until the CLA is signed

thiagohmcruz:
> Just an FYI, we can't really review the change until the CLA is signed

Yep! Dealing with this internally, I believe the setup is done I might have to re-open the PR though (not 100% sure I'll ask). Working on making this reviewable as we speak.

tools/codesigningtool/codesigningtool.py
```diff
@@ -120,16 +123,100 @@ def _certificate_fingerprint(identity):
   fingerprint = fingerprint.replace(":", "")
   return fingerprint
 
+def _certificate_common_name(cert):
+  _, subject, _ = execute.execute_and_filter_output([
+    "openssl",
```
timsutton, line 128: It might be worth pointing explicitly to `/usr/bin/openssl` here, in the off chance that someone consuming these rules happens to have a custom `openssl` in their PATH.

thiagohmcruz:
@googlebot I signed it!

tools/codesigningtool/codesigningtool.py
```diff
@@ -120,16 +123,100 @@ def _certificate_fingerprint(identity):
...
+  # containing smartcards keychain data
+  _, xml, _ = execute.execute_and_filter_output([
+      "/usr/sbin/system_profiler",
+      "SPSmartCardsDataType",
+      "-xml"
+  ], raise_on_failure=True)
+  xml = ET.fromstring(xml)
+  # Find 'AVAIL_SMARTCARDS_TOKEN' node and move up one level
+  # to find the 'array' of available tokens, each one is a 'string' node
+  available_tokens_xml = xml.findall('.//*/string[.="AVAIL_SMARTCARDS_TOKEN"]../array')[0]
```
timsutton, line 167: We may want to not assume that these `AVAIL_SMARTCARDS_TOKEN` and `AVAIL_SMARTCARDS_KEYCHAIN` (below) items exist, and default back to an empty array that could help the function return earlier

timsutton:
One thing we probably will want to establish is if this constrains macOS versions that could be supported. Do we care about supporting anything older than Mojave (10.14) or Catalina (10.15), for example? That might help inform the logic in what's being searched for in the `system_profiler` output.

tools/codesigningtool/codesigningtool.py
```diff
@@ -120,16 +123,100 @@ def _certificate_fingerprint(identity):
...
+
+def _certificate_data(cert):
+  if not isinstance(cert, bytes):
       # Old versions of plistlib return the deprecated plistlib.Data type
       # instead of bytes.
-      identity = identity.data
-    yield _certificate_fingerprint(identity)
+      cert = cert.data
+
+  return cert
```
timsutton, line 152: I don't think we would need this validation as long as we aren't using `plistlib` to parse the output, and so shouldn't be getting any data/bytes types.

tools/codesigningtool/codesigningtool.py
```diff
@@ -120,16 +123,100 @@ def _certificate_fingerprint(identity):
...
+      expiry_date = datetime.datetime.strptime(expiry_date, "%Y-%m-%d %H:%M:%S %z")
+      now = datetime.datetime.now(expiry_date.tzinfo)
+      if now > expiry_date:
+        continue
+
+      # This is valid identity, decode the certificate, extract
+      # Common Name and Fingerprint and handle their values accordingly
+      # as described above
+      cert = data_split[data_split.index(cert_start) + 1]
+      cert = base64.b64decode(cert)
```
timsutton, line 210: Would it be potentially easier to parse this (and parts earlier around `cert_data` if we instead extracted the cert using `re` and just capturing the string between `-----BEGIN CERTIFICATE-----` and `-----END CERTIFICATE-----`?

Technically I think it's also already in a PEM form which could be passed to `openssl -inform PEM` without needing to b64decode, however it doesn't have the line breaks every 64 characters so that'd have to be added

tools/codesigningtool/codesigningtool.py
```diff
@@ -120,16 +123,100 @@ def _certificate_fingerprint(identity):
...
+  for token in tokens:
+    # Used to find certificate info
+    cert_start = "-----BEGIN CERTIFICATE-----"
+    # All 'string' nodes at the same level of 'token'
+    all_string_nodes = smartcards_keychain.findall('.//*/string[.="{}"]../string'.format(token))
+    # Find only the 'string' nodes that hold certificate data and are not empty
+    cert_data = [s.text for s in all_string_nodes if cert_start in s.text and s.text.strip() != ""]
+
+    for data in cert_data:
+      data_split = data.split('\n')
```
timsutton, line 191: Nit: could also do `data.splitlines()` here or below

tools/codesigningtool/codesigningtool.py
```diff
@@ -120,16 +123,100 @@ def _certificate_fingerprint(identity):
...
+      # as described above
+      cert = data_split[data_split.index(cert_start) + 1]
+      cert = base64.b64decode(cert)
+      cert = _certificate_data(cert)
+      common_name = _certificate_common_name(cert)
+      fingerprint = _certificate_fingerprint(cert)
+      if identity == common_name:
+        return [fingerprint]
+      elif not identity:
+        ids.append(fingerprint)
```
timsutton, line 217: Alternatively could [drop](https://www.youtube.com/watch?v=JVVMMULwR4s&t=289s) this `elif` after the early return

tools/codesigningtool/codesigningtool.py
thiagohmcruz: That makes sense, will add the fallback logic

tools/codesigningtool/codesigningtool.py
thiagohmcruz: I just moved existing code. [This came from an old fix](https://github.com/bazelbuild/rules_apple/commit/11aa8d7fd2f99ff8841f29dfaee16b2ded9955bc) re different versions `plistlib` being used, not sure if we can just stop doing the check?

tools/codesigningtool/codesigningtool.py
thiagohmcruz: Let me give this a shot, also the code would look cleaner with such regex I think.

tools/codesigningtool/codesigningtool.py
thiagohmcruz: Totally missed this! Indeed the `elif` is not necessary here. I do have to `if not identity:` though since we only want to populate `ids` if `identity` is not present.

thiagohmcruz:
> One thing we probably will want to establish is if this constrains macOS versions that could be supported. Do we care about supporting anything older than Mojave (10.14) or Catalina (10.15), for example? That might help inform the logic in what's being searched for in the `system_profiler` output.

@timsutton wouldn't be enough to fallback to empty array when searching for `AVAIL_SMARTCARDS_TOKEN` and `AVAIL_SMARTCARDS_KEYCHAIN`? 

Not sure how I would go about checking what is available in each OS version, is there an easy way to do that that I'm not aware of?

tools/codesigningtool/codesigningtool.py
timsutton: I think so, because in parsing XML elements you should be only dealing with strings, whereas with plists objects could be stored as data or other types.

timsutton:
> @timsutton wouldn't be enough to fallback to empty array when searching for AVAIL_SMARTCARDS_TOKEN and AVAIL_SMARTCARDS_KEYCHAIN?

Yes, that seems like just a good defense anyway. I did just check a 10.14 system and these two keys are there as well (full output [here](https://gist.github.com/timsutton/b97e5b6eebf505692be3435d78563ec9).)

Another bit of context I'd forgotten to add in this PR is that persistent tokens (as would be used in Signing Manager or other apps providing a token via an app extension) only [showed up in 10.15.4](https://support.apple.com/en-ca/guide/deployment-reference-ios/apd05d8c6344/web). Prior to then it could have only been possible using physically-insertable smartcards, or that's my understanding anyway.

tools/codesigningtool/codesigningtool.py
thiagohmcruz: I am now using `plistlib` so I'll keep this

tools/codesigningtool/codesigningtool.py
thiagohmcruz: Using a regex for this has made things mode readable IMO, thx for the suggestion!

Regarding the PEM format comment, I'm reusing an existing function `_certificate_fingerprint` and adding a one liner to decode seems simpler than creating a new function for `-inform PEM` and handling the line breaks just for this?

tools/codesigningtool/codesigningtool.py
thiagohmcruz: Done

tools/codesigningtool/codesigningtool.py
thiagohmcruz: This is done

tools/codesigningtool/codesigningtool.py
thiagohmcruz: Using `plistlib` now and getting these keys in a way that empty arrays are returned if not present

tools/codesigningtool/codesigningtool.py
thiagohmcruz: Not necessary anymore since I'm using a regex to get the cert

tools/codesigningtool/codesigningtool.py
```diff
@@ -120,16 +122,119 @@ def _certificate_fingerprint(identity):
...
+  """Get available tokens from the output of 'system_profiler SPSmartCardsDataType -xml'"""
+  tokens = [x for x in xml if x.get("_name", None) == "AVAIL_SMARTCARDS_TOKEN"]
+
+  if len(tokens) == 0:
+    return []
+
+  tokens = tokens[0].get("_items", None)
+  tokens = [x.get("_name", None) for x in tokens]
+
+  if len(tokens) == 0:
```
thiagohmcruz, line 163: This is 100% not necessary, will remove it

tools/codesigningtool/codesigningtool.py
```diff
@@ -120,16 +122,119 @@ def _certificate_fingerprint(identity):
...
+    return []
 
+  return tokens
+
+def _get_smartcard_keychain(xml):
+  """Get keychain items from the output of 'system_profiler SPSmartCardsDataType -xml'"""
+  keychain = [x for x in xml if x.get("_name", None) == "AVAIL_SMARTCARDS_KEYCHAIN"]
+
+  if len(keychain) == 0:
```
thiagohmcruz, line 172: Same as for tokens, not necessary

thiagohmcruz:
@brentleyjones sorry about the direct ping! Just wanted to check if you know who I should bug to get some reviews?

tools/codesigningtool/codesigningtool.py
```diff
@@ -120,16 +122,116 @@ def _certificate_fingerprint(identity):
...
+  ids = []
+  _, xml, _ = execute.execute_and_filter_output([
+      "/usr/sbin/system_profiler",
+      "SPSmartCardsDataType",
+      "-xml"
+  ], raise_on_failure=True)
+  xml = plistlib.loads(str.encode(xml))
+  if len(xml) == 0:
+    return []
+  xml = xml[0].get("_items", None)
```
segiddins, line 186: do we need to consider all entries in here?

segiddins (approved):
It would be awesome to add some test coverage for the xml parsing being done, because some of it gets a bit complicated, but since there's no coverage for this tool at all yet 😶

tools/codesigningtool/codesigningtool.py
```diff
@@ -120,16 +122,116 @@ def _certificate_fingerprint(identity):
...
+    "/usr/bin/openssl",
+    "x509",
+    "-noout",
+    "-inform",
+    "DER",
+    "-subject"
+  ], inputstr=cert, raise_on_failure=True)
+  subject = subject.strip().split('/')
+  cert_cn = [f for f in subject if "CN=" in f][0]
+  cert_cn = cert_cn.replace("CN=", "")
```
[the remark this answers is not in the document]
thiagohmcruz: I guess we should assume it's possible. I'll add the check instead of accessing `[0]` here and make it return `None`. Tested locally and is all good since at the call site the comparison is `if identity == common_name:`

tools/codesigningtool/codesigningtool.py
```diff
@@ -120,16 +122,116 @@ def _certificate_fingerprint(identity):
...
+  return cert
+
+def _get_smartcard_tokens(xml):
+  """Get available tokens from the output of 'system_profiler SPSmartCardsDataType -xml'"""
+  tokens = [x for x in xml if x.get("_name", None) == "AVAIL_SMARTCARDS_TOKEN"]
+
+  if len(tokens) == 0:
+    return []
+
+  tokens = tokens[0].get("_items", None)
```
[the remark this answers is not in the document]
thiagohmcruz: [From the output](https://gist.github.com/thiagohmcruz/d94b747217cd4e25deeaa5f77f47561f#file-system_profiler_output-L75-L91) it's not an issue, all tokens will be located under `_items`

tools/codesigningtool/codesigningtool.py
thiagohmcruz: This is the top level `_items` that [contains all the info we need](https://gist.github.com/thiagohmcruz/d94b747217cd4e25deeaa5f77f47561f#file-system_profiler_output-L23). I think it's safe to ignore the rest for the purpose of this function

thiagohmcruz:
@segiddins should be good to :shipit:?

thiagohmcruz:
@brentleyjones, @keith mind taking a look? Let me know if there's anything else you think I should be addressing here 🙏

tools/codesigningtool/codesigningtool.py
```diff
@@ -108,7 +110,7 @@ def _parse_mobileprovision_file(mobileprovision_file):
 def _certificate_fingerprint(identity):
   """Extracts a fingerprint given identity in a mobileprovision file."""
   _, fingerprint, _ = execute.execute_and_filter_output([
-      "openssl",
+      "/usr/bin/openssl",
```
jerrymarino, line 113: I was curious about using `/usr/bin/openssl` v.s. running whatever is on the env? We may need to think about backwards compat: e.g. if people had a different openssl binary and now this method picks up the system one.

Maybe longer term we'd want to pass these as an input.

tools/codesigningtool/codesigningtool.py
```diff
@@ -120,16 +122,122 @@ def _certificate_fingerprint(identity):
   fingerprint = fingerprint.replace(":", "")
   return fingerprint
 
+def _certificate_common_name(cert):
+  _, subject, _ = execute.execute_and_filter_output([
+    "/usr/bin/openssl",
```
jerrymarino, line 127: Same as above

jerrymarino (commented):
A few inline for now, but overall really excited to see smartcard support in Bazel - this has been a long time coming 👍  One alternative to start using this could be to have a different codesigning tool in `rules_ios`

tools/codesigningtool/codesigningtool.py
brentleyjones: I agree. I would rather keep it un-anchored for now.

thiagohmcruz: I'll put it back. This came from a discussion with Tim to ensure the Apple system one and not some alternate shim or copy from brew. But if folks have any reason to override that I guess we should respect it, it makes sense to me.

cc @timsutton

thiagohmcruz: Done

thiagohmcruz:
>A few inline for now, but overall really excited to see smartcard support in Bazel - this has been a long time coming 👍 One alternative to start using this could be to have a different codesigning tool in rules_ios

@jerrymarino will work on bumping `rules_ios` next, we might be able to use it as is. Just needing to rebase you `rules_swift` fork I think. Still not sure what changes will be necessary but will look into it next.

tools/codesigningtool/codesigningtool.py
```diff
@@ -120,16 +122,122 @@ def _certificate_fingerprint(identity):
...
+  return keychain
+
+def _find_smartcard_identities(identity=None):
+  """Finds smartcard identitites on the current system."""
+  ids = []
+  _, xml, _ = execute.execute_and_filter_output([
+      "/usr/sbin/system_profiler",
+      "SPSmartCardsDataType",
+      "-xml"
+  ], raise_on_failure=True)
```
brentleyjones, line 188: What could cause a failure here? I'm mainly trying to think of cases where this new functionality could cause failures in builds that didn't have them before.

timsutton: I don't think I've ever actually seen `system_profiler` fail unless the datatype happened to be of some type that doesn't yet exist. I'd gone and checked that `SPSmartCardsDataType` is present at least as far back as macOS 10.14.

thiagohmcruz: This is just me copy pasting from the other similar invocations 🤦 I don't think we should expect any failure here. So we don't need `raise_on_failure=True`

brentleyjones: I'm not saying it shouldn't raise (it probably should), I was just wondering in which cases it could fail.

thiagohmcruz: That makes sense! Was about to say that it's probably good to keep it even though we haven't found any failures from this command

brentleyjones (approved):

brentleyjones:
Thanks @thiagohmcruz!

thiagohmcruz:
Thx you all for the feedback here!

jerrymarino:
@thiagohmcruz ok that sounds good to me. If we're blocked on the rules_swift PRs I'll ready a PR to get rid of the fork.

PR #1183, opened 2021-06-09
tools/xctoolrunner: handle cleaning non empty directories

steeve:
`os.removedirs` will fail if the directory isn't empty. This can happen when
using `ctx.actions.declare_directory`, which the Intents rules do.

Use `shutil.rmtree` instead.


thii (approved):

brentleyjones (approved):

brentleyjones:
Thanks @steeve!

PR #1184, opened 2021-06-10
Only read the `AppleFrameworkImportInfo.framework_imports` provider f…

segiddins:
…ield if it is set on the provider.

This matches other access patterns throughout the rules_apple code. Eventually, this should be cleaned up to make the field always present to avoid the ugliness of this access.

PiperOrigin-RevId: 377055205
(cherry picked from commit 324b730edb42998f793e90b43fa1b1463ee5037c)


keith:
have to resolve https://github.com/bazelbuild/rules_swift/pull/646#issuecomment-859072753 to get green CI here

thii (approved):

PR #1186, opened 2021-06-10
Propagate Swift-specific linker flags as an implicit dependency of the toolchain.

segiddins:
Propagate Swift-specific linker flags as an implicit dependency of the toolchain.

Apple binaries/bundles will automatically get these linker flags if they have any Swift target in their dependencies; this eliminates the need to pass them separately at the top level.

PiperOrigin-RevId: 378024438
(cherry picked from commit c253e4f)

# Conflicts:
#	apple/internal/binary_support.bzl

### DIFF
--- a/apple/internal/aspects/swift_dynamic_framework_aspect.bzl
+++ b/apple/internal/aspects/swift_dynamic_framework_aspect.bzl
@@ -82,11 +82,8 @@ framework module {module_name} {{
 def _swift_dynamic_framework_aspect_impl(target, ctx):
     """Aspect implementation for Swift dynamic framework support."""
 
-    if not hasattr(ctx.rule.attr, "deps"):
-        return []
-
-    swiftdeps = [x for x in ctx.rule.attr.deps if SwiftInfo in x]
-    ccinfos = [x for x in ctx.rule.attr.deps if CcInfo in x]
+    swiftdeps = [x for x in [target] if SwiftInfo in x]
+    ccinfos = [x for x in [target] if CcInfo in x]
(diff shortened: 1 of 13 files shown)


keith (approved):

apple/internal/aspects/swift_dynamic_framework_aspect.bzl
```diff
@@ -82,11 +82,8 @@ framework module {module_name} {{
 def _swift_dynamic_framework_aspect_impl(target, ctx):
     """Aspect implementation for Swift dynamic framework support."""
 
-    if not hasattr(ctx.rule.attr, "deps"):
-        return []
-
-    swiftdeps = [x for x in ctx.rule.attr.deps if SwiftInfo in x]
-    ccinfos = [x for x in ctx.rule.attr.deps if CcInfo in x]
+    swiftdeps = [x for x in [target] if SwiftInfo in x]
+    ccinfos = [x for x in [target] if CcInfo in x]
```
thii, line 86: cc @mccorkill1

segiddins:
Now passing with HEAD deps, so we should be able to merge this once we cut a `rules_swift` release

keith:
passed for us

apple/internal/aspects/swift_dynamic_framework_aspect.bzl
mccorkill1: This is great! It cleans up the implementation a lot and makes more sense without the swift_runtime_linkopts dependency

PR #1188, opened 2021-06-14
Invoke `lipo` from the Apple bundling rules instead of from within Bazel.

segiddins:
The use of `should_lipo = False` also changes the behavior of `link_multi_arch_binary` in the following ways that affected the rules:

*   Certain linker flags are now passed explicitly instead of being applied based on the calling rule's `binary_type`/`bundle_loader` attributes. These attributes have been removed from the rules, except in cases where it makes sense to have it remain explicit (for example, a `bundle_loader` on `macos_bundle`).
*   Framework dependencies are now passed explicitly via the `avoid_deps` parameter instead of being implied by the calling rule's `dylibs` attribute. The `dylibs` attribute has been removed from those rules.

PiperOrigin-RevId: 379283689
(cherry picked from commit 0b57552b8edb22539b865ed744fb9106b0be3f87)


thii:
Requires https://github.com/bazelbuild/bazel/commit/0643786736dbbe027611cddea112789aa6050cfd

keith:
more changes are starting to build on this which makes this a bit of a pain

brentleyjones (dismissed):
Waiting for another approval, since I resolved merge conflicts.

brentleyjones (approved):

thii (approved):

PR #1190, opened 2021-06-17
Add local provisioning profile rules

keith:
These rules allow you to fetch provisioning profiles from developer's
installed profiles to use for signing, instead of having to provide them
from source always. This allows a set of fallback profiles in case the
developer doesn't have any profiles for a given target

### DIFF
--- a/apple/internal/local_provisioning_profiles.bzl
+++ b/apple/internal/local_provisioning_profiles.bzl
@@ -1,0 +1,143 @@
+"""# Rules for using locally installed provisioning profiles"""
+
+def _local_provisioning_profiles(repository_ctx):
+    system_profiles_path = "{}/Library/MobileDevice/Provisioning Profiles".format(repository_ctx.os.environ["HOME"])
+    repository_ctx.execute(["mkdir", "-p", system_profiles_path])
+    repository_ctx.symlink(system_profiles_path, "profiles")
+
+    repository_ctx.file("BUILD", """
+filegroup(
+    name = "profiles",
+    srcs = glob(["profiles/*.mobileprovision"], allow_empty = True),
+    visibility = ["//visibility:public"],
+)
+
+filegroup(
+    name = "empty",
+    srcs = [],
+    visibility = ["//visibility:public"],
+)
+
+alias(
+    name = "fallback_profiles",
+    actual = "{}",
+    visibility = ["//visibility:public"],
+)
+""".format(repository_ctx.attr.fallback_profiles or ":empty"))
+
+local_provisioning_profiles = repository_rule(
+    environ = ["HOME"],
+    implementation = _local_provisioning_profiles,
+    attrs = {
+        "fallback_profiles": attr.label(
+            allow_files = [".mobileprovision"],
+        ),
+    },
+    doc = """
+This rule declares an external repository for discovering locally installed
+provisioning profiles. This is consumed by `ios_local_provisioning_profile`.
+You can optionally set 'fallback_profiles' to point at a stable location of
+profiles if a newer version of the desired profile does not exist on the local
+machine. This is useful for checking in the current version of the profile, but
+not having to update it every time a new device or certificate is added.
+
+## Example
+
+load("@build_bazel_rules_apple//apple:apple.bzl", "local_provisioning_profiles")
+
+local_provisioning_profiles(
+    name = "local_provisioning_profiles",
+    fallback_profiles = "//path/to/some:filegroup",
+)
+""",
+)
+
+def _local_provisioning_profile(ctx):
+    if not ctx.files._local_srcs and not ctx.attr._fallback_srcs:
+        ctx.fail("Either local or fallback provisioning profiles must exist")
+
+    selected_profile_path = "{name}.mobileprovision".format(name = ctx.attr.profile_name or ctx.attr.name)
+    selected_profile = ctx.actions.declare_file(selected_profile_path)
+
+    args = ctx.actions.args()
+    args.add(ctx.attr.profile_name or ctx.attr.name)
+    args.add(selected_profile)
+    if ctx.attr.team_id:
+        args.add("--team-id")
+        args.add(ctx.attr.team_id)
+    args.add("--local-profiles")
+    args.add_all(ctx.files._local_srcs)
+    args.add("--fallback-profiles")
+    args.add_all(ctx.files._fallback_srcs)
+
+    ctx.actions.run(
+        executable = ctx.executable._finder,
+        arguments = [args],
+        inputs = ctx.files._local_srcs + ctx.files._fallback_srcs,
+        outputs = [selected_profile],
+        execution_requirements = {"no-sandbox": "1", "no-remote-exec": "1"},
+        progress_message = "Finding provisioning profile %{label}",
+    )
+
+    return [DefaultInfo(files = depset([selected_profile]))]
+
+local_provisioning_profile = rule(
+    attrs = {
+        "profile_name": attr.string(
+            doc = "Name of the profile to use, if it's not provided the name of the rule is used",
+        ),
+        "team_id": attr.string(
+            doc = "Team ID of the profile to find. This is useful for disambiguating between multiple profiles with the same name on different developer accounts.",
+        ),
+        "_fallback_srcs": attr.label(
+            default = "@local_provisioning_profiles//:fallback_profiles",
+        ),
+        "_local_srcs": attr.label(
+            default = "@local_provisioning_profiles//:profiles",
+        ),
+        "_finder": attr.label(
+            cfg = "exec",
+            default = "@build_bazel_rules_apple//tools/local_provisioning_profile_finder",
+            executable = True,
+        ),
+    },
+    implementation = _local_provisioning_profile,
+    doc = """
+This rule declares a bazel target that you can pass to the
+'provisioning_profile' attribute of rules that require it. It discovers a
+provisioning profile for the given attributes either on the user's local
+machine, or with the optional 'fallback_profiles' passed to
+'local_provisioning_profiles'. This will automatically pick the newest profile
+if there are multiple profiles matching the given criteria. By default this
+rule will search for a profile with the same name as the rule itself, you can
+pass profile_name to use a different name, and you can pass team_id if you'd
+like to disambiguate between 2 Apple developer accounts that have the same
+profile name.
+
+## Example
+
+load("@build_bazel_rules_apple//apple:apple.bzl", "local_provisioning_profile")
+
+local_provisioning_profile(
+    name = "app_debug_profile",
+    profile_name = "Development App",
+    team_id = "abc123",
+)
+
+ios_application(
+    name = "app",
+    ...
+    provisioning_profile = ":app_debug_profile",
+)
+
+local_provisioning_profile(
+    name = "app_release_profile",
+)
+
+ios_application(
+    name = "release_app",
+    ...
+    provisioning_profile = ":app_release_profile",
+)
+""",
+)
--- a/apple/internal/local_provisioning_profiles.bzl
+++ b/apple/internal/local_provisioning_profiles.bzl
@@ -1,6 +1,6 @@
 """# Rules for using locally installed provisioning profiles"""
 
-def _local_provisioning_profiles(repository_ctx):
+def _provisioning_profile_repository(repository_ctx):
     system_profiles_path = "{}/Library/MobileDevice/Provisioning Profiles".format(repository_ctx.os.environ["HOME"])
     repository_ctx.execute(["mkdir", "-p", system_profiles_path])
     repository_ctx.symlink(system_profiles_path, "profiles")
@@ -25,9 +25,9 @@ alias(
 )
 """.format(repository_ctx.attr.fallback_profiles or ":empty"))
 
-local_provisioning_profiles = repository_rule(
+provisioning_profile_repository = repository_rule(
     environ = ["HOME"],
-    implementation = _local_provisioning_profiles,
+    implementation = _provisioning_profile_repository,
     attrs = {
         "fallback_profiles": attr.label(
             allow_files = [".mobileprovision"],
@@ -43,11 +43,29 @@ not having to update it every time a new device or certificate is added.
 
 ## Example
 
-load("@build_bazel_rules_apple//apple:apple.bzl", "local_provisioning_profiles")
+### In your `WORKSPACE` file:
 
-local_provisioning_profiles(
+load("@build_bazel_rules_apple//apple:apple.bzl", "provisioning_profile_repository")
+
+provisioning_profile_repository(
     name = "local_provisioning_profiles",
-    fallback_profiles = "//path/to/some:filegroup",
+    fallback_profiles = "//path/to/some:filegroup", # Optional profiles to use if one isn't found locally
+)
+
+### In your `BUILD` files (see `local_provisioning_profile` for more examples):
+
+load("@build_bazel_rules_apple//apple:apple.bzl", "local_provisioning_profile")
+
+local_provisioning_profile(
+    name = "app_debug_profile",
+    profile_name = "Development App",
+    team_id = "abc123",
+)
+
+ios_application(
+    name = "app",
+    ...
+    provisioning_profile = ":app_debug_profile",
 )
 """,
 )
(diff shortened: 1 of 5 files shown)


keith:
This is a RFC

apple/internal/local_provisioning_profiles.bzl
```diff
@@ -0,0 +1,80 @@
...
+        execution_requirements = {"no-sandbox": "1"},
+    )
+
+    return [DefaultInfo(files = depset([selected_profile]))]
+
+local_provisioning_profile = rule(
+    attrs = dict(
+        _fallback_srcs = attr.label(
+            allow_files = [".mobileprovision"],
+            default = "@local_provisioning_profiles//:profiles",
```
brentleyjones, line 68:
```suggestion
            default = "@local_provisioning_profiles//: fallback_profiles",
```

brentleyjones:
Should `local_provisioning_profile` explicitly mention `@local_provisioning_profiles`, or specifying `fallback_profiles` on `local_provisioning_profile`? I could see larger repos wanting different `fallback_profiles`, though just merging them into one `filegroup` is an option.

keith:
Interested to hear what others think, we could definitely allow double overrides for that, in our case we don't need that, so I figured we should wait until someone did

apple/internal/local_provisioning_profiles.bzl
```diff
@@ -0,0 +1,80 @@
...
+alias(
+    name = "fallback_profiles",
+    actual = "{}",
+    visibility = ["//visibility:public"],
+)
+""".format(repository_ctx.attr.fallback_profiles or ":empty"))
+
+local_provisioning_profiles = repository_rule(
+    environ = ["HOME"],
+    local = True,
```
jerrymarino, line 33: Should this be local? Caveat it's gotta run every time.

Because you symlink provisioning profiles directory into the external dir, it seems like enough to just glob the profiles for invalidation?

apple/internal/local_provisioning_profiles.bzl
```diff
@@ -0,0 +1,80 @@
+"""
+Bazel rules for using locally installed provisioning profiles for debug builds
+"""
+
+def _local_provisioning_profiles(repository_ctx):
+    repository_ctx.symlink(
```
jerrymarino, line 6: I think making this symlink might obviate the need to make external. Caveat: if it doesn't exist but somehow exists in the future.

You may want to run `mkdir -p` on this dir to handle that and in general if it may not exist on a new system

jerrymarino (approved):
@keith thanks for adding this, I'm really looking forward to this feature. Minor comments nine for invalidation / local running

apple/internal/local_provisioning_profiles.bzl
```diff
@@ -0,0 +1,80 @@
...
+    name = "profiles",
+    srcs = glob(["profiles/*.mobileprovision"]),
+    visibility = ["//visibility:public"],
+)
+
+filegroup(
+    name = "empty",
+    srcs = [],
+    visibility = ["//visibility:public"],
+)
```
erikkerber, line 22: Docs pending of course, but should the empty case work (allow the finder to find profiles) if no fallback_profiles are specified?

```bash
in (an implicit dependency) attribute of local_provisioning_profile rule //MyApp: '@LocalProfiles//:empty' does not produce any local_provisioning_profile $fallback_srcs files (expected .mobileprovision)
```

https://github.com/bazelbuild/rules_apple/pull/1190/files#diff-6af425aecce524a86c66a19c86dc87aca95abb271885b0aac0fac4801ba92d49R67

keith: Ah thanks I probably need to allow empty or something. I added this in this adaptation since we don't actually have this case so I didn't test it well enough

tools/local_provisioning_profile_finder/local_provisioning_profile_finder.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+def _profile_contents(profile: str) -> Tuple[str, datetime.datetime]:
+    output = subprocess.check_output(["security", "cms", "-D", "-i", profile])
+    plist = plistlib.loads(output)
+    return plist["Name"], plist["CreationDate"]
+
+
+def _find_newest_profile(name: str, profiles: List[str]) -> Optional[str]:
+    sanitized_name = name.replace("_", " ")
```
erikkerber, line 35: This will prevent anyone from having actual underscores in their profile names 😅

keith: ah yea, maybe only works for us

erikkerber: Probably works for most, since I assume "spaces in the name" outscore "underscores in the name" by quite a large margin.

Worth mentioning in any docs, of course, but maybe also refer to the "design limitation" in the error message below. (Ignoring the possibility if this sanitizaition can be worked around of course – I may offer a patch later if possible. Or maybe we'll just change our profile names too)

keith: we can probably use something besides name instead, we could make folks pass the bundle ID since that is more stable anyways

tools/local_provisioning_profile_finder/local_provisioning_profile_finder.py
```diff
@@ -0,0 +1,67 @@
...
+    newest_path: Optional[str] = None
+    newest_date: Optional[datetime.datetime] = None
+    for profile in profiles:
+        name, creation_date = _profile_contents(profile)
+        if name != sanitized_name:
```
erikkerber, line 40: Rare but it might be worth adding `OR team_id != team_id`

and by rare I mean of course this applies to us

erikkerber: (Unless you can extract the bundle ID like you mentioned above which I do see in the Finder preview but not in the output of the security tool)

apple/internal/local_provisioning_profiles.bzl
```diff
@@ -0,0 +1,80 @@
...
+    args.add_all(ctx.files._local_srcs)
+    args.add("--fallback-profiles")
+    args.add_all(ctx.files._fallback_srcs)
+
+    ctx.actions.run(
+        executable = ctx.executable._finder,
+        arguments = [args],
+        inputs = ctx.files._local_srcs + ctx.files._fallback_srcs,
+        outputs = [selected_profile],
+        execution_requirements = {"no-sandbox": "1"},
```
brentleyjones, line 59:
```suggestion
        execution_requirements = {"no-sandbox": "1", "no-remote-exec": "1"},
```

apple/internal/local_provisioning_profiles.bzl
```diff
@@ -0,0 +1,80 @@
...
+    implementation = _local_provisioning_profiles,
+    attrs = dict(
+        fallback_profiles = attr.label(
+            allow_files = [".mobileprovision"],
+        ),
+    ),
+)
+
+def _local_provisioning_profile(ctx):
+    selected_profile_path = "{name}.mobileprovision".format(name = ctx.label.name)
```
erikkerber, line 43: An optional `profile_name` would be useful. This could default to `name` if one isn't provided of course, but keeping the label constant but `select`-ing between different profile names is a use case that we currently have.

erikkerber: One more rationale: target names can't have `:` in them, and Xcode managed profiles will be called something like `iOS Team Provisioning Profile: com.my.app`.

(Ignore for a moment why one would need Xcode managed profiles 😸)

erikkerber, line 43:
```suggestion
    selected_profile_path = "{name}.mobileprovision".format(name = ctx.attr.profile_name or ctx.label.name)
```

apple/internal/local_provisioning_profiles.bzl
```diff
@@ -0,0 +1,80 @@
...
+    selected_profile = ctx.actions.declare_file(selected_profile_path)
+
+    args = ctx.actions.args()
+    args.add(ctx.attr.name)
```
erikkerber, line 47:
```suggestion
    args.add(ctx.attr.profile_name or ctx.attr.name)
```

tinder-maxwellelliott:
I just wrote a version of this locally, would be great if we could get this merged in! Thanks @keith

apple/internal/local_provisioning_profiles.bzl
```diff
@@ -0,0 +1,80 @@
...
+        _fallback_srcs = attr.label(
+            allow_files = [".mobileprovision"],
+            default = "@local_provisioning_profiles//:fallback_profiles",
+        ),
+        _local_srcs = attr.label(
+            default = "@local_provisioning_profiles//:profiles",
+        ),
+        _finder = attr.label(
+            cfg = "exec",
+            default = "@build_bazel_rules_apple//tools/local_provisioning_profile_finder",
```
tinder-maxwellelliott, line 75:
```suggestion
            default = "@build_bazel_rules_apple//tools:local_provisioning_profile_finder",
```

apple/internal/local_provisioning_profiles.bzl
tinder-maxwellelliott: Running into this as well

tinder-maxwellelliott:
```suggestion
    selected_profile_path = "{name}.mobileprovision".format(name = ctx.attr.profile_name or ctx.attr.name)
```

apple/internal/local_provisioning_profiles.bzl
```diff
@@ -0,0 +1,80 @@
...
+        inputs = ctx.files._local_srcs + ctx.files._fallback_srcs,
+        outputs = [selected_profile],
+        execution_requirements = {"no-sandbox": "1", "no-remote-exec": "1"},
+    )
+
+    return [DefaultInfo(files = depset([selected_profile]))]
+
+local_provisioning_profile = rule(
+    attrs = dict(
+        _fallback_srcs = attr.label(
```
tinder-maxwellelliott, line 66:
```suggestion
        profile_name = attr.string(),
        _fallback_srcs = attr.label(
```

tinder-maxwellelliott:
We have been using an extremely similar version of this logic for the last two weeks and its been working great. I think if we can get the last suggestions in this is good to go.

tinder-maxwellelliott (approved):

ThreeFingerCoder:
@keith Do you plan to merge this PR? I have pulled in these changes and was able to use it with the local provisioning profiles successfully and would personally like to have this change available in the upstream. If not, I am more than happy to pick up whatever is left!

keith:
I would like to merge something like this, but yea I clearly haven't had the time to clean it up. I have some more internal changes that have been helpful for us that I'd like to throw up here ideally. I also haven't investigated the PR feedback here at all yet 😞

ThreeFingerCoder:
@keith thanks for the reply! Will wait for the updated PR then!

apple/internal/local_provisioning_profiles.bzl
keith: I don't think this is right? Since the tool full label is `tools/local_provisioning_profile_finder:local_provisioning_profile_finder`?

tools/local_provisioning_profile_finder/local_provisioning_profile_finder.py
```diff
@@ -0,0 +1,67 @@
...
+        "--fallback-profiles",
+        nargs="+",
+        required=True,
+        help="Fallback provisioning profiles to use if not found locally",
+    )
+    return parser
+
+
+def _profile_contents(profile: str) -> Tuple[str, datetime.datetime]:
+    output = subprocess.check_output(["security", "cms", "-D", "-i", profile])
```
ThreeFingerCoder, line 29: I met an issue running `security` on CI and was able to prevent the permission error with the following change : https://github.com/autopkg/recipes/commit/657b2eb51f22dd14d44b90a0ae7a1d3e3313087e

apple/internal/local_provisioning_profiles.bzl
keith: Yea I think you're right!

tools/local_provisioning_profile_finder/local_provisioning_profile_finder.py
keith: Adding this as we also had this case

tools/local_provisioning_profile_finder/local_provisioning_profile_finder.py
keith: Ouch what was the error you saw? we successfully run this on CI today, with the standard keychain setup workarounds required for signing. So if I understand correctly your workaround is to create a dummy keychain and point the command to that?

ThreeFingerCoder: The error we saw was : 
```
security: cert import failed: write permissions error
security: problem decoding
```

Based on the context I read about why this issue happens, it's possible that the error could happen due to missing field in the `.mobileprovision` file being passed to the `security` command. So, it's also possible that you don't need this dummy keychain in Lyft's environment. 

> So if I understand correctly your workaround is to create a dummy keychain and point the command to that?
Yes

keith: Don't you also have a keychain setup on CI to do the actual signing? We're just relying on that one I guess

apple/internal/local_provisioning_profiles.bzl
```diff
@@ -0,0 +1,142 @@
...
+You can optionally set 'fallback_profiles' to point at a stable location of
+profiles if a newer version of the desired profile does not exist on the local
+machine. This is useful for checking in the current version of the profile, but
+not having to update it every time a new device or certificate is added.
+
+## Example
+
+load("@build_bazel_rules_apple//apple:apple.bzl", "local_provisioning_profiles")
+
+local_provisioning_profiles(
```
keith, line 48: these rule names are maybe a bit too similar, maybe we should name this `provisioning_profile_repository` or something? up for suggestions!

BalestraPatrick: I think that would be better, and maybe add an example in the docs about how it interacts with `local_provisioning_profile`?

keith: Updated!

BalestraPatrick (approved):

apple/internal/local_provisioning_profiles.bzl
```diff
@@ -0,0 +1,161 @@
+"""# Rules for using locally installed provisioning profiles"""
+
+def _provisioning_profile_repository(repository_ctx):
+    system_profiles_path = "{}/Library/MobileDevice/Provisioning Profiles".format(repository_ctx.os.environ["HOME"])
+    repository_ctx.execute(["mkdir", "-p", system_profiles_path])
+    repository_ctx.symlink(system_profiles_path, "profiles")
+
+    repository_ctx.file("BUILD", """
```
thii, line 8: This helps remove the empty line at the top of the generated BUILD file.
```suggestion
    repository_ctx.file("BUILD", """\
```

apple/internal/local_provisioning_profiles.bzl
```diff
@@ -0,0 +1,161 @@
...
+    args = ctx.actions.args()
+    args.add(ctx.attr.profile_name or ctx.attr.name)
+    args.add(selected_profile)
+    if ctx.attr.team_id:
+        args.add("--team-id")
+        args.add(ctx.attr.team_id)
+    args.add("--local-profiles")
+    args.add_all(ctx.files._local_srcs)
+    args.add("--fallback-profiles")
+    args.add_all(ctx.files._fallback_srcs)
```
thii, line 89: Should argument name and value be in the same call?
```suggestion
        args.add("--team-id", ctx.attr.team_id)
    args.add_all("--local-profiles", ctx.files._local_srcs)
    args.add_all("--fallback-profiles", ctx.files._fallback_srcs)
```

tools/local_provisioning_profile_finder/local_provisioning_profile_finder.py
```diff
@@ -0,0 +1,86 @@
...
+    parser.add_argument("name", help="The name of the profile to find")
+    parser.add_argument("output", help="The path to copy the profile to")
+    parser.add_argument(
+        "--local-profiles",
+        nargs="*",
+        required=True,
+        help="All local provisioning profiles to search through",
+    )
+    parser.add_argument(
+        "--fallback-profiles",
```
thii, line 21:
```suggestion
        "--fallback_profiles",
```

tools/local_provisioning_profile_finder/local_provisioning_profile_finder.py
```diff
@@ -0,0 +1,86 @@
...
+        nargs="*",
+        required=True,
+        help="Fallback provisioning profiles to use if not found locally",
+    )
+    parser.add_argument(
+        "--team-id",
```
thii, line 27:
```suggestion
        "--team_id",
```

thii (approved):

apple/internal/local_provisioning_profiles.bzl
keith: Cool!

PR #1191, opened 2021-06-19
Detect architecture in the macOS test runner

Aigeruth:
Hardcoded architecture results in failure on Apple Silicon with
arm64 build of 4.1.0. Removing this parameter enables autodection.


keith:
Nice. But can we just not pass it? Seems like xcodebuild should detect it as needed

Aigeruth:
> Nice. But can we just not pass it? Seems like xcodebuild should detect it as needed

Good spot, that works too. 👍 I've updated the PR.

keith:
Thanks!

Issue #1189, opened 2021-06-16
Skipping tests in non-macOS environment


Aigeruth:
Hey, I've started using `rules_apple` recently and noticed that the [_macos_test_runner_impl](https://github.com/bazelbuild/rules_apple/blob/c99b0fd9f58a75ecc5324b073bc1e901d9e0e55b/apple/testing/default_runner/macos_test_runner.bzl) function gets called in the [Analysis Phase](https://docs.bazel.build/versions/main/guide.html#analysis-phase) and this results in failure in non-macOS environments.

I've `bazel test //...` running in GitHub Actions on Ubuntu Linux and the pipeline fails when Bazel loads macOS unit test targest. I tried to apply `target_compatible_with = ["@platforms//os:macos"],` on the `macos_unit_test()` target, but it still fails with missing XCode configuration.

I can see the following error:

```
ERROR: /home/runner/.cache/bazel/_bazel_runner/88966966f60ea6cf3f8f0b764058cb22/external/build_bazel_rules_apple/apple/testing/default_runner/BUILD:57:18: in macos_test_runner rule @build_bazel_rules_apple//apple/testing/default_runner:macos_default_runner: 

Traceback (most recent call last):
	File "/home/runner/.cache/bazel/_bazel_runner/88966966f60ea6cf3f8f0b764058cb22/external/build_bazel_rules_apple/apple/testing/default_runner/macos_test_runner.bzl", line 76, column 62, in _macos_test_runner_impl
		substitutions = _get_xctestrun_template_substitutions(ctx),
	File "/home/runner/.cache/bazel/_bazel_runner/88966966f60ea6cf3f8f0b764058cb22/external/build_bazel_rules_apple/apple/testing/default_runner/macos_test_runner.bzl", line 34, column 47, in _get_xctestrun_template_substitutions
		if xcode_support.is_xcode_at_least_version(xcode_version, "10.0"):
	File "/home/runner/.cache/bazel/_bazel_runner/88966966f60ea6cf3f8f0b764058cb22/external/build_bazel_apple_support/lib/xcode_support.bzl", line 35, column 13, in _is_xcode_at_least_version
		fail("Could not determine Xcode version at all. This likely means Xcode isn't available; " +
Error in fail: Could not determine Xcode version at all. This likely means Xcode isn't available; if you think this is a mistake, please file an issue.
[...]

PR #1192, opened 2021-06-19
Detect XCode version gracefully

Aigeruth:
is_xcode_at_least_version calls fail() when it cannot detect an XCode
version. This happens on non-macOS platforms. As this function is
called in Analysis Phase, it makes build fail early and potenitally
prevents the usage of target_compatible_with.

Relates to #1189

### DIFF
--- a/apple/testing/default_runner/macos_test_runner.bzl
+++ b/apple/testing/default_runner/macos_test_runner.bzl
@@ -25,13 +25,20 @@ load(
 
 def _get_xctestrun_template_substitutions(ctx):
     """Returns the template substitutions for the xctestrun template."""
-    xcode_version = ctx.attr._xcode_config[apple_common.XcodeVersionConfig]
+    xcode_config = ctx.attr._xcode_config[apple_common.XcodeVersionConfig]
+
+    # Xcode version is only set when Xcode is installed. Xcode tool
+    # chain can be available both on macOS and Linux.
+    if not xcode_config.xcode_version():
+        return {
+            "%(xctestrun_insert_libraries)s": "",
+        }
 
     # Xcode 10 introduced new dylibs that are required when running unit tests, so we need to
     # provide different values in the xctestrun file that describes the test.
     # TODO(kaipi): Revisit this when Nitro has support for macOS. Nitro should be the one detecting
     #              Xcode version and configuring it appropriately.
-    if xcode_support.is_xcode_at_least_version(xcode_version, "10.0"):
+    if xcode_support.is_xcode_at_least_version(xcode_config, "10.0"):
         xctestrun_insert_libraries = [
             "__PLATFORMS__/MacOSX.platform/Developer/usr/lib/libXCTestBundleInject.dylib",
             "__DEVELOPERUSRLIB__/libMainThreadChecker.dylib",


apple/testing/default_runner/macos_test_runner.bzl
```diff
@@ -14,38 +14,40 @@
...
+    XCTESTRUN_INSERT_LIBRARIES = "%(xctestrun_insert_libraries)s"
+    substitutions = {
+        XCTESTRUN_INSERT_LIBRARIES: "",
+    }
+
+    xcode_version = ctx.attr._xcode_config[apple_common.XcodeVersionConfig].xcode_version()
+
+    # XCode version is not present on non-macOS platforms
+    if not xcode_version:
+        return substitutions
```
thii, line 33: This would hide the actual failure if you build on macOS _and_ Xcode isn't present.

Also, `xcode_version()` can be present on non-macOS platforms with a cross-compiling toolchain.

I'm not sure how to deal with this 😢

Aigeruth: I think, it shouldn't hide the failure when XCode isn't present on macOS. I would expect that this change just moves the failure to the Execution Phase. I assume the build rule will look for an clang/llvm binary and won't find it. I could try to test this by uninstalling XCode but it would probably useful to capture this in a CI environment too.

Good point on `xcode_version()` being present during cross-platform compilation. Is any of these rules used for cross-compiling to other platfroms? If not, I think, these rules will be just skipped and it should be safe.

brentleyjones: https://github.com/apple-cross-toolchain/rules_applecross

Aigeruth: This looks useful. Thanks for sharing, @brentleyjones ! I'll give it a try.

Out of curiosity: what are your thoughts around using `target_compatible_with`? Some people might opt for not setting up a toolchain on Linux.

brentleyjones: It would be useful to be able to use `target_compatible_with`, but we need to be able to support cross-compiling and when Xcode isn't available on macOS.

Aigeruth: It's probably my lack of experience with rules_apple, so I'll try to summaries the cases to consider:

- macOS + XCode
- macOS and no XCode
- non-macOS platform + XCode
- non-macOS platform and no XCode

`if not xcode_version:` should not change the current logic as the `is_xcode_at_least_version()` contains the following:

```starlark
    current_version = xcode_config.xcode_version()
    if not current_version:
        fail("Could not determine Xcode version at all. This likely means Xcode isn't available; " +
             "if you think this is a mistake, please file an issue.")
```

This change should just inline the [is_xcode_at_least_version()](https://github.com/bazelbuild/apple_support/blob/master/lib/xcode_support.bzl#L17) function and make it graceful by returning an 'empty' dictionary.

Only difference I can see that if the target is actually executed, it fails at execution phase and not analysis phase. I'm not sure what you mean by hiding the error, because I would still expect the build to fail if XCode is not present, just at a different time and perhaps with a different reason. I think, there a potential to improve developer experience and giving clear feedback on missing XCode by doing a check in `macos_test_runner.template.sh`. What do you think about that approach? Would that work?

Aigeruth: Yes, I'm aware that this runner can be overridden. I experimented with adding a `select()` and using the `//test/testdata/rules:dummy_runner` on Linux. However, this file still gets loaded and analysed by Bazel and that makes the build fail. My understanding is because this particular code executes during the creation of the `test_runner_rule()`, so it happens before it's even used but please correct me if I misunderstood this.

I've reverted the refactoring. 👍

thii: What I mean was:

1. Copy this test runner rule into your repo, for example into `rules/macos_test_runner.bzl`.
2. Make changes to it as you like.
3. Create a new runner target using your new rule, for example `//:custom_macos_runner`.
4. Pass the new runner to your test target's `runner` attribute.

```starlark
# BUILD
load("//rules:macos_test_runner.bzl", "macos_test_runner")

macos_test_runner(
    name = "custom_macos_runner",
)

# ...

macos_unit_test(
    name = "ButtonsMacLogicTests",
    bundle_id = "com.google.logic",
    minimum_os_version = "10.10",
    deps = [":ButtonsMacTestsLib"],
    runner = ":custom_macos_runner",
    target_compatible_with = ["@platforms//os:macos"],
)
```

This way, the default runner won't be evaluated in the analysis phase anymore.

apple/testing/default_runner/macos_test_runner.bzl
```diff
@@ -14,24 +14,27 @@
...
     "@build_bazel_rules_apple//apple/testing:apple_test_rules.bzl",
     "AppleTestRunnerInfo",
 )
 
 def _get_xctestrun_template_substitutions(ctx):
     """Returns the template substitutions for the xctestrun template."""
-    xcode_version = ctx.attr._xcode_config[apple_common.XcodeVersionConfig]
+    xcode_version = ctx.attr._xcode_config[apple_common.XcodeVersionConfig].xcode_version()
+
+    # XCode version is not present on non-macOS platforms
```
thii, line 26: Nit: XCode -> Xcode

apple/testing/default_runner/macos_test_runner.bzl
```diff
@@ -14,24 +14,27 @@
...
+    if not xcode_version:
+        return {
+            "%(xctestrun_insert_libraries)s": "",
+        }
 
     # Xcode 10 introduced new dylibs that are required when running unit tests, so we need to
     # provide different values in the xctestrun file that describes the test.
     # TODO(kaipi): Revisit this when Nitro has support for macOS. Nitro should be the one detecting
     #              Xcode version and configuring it appropriately.
-    if xcode_support.is_xcode_at_least_version(xcode_version, "10.0"):
```
thii, line 34: Since it's already returning above, I think it's safe to use `is_xcode_at_least_version` here?

thii (approved):

thii:
Thanks!

PR #1197, opened 2021-07-20
Set minimum_deployment_os_version in AppleBundleInfo provider

jschear:
This field was added to the provider in https://github.com/bazelbuild/rules_apple/pull/1099/files#diff-f06c6d6866f9332aa9ea85f2e0ccc46f67dfa8e67814a57f243bad6441856f4eR75, but it looks like it was never populated.


brentleyjones (approved):

brentleyjones:
Thanks @jschear!

PR #1198, opened 2021-07-21
Introduce a new xcodeproj rule to generate an Xcode project

zllak:
This PR introduces a new way to generate an Xcode project (aka xcodeproj) from ios_application rules.
This PR is a conversation starter, and a base for others to test and build on. We have discussed with a lot of companies about this and they are really interested.

This is heavily inspired by [rules_ios](https://github.com/bazel-ios/rules_ios) and leverage [Tulsi](https://github.com/bazelbuild/tulsi)'s aspects. 

## What's in this PR

This PR introduces a new `xcodeproj` rule that will be used to generate a valid xcodeproj that Xcode can open and use.

## How does it work

As of today, we leverage tulsi's aspect to generate `.tulsiinfo` JSON files. Those JSON files are then fed to a python script that will generate another JSON file that `XCodeGen` can then take to generate the final xcodeproj.
Because Xcode handles weirdly its sources with relative path, it's easier to move the generated xcodeproj outside the bazel sandbox and in bazel's workspace directory.
We make Xcode build using bazel through pre build scripts, and stub all tools for Xcode to make it believe it actually built something (cc, swiftc, ld, ...).
We leverage rules_swift index_while_building option to generate the indexstore that we move to the DerivedData for sourcekit.
We also move all generated swiftmodule/swiftdoc to DerivedData.

## How to test this PR

In one of your `BUILD.bazel` file, simply use the new `xcodeproj` rule:

```python
load("@build_bazel_rules_apple//apple:xcodeproj.bzl", "xcodeproj")

xcodeproj(
    name = "xcodeproj",
    project_name = "Zenly",
    deps = ["//Zenly:Zenly"], # your ios_application
)
```

To start fresh, it's better to first clean up the DerivedData

```bash
$ rm -rf ~/Library/Developer/Xcode/DerivedData
```
Clean bazel
```bash
$ bazel clean
```
Then run the xcodeproj
```bash
$ bazel run //path/to:xcodeproj
```

## What's working

- Generating a valid xcodeproj
- Indexing of the whole project
- Build the whole project with bazel
- Code completion, go to def, build as you type, etc... Xcode full development experience

## What's not working yet/improved

- Some `objc_library` cannot be properly linked/included because tulsi does not like path like `bazel-out/ios-x86_64...` so it rewrites them as `bazel-tulsi-includes/x/x/` which is unusable for us.
- SwiftUI previews are not working
- Handling of infoplist/entitlements
- It has not been tested with more than one `ios_application` as `deps` in `xcodeproj`

## What to do
- [ ] Do not rely on tulsi's aspect and rebuild our own
- [ ] Support multiple ios_application as source for `xcodeproj`

The work is still underway and this PR will be updated often until it's mergeable. It's also here for others to test and contribute.


jerrymarino:
I saw the mention of rules_ios and couldn't help but give this a drive by. IMO, there's still a lot to be done on the `rules_ios` one which is similar functionality to [XCHammer](https://github.com/bazel-ios/xchammer)'s rule. I was hoping one day the two rues converge to support `rules_ios`'s format but carry the DSL, performance, API, and feature set of XCHammer which uses Tulsi's core including, aspects, build scripts, and other bits.

Today, there really isn't really a 100% complete Xcode generator: one that correctly handles all cases and bundle types like extensions, widgets, profiling, new build system, new code signing, xcconfig integration, and progress bar, incremental indexing, and community rules like rules_ios.  Someone might wonder reasons to spin up another generator instead of working through existing generator's issues.

Anyhow, I was curious if there was a design or plan to align build settings with Xcode and Bazel. Unless Xcode has identical build settings / compiler invocations to Bazel, the IDE will fail to work correctly. This may not impact some of the homogenous apps out there, but for more configured ones there's a lot of sharp edges.

jerrymarino:
It just dawned on me that I didn't mention I'm a maintainer of both XCHammer and rules_ios 🤦 . I'd love to know if there are any technical reasons of why neither worked out.

The feedback would be valuable to help these repos improve, and also maybe we could collab on something. If it'd be easier to discuss offline, I'd be happy to sync up sometime!

zllak:
From what we tested, nothing was working on our project. The only one that gives an almost useful Xcode project is Tulsi, but unfortunately it takes very long to generate the project, and it generates it at runtime and not build time.
We tried rules_ios, don't really remember what was not working but we did not end up with a valid Xcode project with all the features.
Also, we want to avoid fragmenting the ecosystem and bring this important feature back into rules_apple.

jschear:
CLA bot please chill out

keith:
nice, i do think the ideal state would be there was a common path for this as long as it was general enough

steeve:
So far so good, our devs says it's a quantum leap vs tulsi.

PR #1202, opened 2021-07-26
Clone file on macOS in bundletool_experimental.py

brentleyjones:


tools/bundletool/bundletool_experimental.py
```diff
@@ -199,7 +200,10 @@ def _copy_file(self, src, dest, executable, bundle_root):
       raise BundleConflictError(dest)
 
     self._makedirs_safely(os.path.dirname(full_dest))
-    shutil.copy(src, full_dest)
+    if sys.platform == "darwin":
+      subprocess.check_output(["cp", "-c", src, full_dest])
+    else:
+      shutil.copy(src, full_dest)
     os.chmod(full_dest, 0o755 if executable else 0o644)
```
keith, line 207: does this not qualify as "writing" and result in it immediately being duplicated? maybe not if they don't actually change here depending on initial permissions?

tools/bundletool/bundletool_experimental.py
```diff
@@ -199,7 +200,10 @@ def _copy_file(self, src, dest, executable, bundle_root):
       raise BundleConflictError(dest)
 
     self._makedirs_safely(os.path.dirname(full_dest))
-    shutil.copy(src, full_dest)
+    if sys.platform == "darwin":
+      subprocess.check_output(["cp", "-c", src, full_dest])
```
keith, line 204:
```suggestion
      subprocess.check_output(["/bin/cp", "-c", src, full_dest])
```

keith (approved):

tools/bundletool/bundletool_experimental.py
brentleyjones: It might only change a block. AFPS allows for partial file modifications with the clones.

brentleyjones: Also, this:
>The cloned file dst shares its data blocks with the src file but has its own copy of attributes, extended attributes and ACLs which are identical to those of the named file src with the exceptions [of ownership information, setuid and setgid bits]

PR #1207, opened 2021-08-04
Move `-install_name` linker flag from the `{ios,tvos}_framework` macro into the rule implementation.

thii:
PiperOrigin-RevId: 382086173
(cherry picked from commit 172ad76549d732b3380bef6a7c255c32cc4462d3)

Conflicts:
	apple/internal/ios_rules.bzl
	apple/internal/tvos_rules.bzl
	apple/ios.bzl
	apple/tvos.bzl

### DIFF
--- a/apple/internal/ios_rules.bzl
+++ b/apple/internal/ios_rules.bzl
@@ -632,21 +632,6 @@ def _ios_app_clip_impl(ctx):
 
 def _ios_framework_impl(ctx):
     """Experimental implementation of ios_framework."""
-
-    # TODO(kaipi): Add support for packaging headers.
-    extra_linkopts = ["-dynamiclib"]
-    if ctx.attr.extension_safe:
-        extra_linkopts.append("-fapplication-extension")
-
-    link_result = linking_support.register_linking_action(
-        ctx,
-        avoid_deps = ctx.attr.frameworks,
-        extra_linkopts = extra_linkopts,
-        stamp = ctx.attr.stamp,
-    )
-    binary_artifact = link_result.binary
-    debug_outputs_provider = link_result.debug_outputs_provider
-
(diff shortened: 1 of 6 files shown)


keith (approved):

thii:
Depends on https://github.com/bazelbuild/rules_apple/pull/1188

apple/internal/ios_rules.bzl
```diff
@@ -680,6 +665,25 @@ def _ios_framework_impl(ctx):
         res_attrs = ["resources"],
     )
 
+    extra_linkopts = [
+        "-dynamiclib",
+        "-Wl,-install_name,@rpath/{name}{extension}/{name}".format(
```
brentleyjones, line 670: I used this version to work around https://github.com/bazelbuild/bazel/blob/1c3a2456c95fd19974a5b2bd33c5ebdb2b2277e4/src/main/java/com/google/devtools/build/lib/rules/cpp/LinkCommandLine.java#L295 causing an issue. Think this is bug worthy? One nice thing about doing it this way: it's the same argument that was being sent before.

thii: Thanks. Yes I think we should report it.

brentleyjones:
Hmm, it should have re-branched, not closed this...

brentleyjones:
@thii I'm in the middle of a rebase right now. Care to open a new one with the changes that exist already in your repo?

brentleyjones:
The last two commits just need to be rebased onto master now.

apple/internal/ios_rules.bzl
brentleyjones: https://github.com/bazelbuild/bazel/issues/14316

PR #1210, opened 2021-08-04
Allow tests to consume files under `doc`

thii:
So that it can reference to the patch file.

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -17,6 +17,7 @@ filegroup(
     srcs = [
         "WORKSPACE",
         "//apple:for_bazel_tests",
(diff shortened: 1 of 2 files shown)


thii:
git bisect showed that the failure with last green bazel is because of https://github.com/bazelbuild/bazel/commit/2b2d4229a60d41b250f9d2fab3983a681bfdbf2d

thii:
It was reverted in https://github.com/bazelbuild/bazel/commit/aa4191bcafe7a5bacdc646ec8cef1240f29fc9d0

keith (approved):

PR #1211, opened 2021-08-05
Allow `features` to propograte to bundle rules

brentleyjones:


brentleyjones:
This is needed for #958 to work.

keith (approved):

PR #1220, opened 2021-08-18
Add feature flag to disable legacy code signing. This will be used to enable rollout of new manifest driven code signing.

keith:
PiperOrigin-RevId: 384377651
(cherry picked from commit 0478507a6480ff5e06b4ca24514409f6522e43a2)


keith:
I'm not sure about the public applicability of the general feature here, but threading features through is probably nice long term

thii (approved):

PR #1223, opened 2021-08-20
Generate lcov file for tests runs with coverage

keith:


tinder-maxwellelliott (approved):

tinder-maxwellelliott:
Thanks for the assist on this. Learned a good deal reading this file

apple/testing/default_runner/ios_test_runner.template.sh
```diff
@@ -146,10 +158,62 @@ else
...
+fi
+
+xcrun llvm-cov \
+  export \
+  -format lcov \
+  -instr-profile "$profdata" \
+  -path-equivalence="$ROOT","$PWD" \
+  "$test_binary" \
+  @"$COVERAGE_MANIFEST" \
+  > "$COVERAGE_OUTPUT_FILE" \
```
tinder-maxwellelliott, line 209: Do you have docs on these variables?

keith: don't think there is any, some are in here https://github.com/bazelbuild/bazel/blob/master/tools/test/collect_coverage.sh but I just ran `env; exit 1` in this script to see what was available. it does seem to treat these special (we just throw things in the undeclared outputs)

apple/testing/default_runner/ios_test_runner.template.sh
```diff
@@ -146,10 +158,62 @@ else
...
+# llvm-cov export doesn't print any warnings/errors when files that it's trying
+# to reference don't exist. Unfortunately these are absolute paths that won't
+# be valid across machines. When using llvm-cov show it prints a warning for
+# this so we can validate we don't hit this case.
+error_file=$(mktemp)
+llvm_cov_status=0
+xcrun llvm-cov \
+  show \
+  -instr-profile "$profdata" \
+  -path-equivalence="$ROOT","$PWD" \
```
keith, line 189: this is the biggest part I'm unsure about, this isn't right for folks using the remote cache, they need to use coverage-prefix-map and remap `.` to `$PWD`

apple/testing/default_runner/ios_test_runner.template.sh
tinder-maxwellelliott: TIL just go into Bazel stuff and do `env; exit 1`

apple/testing/default_runner/ios_test_runner.template.sh
keith: I'll leave this and we can fix if something becomes more clear

keith:
I pushed macOS coverage support here, @tinder-maxwellelliott can you try it?

tinder-maxwellelliott:
> I pushed macOS coverage support here, @tinder-maxwellelliott can you try it?

Will take a look. Thanks again Keith

brentleyjones (approved):

brentleyjones:
@tinder-maxwellelliott Were you able to take a look?

PR #1229, opened 2021-08-31
Abstract entitlement matching value validation

keith:
There are many different entitlements that, when set, require they are
set to the same value in the entitlements file and the
provisioning profile. This extracts that use case into a single set that
we can add these keys to over time.

### DIFF
--- a/tools/plisttool/plisttool.py
+++ b/tools/plisttool/plisttool.py
@@ -287,14 +287,13 @@ ENTITLEMENTS_APS_ENVIRONMENT_MISMATCH = (
     'match the value in the provisioning profile ("%s").'
 )
 
-ENTITLEMENTS_WIFI_INFO_MISSING = (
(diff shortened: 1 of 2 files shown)


tools/plisttool/plisttool.py
```diff
@@ -336,6 +335,15 @@
     'bundle_id', 'profile_metadata_file', 'validation_mode',
 ])
 
+# Keys which should match in the profile and entitlements if they're expected
+_BOOLEAN_KEYS = frozenset([
+  'com.apple.developer.networking.wifi-info',
+  'com.apple.developer.passkit.pass-presentation-suppression',
+  'com.apple.developer.payment-pass-provisioning',
+  'com.apple.developer.siri',
+  'com.apple.developer.usernotifications.time-sensitive',
```
keith, line 344: Added a few here we weren't previously validating

brentleyjones (approved):

PR #1254, opened 2021-10-28
Update xctestrunner

keith:
https://github.com/google/xctestrunner/compare/5a1b0c158f2debace60722ddbcb5035e3387810e...a13ec9de45d225620dd4b5e30b23bdc663631f31

### DIFF
--- a/apple/repositories.bzl
+++ b/apple/repositories.bzl
@@ -142,9 +142,9 @@ def apple_rules_dependencies(ignore_version_differences = False):
         http_archive,
         name = "xctestrunner",
         urls = [
-            "https://github.com/google/xctestrunner/archive/5a1b0c158f2debace60722ddbcb5035e3387810e.tar.gz",
+            "https://github.com/google/xctestrunner/archive/7f8fc81b10c8d93f09f6fe38b2a3f37ba25336a6.tar.gz",
         ],
-        strip_prefix = "xctestrunner-5a1b0c158f2debace60722ddbcb5035e3387810e",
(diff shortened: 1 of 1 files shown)


thii:
It seems like this update is not compatible with RBE if tests are allowed to run on the remote executors.

```
exec ${PAGER:-/usr/bin/less} "$0" || exit 1
Executing tests from //sometest
-----------------------------------------------------------------------------
2021-10-29 23:01:13,665 Creating a new simulator:
Name: New-iPhone 13 Pro Max-15.0
OS: iOS 15.0
Type: iPhone 13 Pro Max
2021-10-29 23:01:14,639 Created new simulator 14753767-9861-450F-BF51-FF192687DF13.
2021-10-29 23:01:14,688 Will consider the test as test type logic_test to run.
An error was encountered processing the command (domain=com.apple.CoreSimulator.SimError, code=404):
Invalid device: The specified device is not available for spawning processes.
2021-10-29 23:01:18,106 Deleting simulator 14753767-9861-450F-BF51-FF192687DF13 asynchronously.
2021-10-29 23:01:18,112 Done.
```

keith:
huh is this from your internal testing? I guess we should report upstream. It is a large diff

thii:
All of our tests failed with this error after trying this update. Running with `--modify_execution_info=TestRunner=+no-remote-exec` worked. Reverting this change also worked.

thii:
Reported: https://github.com/google/xctestrunner/issues/38

Will try to debug more later.

keith:
Thanks. We could Also take the other commits without the biggest one if we wanted

thii:
The second update seems fixed the issue.

thii (approved):

PR #1260, opened 2021-11-03
Convert codesignopts to use expand_location

steeve:
`expand_make_variables` is deprecated, and the usage was to reference a label
anyway. `$(location)` is the proper tool for that, and simplifies the experience.


thii:
There was a discussion about this in the past, `expand_make_variables` is documented as deprecated but there's no replacement yet. Can we support both?

https://github.com/bazelbuild/bazel/issues/5859

steeve:
We could support both, although having expand_make_variables is rather clunky to use imho (dedicated rule + toolchain + codesign inputs).
Chances are most people are not even using it as is.

Or perhaps we could make `--resource_rules` a higher level abstraction? Since i'd wager that's how why most people would use it

steeve:
Also, `ctx.var` is the replacement and so it should be possible to use regular templating with it:
```
"hello {BINDIR}".format(ctx.var)
```

thii:
cc @michaeleisel since I think they were using it when added this `codesignopts` attribute.

brentleyjones:
Here is how we use it today:
```
codesignopts = select({
    "//:debug": [
        "--digest-algorithm=sha1",
        "--resource-rules=$(RULES)",
    ],
    "//conditions:default": [],
}),
codesign_inputs = select({
    "//:debug": ["@//bazel:resource_rules"],
    "//conditions:default": [],
}),
toolchains = select({
    "//:debug": ["@//bazel:resource_rules"],
    "//conditions:default": [],
})
```

test/starlark_tests/targets_under_test/ios/BUILD
```diff
@@ -1026,29 +1025,21 @@ ios_extension(
 
 # ---------------------------------------------------------------------------------------
 
-substitution(
-    name = "resource_rules",
-    src = "//test/starlark_tests/resources:ResourceRules.plist",
-    tags = FIXTURE_TAGS,
-    var_name = "RULES",
-)
```
keith, line 1034: we should def keep this test for now

steeve: which one? the test is still there, only that toolchain rule is removed (since it's not useful anymore)

steeve: or perhaps you're advising we keep both behaviours ?

keith: otherwise this is a breaking change right?

keith: which we could do ofc, but im not sure we want to?

steeve: it would be a breaking change indeed. knowing how difficult it was to implement, I figured those who were using it would welcome the change anyway (since it's so much simpler)

steeve:
Gentle ping on this one

brentleyjones:
While `ctx.expand_make_variables` is deprecated, it says to use [ctx.var](https://docs.bazel.build/versions/main/skylark/lib/ctx.html#var) instead. Are we able to continue to support toolchain expansion along with location expansion? Sure, the current use case shown is for resource rules, and `location` works for that, but since we've added toolchain support, people might have become dependent on it in another way.

steeve:
@brentleyjones it's possible to almost get the old behaviour back by switching from using `@()` to starlark formatting + `ctx.var`, such as: `"hello {BINDIR}".format(**ctx.var)`
I could double wrap `expand_make_variables` around `expand_location`, but I thought since it's deprecated, might as well bite the bullet.

PR #1262, opened 2021-11-17
Gather resources from swift’s private deps

naluhh:
Resources were not propagated if you depend on a module privately.


keith:
Makes sense! Can you add a test?

brentleyjones:
@naluhh Can you add a test to make sure we don't regress this fix?

keith:
Looks good! Can you add a test validating the resources get picked up as expected?

keith:
this change happened at some point. i don't think it's tested either

Issue #1215, opened 2021-08-17
Examples and/or documentation for running iOS tests on a physical device


jiawen:
I've been using `rules_apple` successfully for a few projects and can run `ios_unit_test` in a simulator. However, there doesn't seem to be much documentation regarding how to run tests on a physical device. A great use case would be to easily measure on-device performance.

It would be tremendously helpful to have at least some documentation, if not working examples under the `//examples` directory.

I'm imagining something along the lines of:
* How to run a single `ios_unit_test` and `ios_ui_test`.
* How to run `ios_unit_test_suite` and `ios_ui_test_suite`.
* Examples of different `runner`s and `test_host`s and what they mean.
    * What's required in a host `ios_application`.
* Tulsi example.
* How this all fits with [google/xctestrunner](https://github.com/google/xctestrunner).

keith:
I agree the general workflows here should be documented. But the tl;dr is that you need to use Tulsi (or some other Xcode bridge) to do anything on device. Bazel and the rules themselves have no knowledge of how to run on device. (xctestrunner does, but that isn't exposed through the rules today)

jiawen:
Thanks! The pointer about needing Tulsi is already helpful!

I also discovered today via browsing through the issues that it's possible to get `bazel run` to execute tests inside a simulator with automated device matching - that's awesome! We should put it in the examples (I'm an Xoogler and happy to contribute).

keith:
Great! We can definitely start some new document in docs/ with this kinda info

thii:
@jiawen Can you try this patch https://github.com/bazelbuild/rules_apple/pull/1266?

DoDoENT:
I did exactly the same and running test fails with

```
xctestrunner.shared.ios_errors.BundleError: Failed to extract signing identity from Executable
```

Running the very same test via Xcode generated by `rules_xcodeproj` works, so there is nothing wrong with my provisioning profile.

PR #1266, opened 2021-11-24
Support running tests on device

thii:
Testing on device can be done by passing
`--test_arg=--destination=platform=ios_device,id=YOUR_DEVICE_UDID`
to your `bazel test` command.

### DIFF
--- a/apple/testing/default_runner/ios_test_runner.template.sh
+++ b/apple/testing/default_runner/ios_test_runner.template.sh
@@ -25,12 +25,15 @@ basename_without_extension() {
 
 custom_xctestrunner_args=()
 command_line_args=()
-simulator_id=""
+device_id=""
+platform=""
 while [[ $# -gt 0 ]]; do
   arg="$1"
   case $arg in
-    --destination=platform=ios_simulator,id=*)
-      simulator_id="${arg##*=}"
+    --destination=platform=*,id=*)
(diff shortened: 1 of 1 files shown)


keith (approved):
nice, didn't know this would "just work"

brentleyjones (approved):

brentleyjones:
@thii Is this waiting for anything?

thii:
Still having some codesigning issue, seems like this is not enough.

jiawen:
> Still having some codesigning issue, seems like this is not enough.

I got this working. Had to code sign the `ios_unit_test` as well. More details in #1215.

keith:
So maybe we could require a profile in the case you want to run this way? 🤔

jiawen:
> So maybe we could require a profile in the case you want to run this way? 🤔

Makes sense to me. We should just document it - in `ios_unit_test` and `ios_ui_test` and ideally also in an example. As part of a tutorial, I'd have the test fail by default. Then as step 2, the user brings their own provisioning profile to make it pass.

It's hard to test but testing anything on device is hard :).

jiawen:
Anything preventing this PR from being merged? I assume the test failures are unrelated?

thii:
Thanks for verifying. Rebased and all tests passed now, so I'm merging this.

PR #1271, opened 2021-11-24
Added dSYM bundle and linkmap support for XCFrameworks.

brentleyjones:
PiperOrigin-RevId: 396385908
(cherry picked from commit e79562994e260143a26706fbe6f33542af111f72)

Also re-applied changes from commit d4c920a.


brentleyjones:
@thii This doesn't (fully) include the changes from https://github.com/bazelbuild/rules_apple/pull/828. Can you take a look at add to this PR what you think the solution should be?

thii:
Will look later today.

thii (approved):

apple/internal/partials/debug_symbols.bzl
```diff
@@ -201,26 +210,27 @@ def _bundle_dsym_files(
...
         for (arch, arch_outputs) in debug_outputs_provider.outputs_map.items():
-            outputs.append(_copy_dsym_into_declared_bundle(
-                actions = actions,
-                arch = arch,
-                debug_output_filename = debug_output_filename,
-                dsym_binary = arch_outputs["dsym_binary"],
-                dsym_bundle_name = dsym_bundle_name,
-                executable_name = executable_name,
-            ))
+            dsym_binaries[arch] = arch_outputs["dsym_binary"]
```
brentleyjones, line 215: @thii Since this is supposed to work for xcframeworks, what if it gets an arm64 simulator and an arm64 device? lipo won't work in that case.

brentleyjones: Ahh, I think this is called for each platform, so we should be good. I'll run some tests.

brentleyjones:
Thanks @thii!

apple/internal/partials/debug_symbols.bzl
```diff
@@ -202,6 +273,10 @@ def _debug_symbols_partial_impl(
         if _AppleDebugInfo in x
     ]
 
+    debug_output_filename = bundle_name
```
rsahara, line 276: @thii This line changes the dSYM filename when a custom executable name is used.
Shouldn't we use the `executable_name` instead of the `bundle_name`?
(According to your PR: https://github.com/bazelbuild/rules_apple/pull/934)

thii: Yes, it should be `executable_name` if it's not `None` (I think there's a helper from `bundling_support`). Are you able to submit a fix?

rsahara: I see. I'll test with a patch and file a PR if it works.

PR #1299, opened 2021-11-29
Add universal_apple_binary rule for multi-arch binaries

brentleyjones:
Introduces a new `universal_apple_binary` rule to support building
universal ('fat') binaries from other binary targets (e.g. `cc_binary`,
`go_binary`) using a combination of split transitions and Apple's `lipo`
tool for universal binaries.

PiperOrigin-RevId: 407859091
(cherry picked from commit 73c041db161f5058f2e4cc124b7591300618b5a6)

### DIFF
--- a/test/starlark_tests/rules/apple_verification_test.bzl
+++ b/test/starlark_tests/rules/apple_verification_test.bzl
@@ -38,9 +38,16 @@ load(
 
 def _apple_verification_transition_impl(settings, attr):
     """Implementation of the apple_verification_transition transition."""
+
+    # This was added because this transition is also used by
+    # `infoplist_contents_test` and has no "macos_cpus" attribute.
+    macos_cpus = "x86_64"
+    if hasattr(attr, "macos_cpus"):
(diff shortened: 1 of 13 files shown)


test/starlark_tests/rules/apple_verification_test.bzl
```diff
@@ -38,9 +38,16 @@ load(
 
 def _apple_verification_transition_impl(settings, attr):
     """Implementation of the apple_verification_transition transition."""
+
+    # This was added because this transition is also used by
+    # `infoplist_contents_test` and has no "macos_cpus" attribute.
+    macos_cpus = "x86_64"
```
keith, line 44: hrm prob wrong, but we can fix later

keith (approved):

test/starlark_tests/rules/apple_verification_test.bzl
brentleyjones: Ahh, probably. Let's fix it after if it is.

PR #1300, opened 2021-12-01
Skip signing of frameworks in tests as well

brentleyjones:
When not signing the simulator bundle, we don't need to sign the frameworks either.

### DIFF
--- a/apple/internal/codesigning_support.bzl
+++ b/apple/internal/codesigning_support.bzl
@@ -415,6 +415,14 @@ def _codesigning_command(
     if not should_sign_bundles:
         return ""
 
+    should_sign_sim_bundles = _should_sign_simulator_bundles(
+        config_vars = platform_prerequisites.config_vars,
+        features = platform_prerequisites.features,
+        rule_descriptor = rule_descriptor,
+    )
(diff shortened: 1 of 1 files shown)


apple/internal/codesigning_support.bzl
```diff
@@ -450,7 +450,9 @@ def _codesigning_command(
         features = platform_prerequisites.features,
         rule_descriptor = rule_descriptor,
     )
-    if platform_prerequisites.platform.is_device or should_sign_sim_bundles:
+    if not should_sign_sim_bundles:
+        return ""
```
keith, line 454: this same condition is on 415, not sure if we are trying to rely on some of the side effects in between?

brentleyjones: 415 is `should_sign_bundles` but this is checking `should_sign_sim_bundles`. But yes, it doesn't need any of the code between, so I should move `should_sign_sim_bundles` calculation up and make it an `or` there.

brentleyjones:
I just realized I got everything that should go with this in https://github.com/bazelbuild/rules_apple/pull/958/commits/456431d336c9a1ded5f0cc568b319bc7de74801a.

Issue #1253, opened 2021-10-28
Support backdeployed concurrency libraries


keith:
We need to do some work here with Xcode 13.2+ to support copying these based on the minimum supported OS version

keith:
Here's the first step of this https://github.com/apple/swift/pull/39996

keith:
`swift-stdlib-tool` is smart enough to only copy the libraries that need to be copied based on the executables we process. My current thinking is that we can safely update these versions:

https://github.com/bazelbuild/rules_apple/blob/dc63441f25ffc60d04160fdf391b10e9c8d78cb2/apple/internal/partials/swift_dylibs.bzl#L68-L76

(with a few other changes based on the PR above) to re-use the same action for this case. Theoretically we could split these up but the same logic needs to be applied, so I don't see a benefit in doing that.

keith:
Here's a slightly related change https://github.com/bazelbuild/rules_apple/pull/1258 since with the PR above we either needed to update the `--source-libraries` arguments or remove them

keith:
https://github.com/bazelbuild/rules_apple/pull/1301 here's what the min version bump PR would look like to support this, probably still going to wait for the Xcode 13.2 release to merge anything for it

keith:
As far as I know we've covered everything here, but do note with Xcode 13.2.x there is a bug where even if you annotate your APIs with the correct versions, your app will crash on iOS 12.x *devices only* no matter what you do https://forums.swift.org/t/swift-concurrency-back-deploy-issue/53917 so be careful out there!

Please file any issues you find!

PR #1301, opened 2021-12-03
Update min version for Swift dylib copying

keith:
To support the new back deployed concurrency libraries we need to run
swift-stdlib-tool for newer versions than before. This bumps the minimum
version until we can stop running this action to reflect this.
swift-stdlib-tool then does the right thing to only copy the concurrency
library, in the case you're deploying to a version newer than 12.2.

Part of https://github.com/bazelbuild/rules_apple/issues/1253

### DIFF
--- a/apple/internal/partials/swift_dylibs.bzl
+++ b/apple/internal/partials/swift_dylibs.bzl
@@ -61,14 +61,21 @@ File object that represents a directory containing the Swift dylibs to package f
     },
 )
 
-# Minimum OS versions after which the Swift StdLib dylibs are packaged with the OS. If the minimum
-# OS version for the current target and platform is equal or above to the versions defined here,
-# then we can skip copying the Swift dylibs into Frameworks and SwiftSupport.
+# Minimum OS versions for which we no longer need to potentially bundle any
+# Swift dylibs with the application. The first cutoff point was when the
+# platforms bundled the standard libraries, the second was when they started
+# bundling the Concurrency library. There may be future libraries that require
+# us to continue bumping these values. The tool is smart enough only to bundle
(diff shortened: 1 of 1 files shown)


apple/internal/partials/swift_dylibs.bzl
```diff
@@ -65,7 +65,7 @@ File object that represents a directory containing the Swift dylibs to package f
 # OS version for the current target and platform is equal or above to the versions defined here,
 # then we can skip copying the Swift dylibs into Frameworks and SwiftSupport.
 _MIN_OS_PLATFORM_SWIFT_PRESENCE = {
-    "ios": apple_common.dotted_version("12.2"),
+    "ios": apple_common.dotted_version("15.2"),
     "macos": apple_common.dotted_version("10.14.4"),
```
keith, line 69: need to bump these 3 as well

apple/internal/partials/swift_dylibs.bzl
```diff
@@ -65,7 +65,7 @@ File object that represents a directory containing the Swift dylibs to package f
 # OS version for the current target and platform is equal or above to the versions defined here,
 # then we can skip copying the Swift dylibs into Frameworks and SwiftSupport.
 _MIN_OS_PLATFORM_SWIFT_PRESENCE = {
-    "ios": apple_common.dotted_version("12.2"),
+    "ios": apple_common.dotted_version("15.2"),
```
keith, line 68: Need to check if it's 15.0 or 15.2, i think it's 15.0 but there were some issues so I'm not sure how that plays into this

keith:
I think this is good now that Xcode 13.2 has been released

brentleyjones (approved):

PR #1302, opened 2021-12-03
[HACK] Example diff for testing swift-stdlib-tool builds locally

keith:
This is more complicated than just changing the binary path because the
way we use it it expects relative paths inside of Xcode, so we have to
duplicate those details as if it is there.


keith:
never going to merge this but it's a useful example for ironing out upstream regressions

thii:
Can it be built with Bazel? If yes, we can fetch the source and build it in the rules.

keith:
huh interesting. Yes it could be, it's 1 c++ file with no deps (I think)

thii:
https://github.com/bazelbuild/rules_apple/pull/1314

PR #1308, opened 2021-12-09
Fix entitlement validation with appattest

keith:
f0404bc5e593f77d84820ad35fd016fc6a82eb24 introduced a regression where
the value of this entitlement was in the potential list, so it fell
through to the next conditional which validated they were equal, which
will never be the case. This also adds a test for this successful case.


brentleyjones (approved):

tools/plisttool/plisttool.py
```diff
@@ -1215,10 +1225,6 @@ def _validate_entitlements_against_profile(self, entitlements):
         'keychain-access-groups', self.target,
         supports_wildcards=True)
 
-    self._check_entitlements_array(
```
keith, line 1218: this was duplicated

keith:
FYI @kaylathar

Issue #1307, opened 2021-12-09
Slow bundling with a lot of resources caused by expensive copying action


BalestraPatrick:
Hello!

After a conversation with @brentleyjones I debugged a bit why our incremental builds can spend up to 2 minutes in the `BundleTreeApp` action. Our app has over 16K files that need to be copied, and the part that seems the slowest is the following:
https://github.com/bazelbuild/rules_apple/blob/b5fe128892c5193d664ef8e3637883ed47e354b6/tools/bundletool/bundletool_experimental.py#L203-L207

This was recently changed as an optimisation in https://github.com/bazelbuild/rules_apple/pull/1202 to reduce disk usage. Unfortunately for it, the trade-off here is disk-usage vs speed. During incremental builds, we might want to give up on reducing disk-usage by making the build faster. When compiling hundreds of test targets that share resources on CI instead, we might choose to prefer a reduce space taken up by these bundles.

In my specific case, I tried reverting the usage of `cp -c` with `shutil.copy(src, full_dest)` (which is currently used on non-Darwin) platforms and we can reduce the time it takes for this action to run from 49s to 17s in our specific case.

I would suggest making this behaviour optional, so that a build can decide what to favour: either disk-usage or speed.

What do people think? I'm open to suggestions.

brentleyjones:
The problem seems to be that the call out to a subprocess is much more expensive than the native binding that Python has for `shutil.copy`.

Longer term maybe the bundler should be replaced with a compiled binary that can efficiently invoke CloneFile. In the meantime though, I agree, this should be optional, or even simply reverted.

keith:
Is popen or any other related API faster?

dflems:
@brentleyjones  I suppose using `cp` takes better advantage of APFS cloning and doesn't actually create duplicate files? Since we're using python here, maybe good ol' FFI might do the trick?

```python
from ctypes.util import find_library
from ctypes import cdll, c_char_p, c_int

# load libSystem
system = cdll.LoadLibrary(find_library('libSystem'))

# set types for clonefile function
clonefile = system.clonefile
clonefile.argtypes = [c_char_p, c_char_p, c_int] # src, dest, flags
clonefile.restype = c_int  # 0 on success

# invoke
clonefile(b'foo.txt', b'bar.txt', 0)
```

brentleyjones:
> I suppose using `cp` takes better advantage of APFS cloning and doesn't actually create duplicate files?

[...]

PR #1311, opened 2021-12-13
Use clonefile directly on macOS

keith:
Fixes https://github.com/bazelbuild/rules_apple/issues/1307

For our largest app this improves tree bundling time from 27s to 10s


BalestraPatrick:
I was testing this right now and it makes the specific copy action go from ~36s to 4s now for our app 🥳

PR #1166, opened 2021-05-24
basic support for importing .xcframeworks

zachgray-slack:
A very simple implementation of `.xcframework` import support which relies on the user specifying the `.framework` path for each platform.

I've been working on a draft of a more automated approach which reads the `Info.plist` and outputs a `TreeArtifact` containing the "selected" .framework based on the current build, but there's significant challenges here as the internals of this ruleset really are expecting `File` handles to the inputs which are lost when using the "generated" (selected) .framework. 

In the meantime, something like this initial implementation might at least be a good starting point worth consideration as it's much simpler, albeit less convenient. 

Usage example:

```
apple_dynamic_framework_import(
    name = "AmazonChimeSDK",
    framework_imports = glob(["AmazonChimeSDK.xcframework/**"]),
    visibility = ["//visibility:public"],
    xcframework_platform_paths = {
        "iphoneos": "ios-arm64_armv7/AmazonChimeSDK.framework",
        "iphonesimulator": "ios-i386_x86_64-simulator/AmazonChimeSDK.framework",
    },
)
```

This initial work is inspired by [this patch](https://github.com/axl411/rules_apple/commit/184efe6fbbc38535632b2a212b864871c62396cb?branch=184efe6fbbc38535632b2a212b864871c62396cb&diff=split) which is similar to what some folks in the community are doing now. 

Note: I based this work on ed2bceef7ac5a3071b023e3122a045a133d2245c as this is the version pinned in `rules_ios` at time of writing, but of course this can be easily updated if there's any interest in landing this.

related issues: #851 #1091

Issue #851, opened 2020-07-24
Consuming .xcframework


sgammon:
Hello `rules_apple` authors,

I am trying to use a `.xcframework` in an iOS app project (specifically, Firebase's static library, which contains `.xcframework`s for various components).

Is there a supported way to do this? I had tried via `apple_static_framework_import`, but found it rejecting anything not ending in `.framework`. So, examining the structure of an `.xcframework`, I tried:

```starlark
def _apple_xcframework(name, path, **kwargs):

    """ Consume an Apple XCFramework. """

    apple_static_framework_import(
        name = name,
        framework_imports = select({
            # x86: simulator or target
            "//defs/conditions:dev": native.glob([
                 "%s/ios-i386_x86_64-simulator/**" % path,
            ]),

            # ARM: prod or debug
            "//defs/conditions:release": native.glob([
                 "%s/ios-armv7_arm64/**" % path,
            ]),
        }),
        **kwargs
    )
```

i am using this like so:
```starlark
apple_xcframework(
    name = "FirebaseCore-XCFramework",
    path = "Firebase/FirebaseAnalytics/FirebaseCore.xcframework",
)
```

which roughly translates to:
```starlark
apple_static_framework_import(
    name = "FirebaseCore-XCFramework",
    framework_imports = select({
        "//defs/conditions:dev": native.glob([
             "Firebase/FirebaseAnalytics/FirebaseCore.xcframework/ios-i386_x86_64-simulator/**",
        ]),

        "//defs/conditions:release": native.glob([
             "Firebase/FirebaseAnalytics/FirebaseCore.xcframework/ios-armv7_arm64/**",
        ]),
    }),
)
```

is this the right way to consume `.xcframework`s?

keith:
FWIW up to this point there hasn't been any work AFAIK on supporting xcframeworks. I would love to hear if others are working on this.

allevato:
There's some work currently in progress to develop a rule to *produce* an .xcframework bundle (with the goal of replacing `ios_static_framework` and also providing a way to create distributable dynamic frameworks, since that use case isn't served by `ios_framework`). No timeline on this though.

[...]

PR #1312, opened 2021-12-14
basic support for importing .xcframeworks

xiemotongye:
This pull request is related to https://github.com/bazelbuild/rules_apple/pull/1166.

The original pull request did a great job, but it seems that there is no follow-up for a long time.

Now I'm facing the same thing, I hope rules_apple can officially support for the xcframwork import. So I did a little work on it.

Usage example:

```
apple_dynamic_framework_import(
    name = "AmazonChimeSDK",
    framework_imports = glob(["AmazonChimeSDK.xcframework/**"]),
    xcframework_platform_ids = {
        "iphoneos": "ios-arm64_armv7",
        "iphonesimulator": "ios-i386_x86_64-simulator",
    },
)
```

### DIFF
--- a/apple/internal/apple_framework_import.bzl
+++ b/apple/internal/apple_framework_import.bzl
@@ -507,7 +507,7 @@ platforms directory, without the extension (for example, iphoneos or iPhoneSimul
 should be case-sensitive variants of values that might be found in the LibraryIdentifier of an Info.plist 
 file in xcframework's root(for example, ios-arm64_i386_x86_64-simulator or ios-arm64_armv7).
 """,
-        ),
+    ),
 }
 
 apple_dynamic_framework_import = rule(
--- a/apple/internal/apple_framework_import.bzl
+++ b/apple/internal/apple_framework_import.bzl
@@ -318,11 +318,36 @@ def _debug_info_binaries(
 
     return all_binaries_dict.values()
 
+def _get_framework_imports(ctx):
+    if not ctx.attr.xcframework_platform_ids:
+        return ctx.files.framework_imports
+    curr_platform = ctx.fragments.apple.single_arch_platform.name_in_plist.lower()
+    xcframework_path = ""
+    for f in ctx.files.framework_imports:
+        if ".xcframework" in f.path:
+            xcframework_path = f.path.split(".xcframework")[0] + ".xcframework"
+            break
+    if not xcframework_path.endswith(".xcframework"):
+        fail("couldn't find xcframework at framework_imports")
+    framework_name = paths.split_extension(paths.basename(xcframework_path))[0]
+
+    if curr_platform not in [p.lower() for p in ctx.attr.xcframework_platform_ids]:
+        fail(
+            "Missing framework path mapping for platform `{}`; is this platform supported?"
+                .format(str(curr_platform)),
+        )
+    else:
+        platform_path = "{}/{}/{}.framework".format(xcframework_path, ctx.attr.xcframework_platform_ids[curr_platform], framework_name)
+        framework_imports_for_platform = [f for f in ctx.files.framework_imports if platform_path in f.short_path]
+        if len(framework_imports_for_platform) == 0:
+            fail("couldn't find framework at path `{}`".format(platform_path))
+        return framework_imports_for_platform
+
 def _apple_dynamic_framework_import_impl(ctx):
     """Implementation for the apple_dynamic_framework_import rule."""
     providers = []
 
-    framework_imports = ctx.files.framework_imports
+    framework_imports = _get_framework_imports(ctx)
     bundling_imports, header_imports, module_map_imports = (
         _classify_framework_imports(ctx.var, framework_imports)
     )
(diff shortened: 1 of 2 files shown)


thii (approved):
It would be nice if you can add a test.

apple/internal/apple_framework_import.bzl
```diff
@@ -318,11 +318,36 @@ def _debug_info_binaries(
 
     return all_binaries_dict.values()
 
+def _get_framework_imports(ctx):
+    if not ctx.attr.xcframework_platform_ids:
+        return ctx.files.framework_imports
+    curr_platform = ctx.fragments.apple.single_arch_platform.name_in_plist.lower()
```
brentleyjones, line 324: @thii Because of the known bug with `ctx.fragments.apple.single_arch_platform`, we should patch this similar to https://github.com/bazelbuild/rules_swift/pull/694, after this is merged. Then we can revert both patches when Bazel core is fixed.

apple/internal/apple_framework_import.bzl
```diff
@@ -472,10 +497,23 @@ def _apple_static_framework_import_impl(ctx):
 
     return providers
 
+_common_attrs = {
+    "xcframework_platform_ids": attr.string_dict(
```
brentleyjones, line 501: I guess I haven't thought through this enough, but is this something that needs to be specified, versus Bazel figuring out itself?

thii: Because Bazel cannot read files at this phase https://github.com/bazelbuild/rules_apple/issues/851#issuecomment-663770260

We could move everything to the tool so that the detection would happen in the execution phase, but that would add more complexity and the imported files would be opaque to Bazel.

All other rules here also just defer the `Info.plist` reading to the rules consumers https://github.com/bazelbuild/rules_apple/blob/54e906af5a45ecd6863bc73d60c7a7419d01bea7/doc/common_info.md#variable-substitution

apple/internal/apple_framework_import.bzl
thii: The bug is with `single_arch_cpu`, not `single_arch_platform`.

keith:
note this approach doesn't solve consuming static xcframeworks that are themselves produced by rules_apple

apple/internal/apple_framework_import.bzl
zachgrayio: Yeah FWIW i did prototype doing all this automatically, and hit some big blockers like @thii mentioned. The internals need the `File` handles and if you do it more dynamically you're left with just the opaque `TreeArtifact` and it breaks most of the functionality. Still have that commit around if anyone wants to see it again--not super clean but you can see what I was working towards. https://github.com/zachgray-slack/rules_apple/commit/201c0577eaab79e65347164bcb8d1f571d8d3115

zachgrayio:
@keith now rules_apple is producing xcframeworks im happy to TAL at this again and find a solution that works for consuming these as well. I'm actually looking at this stuff again as it turns out, maybe we can get something landed this time lol

zachgrayio (changes requested):
so at a high level, this LGTM as it's an improvement on my work, but i guess there's 2 blockers:

1) needs a test probably, something I didn't get to either lol
2) if this patch doesnt support the xcframeworks this ruleset itself creates, we should definitely address that. When I created my draft, this stuff wasn't landed at all yet, so I didn't factor it in. if @keith can share more info here maybe we can figure out what changes we'd need to support that (or i can run a few quick experiments).

FWIW this same patch is applied manually in a few orgs and definitely in use in production, so it does work pretty well for the majority of cases...however we can assume it will quickly become an issue with upcoming Google releases of .xcframeworks that this won't support 😆

apple/internal/apple_framework_import.bzl
```diff
@@ -318,11 +318,36 @@ def _debug_info_binaries(
...
+    framework_name = paths.split_extension(paths.basename(xcframework_path))[0]
+
+    if curr_platform not in [p.lower() for p in ctx.attr.xcframework_platform_ids]:
+        fail(
+            "Missing framework path mapping for platform `{}`; is this platform supported?"
+                .format(str(curr_platform)),
+        )
+    else:
+        platform_path = "{}/{}/{}.framework".format(xcframework_path, ctx.attr.xcframework_platform_ids[curr_platform], framework_name)
+        framework_imports_for_platform = [f for f in ctx.files.framework_imports if platform_path in f.short_path]
```
erikkerber, line 341: Might want `path` here over `short_path`, which will return a relative path when the framework imports are in a different repo.

keith:
The issue I saw is that when you produce a static binary xcframework it doesn't use frameworks, it uses static libs directly, so this doesn't work in general for that. It also makes me wonder if these rules should be used at all for importing xcframeworks, or if there should be separate rules for that. It sounds like static framework xcframework support is spotty in other places, which is why it wasn't implemented in the bazel rule, I haven't tested it but I guess swiftpm doesn't support that case.

thii:
We could add support for that too. I think xcframeworks with static libs aren't that common so no one has asked for it.

keith:
Yea you just don't have another option if you want to support the new sim and statically link things in general. But that made me wonder if these rules weren't the right ones to bake this in to

thii:
@xiemotongye Do you mind if I finish this?

thii:
@xiemotongye Pinging again - if you see this, can you update your PR to allow maintainers to push to your branch? Thanks.

thii:
Submitted https://github.com/bazelbuild/rules_apple/pull/1344 that also supports for static libs, and doesn't need to pass platform ids.

xiemotongye:
> @xiemotongye Do you mind if I finish this?

Sorry for reply so late, I was just returning from vacation of Chinese New Year. And I'm very pleasure that you've finished this, buddy!

PR #1314, opened 2021-12-16
Build `swift_stdlib_tool` from source

thii:
This makes it easier to iterate on this, as Xcode 13.2 still ships with a broken version.


keith:
Note the version with Xcode 13.2 should not be broken. There is an issue with the unsigned directory but we don't use that. Did you find another issue?

thii:
ah you're right we don't do codesigning here

PR #1322, opened 2022-01-19
Allow re-use of iOS simulator in tests

keith:
Previously the test runner allowed you to pass a specific simulator UDID
to run on that, but since you passed that via --test_arg, that UDID
would invalidate any previous test runs on a different simulator UDID.
With this change you can pass `--test_env=REUSE_GLOBAL_SIMULATOR=yes` in
order to either create and boot, or reuse, a simulator with a specific
device / OS type. The potential downsides of this are:

- You could leave around state on your simulator that invalidates future
  test runs, either by causing invalid failures, or hiding real issues
- You have to cleanup the simulator yourself if you need to on CI,
  running `xcrun simctl delete all` is likely enough

For these tradeoffs you run subsequent tests much faster than if you
have to create and boot a new simulator for every test bundle.

### DIFF
--- a/apple/testing/default_runner/ios_test_runner.bzl
+++ b/apple/testing/default_runner/ios_test_runner.bzl
@@ -27,6 +27,7 @@ def _get_template_substitutions(ctx):
         "device_type": device_type,
         "os_version": str(os_version),
         "testrunner_binary": ctx.executable._testrunner.short_path,
+        "simulator_creator": ctx.executable._simulator_creator.short_path,
     }
     return {"%(" + k + ")s": subs[k] for k in subs}
 
@@ -54,7 +55,9 @@ def _ios_test_runner_impl(ctx):
             test_environment = ctx.attr.test_environment,
         ),
         DefaultInfo(
-            runfiles = ctx.attr._testrunner[DefaultInfo].default_runfiles,
+            runfiles = ctx.runfiles(
+                files = [ctx.file._simulator_creator],
+            ).merge(ctx.attr._testrunner[DefaultInfo].default_runfiles),
         ),
     ]
 
@@ -107,6 +110,14 @@ into the XCTest invocation.
 It is the rule that needs to provide the AppleTestRunnerInfo provider. This
 dependency is the test runner binary.
 """,
+        ),
+        "_simulator_creator": attr.label(
+            default = Label(
+                "@build_bazel_rules_apple//apple/testing/default_runner:simulator_creator.py",
+            ),
+            allow_single_file = True,
+            executable = True,
+            cfg = "exec",
         ),
         "_xcode_config": attr.label(
             default = configuration_field(
--- a/apple/testing/default_runner/ios_test_runner.bzl
+++ b/apple/testing/default_runner/ios_test_runner.bzl
@@ -55,9 +55,8 @@ def _ios_test_runner_impl(ctx):
             test_environment = ctx.attr.test_environment,
         ),
         DefaultInfo(
-            runfiles = ctx.runfiles(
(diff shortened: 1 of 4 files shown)


jschear (commented):
This also assumes that you're making tests run serially via `--local_test_jobs=1`, right? (Probably worth mention that in some documentation if this lands.)

apple/testing/default_runner/ios_test_runner.bzl
```diff
@@ -108,6 +111,14 @@ It is the rule that needs to provide the AppleTestRunnerInfo provider. This
 dependency is the test runner binary.
 """,
         ),
+        "_simulator_creator": attr.label(
+            default = Label(
+                "@build_bazel_rules_apple//apple/testing/default_runner:simulator_creator.py",
```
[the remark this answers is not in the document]
keith: yea probably

keith:
Technically it's not required, I've heard some folks have run some tests on the same sims at the same time, but that sounds risky to me, so yea I agree it should be recommended

mattrobmattrob:
I was wondering less on the concern of "two test on the same sim" and more contention on the creation of a sim with the same name. Could there be issues there? We create our persistent sim out of Bazel process so don't have to be concerned with parallel invocation of the sim creation by Bazel in the `test --local_test_jobs=N where N > 1` case.

keith:
Yep good point, this doesn't block against that, maybe we could update the script to "lock" when creating a simulator

apple/testing/default_runner/simulator_creator.py
```diff
@@ -0,0 +1,90 @@
+#!/usr/bin/python3
```
thii, line 1: This file is missing a license header.

acecilia:
Adding here for visibility: the way I worked around this issue is by creating a simulator with a custom ID that I can choose. See the script here: https://gist.github.com/acecilia/a5ff5566c785c0f9e17490fe034d2070

apple/testing/default_runner/simulator_creator.py
```diff
@@ -0,0 +1,90 @@
...
+            ["create", device_name, device_type, runtime_identifier]
+        ).strip()
+        _boot_simulator(simulator_id)
+
+    print(simulator_id.strip())
+
+
+if __name__ == "__main__":
+    args = _build_parser().parse_args()
+    _main(args.os_version, args.device_type, args.name)
```
acecilia, line 90: Would it make sense for this code to be part of the test runner binary in https://github.com/google/xctestrunner? (Of course, another problem is the current not-so-great maintenance state of such runner 😞 )

acecilia: Or maybe this `simulator_creator.py` is a good starting point to eventually replace the test runner binary all together

keith: it definitely could be, I don't see a huge advantage in that at this point though, so yea the ownership differences makes this a bit easier

apple/testing/default_runner/simulator_creator.py
keith: hrm do we really need that? some BUILD files don't have it for example

thii: I don't know but BUILD files don't seem to need license header, from what I've seen in most Bazel repos (maybe they're not consider source code?)

thii (approved):

apple/testing/default_runner/simulator_creator.py
keith: added it

PR #1338, opened 2022-01-28
Ignore stderr from commands

keith:
On m1 machines xcrun and xcodebuild prints warnings like these:

```
objc[11792]: Class AMSupportURLConnectionDelegate is implemented in both /usr/lib/libauthinstall.dylib (0x207db1160) and /System/Library/PrivateFrameworks/MobileDevice.framework/Versions/A/MobileDevice (0x1186102b8). One of the two will be used. Which one is undefined.
```

I reported this 4/21 as FB9089778, but in the meantime we can just
ignore it.

### DIFF
--- a/tools/environment_plist/environment_plist.sh
+++ b/tools/environment_plist/environment_plist.sh
@@ -47,7 +47,7 @@ shift
 done
 
 set +e
-PLATFORM_DIR=$(/usr/bin/xcrun --sdk "${PLATFORM}" --show-sdk-platform-path)
+PLATFORM_DIR=$(/usr/bin/xcrun --sdk "${PLATFORM}" --show-sdk-platform-path 2>/dev/null)
 XCRUN_EXITCODE=$?
 set -e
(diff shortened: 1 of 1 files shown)


thii (approved):
LGTM. But this would hide unknown errors too. Can we only filter the warning?

keith:
if the commands actually fail they will still exit here, the xcrun case is also meant to handle this and I'm not silencing the output there:

```
if [[ ${XCRUN_EXITCODE} -ne 0 ]] ; then
  echo "environment_plist: SDK not located. This may indicate that the xcode \
and SDK version pair is not available."
  # Since this already failed, assume this is going to fail again. With
  # set -e, this will produce the appropriate stderr and error code.
  /usr/bin/xcrun --sdk "${PLATFORM}" --show-sdk-platform-path 2>&1
fi```

brentleyjones (approved):

thii (approved):

PR #1343, opened 2022-01-31
[HACK] Add Xcode 13.3 beta 1 swift-stdlib-tool workaround

keith:
This tool is broken in this Xcode beta, this can be used for testing
until a fix comes hopefully in a new beta, otherwise we'll have to
consider fixing this up and merging.


keith:
https://github.com/bazelbuild/rules_apple/pull/1352

PR #1346, opened 2022-02-02
Update skylib

keith:
This is the newest release, and it has the custom commit we were pinned
to. This also adds the bazel mirror as a fallback

### DIFF
--- a/apple/repositories.bzl
+++ b/apple/repositories.bzl
@@ -100,10 +100,10 @@ def apple_rules_dependencies(ignore_version_differences = False):
         http_archive,
         name = "bazel_skylib",
         urls = [
-            "https://github.com/bazelbuild/bazel-skylib/archive/df3c9e2735f02a7fe8cd80db4db00fec8e13d25f.tar.gz",
+            "https://github.com/bazelbuild/bazel-skylib/releases/download/1.1.1/bazel-skylib-1.1.1.tar.gz",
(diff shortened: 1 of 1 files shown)


apple/repositories.bzl
```diff
@@ -100,10 +100,10 @@ def apple_rules_dependencies(ignore_version_differences = False):
         http_archive,
         name = "bazel_skylib",
         urls = [
-            "https://github.com/bazelbuild/bazel-skylib/archive/df3c9e2735f02a7fe8cd80db4db00fec8e13d25f.tar.gz",
+            "https://github.com/bazelbuild/bazel-skylib/releases/download/1.1.1/bazel-skylib-1.1.1.tar.gz",
+            "https://mirror.bazel.build/github.com/bazelbuild/bazel-skylib/releases/download/1.1.1/bazel-skylib-1.1.1.tar.gz",
```
thii, line 104: I think Bazel will try to fetch each one by this order so maybe we should put the mirror one first.

```suggestion
            "https://mirror.bazel.build/github.com/bazelbuild/bazel-skylib/releases/download/1.1.1/bazel-skylib-1.1.1.tar.gz",
            "https://github.com/bazelbuild/bazel-skylib/releases/download/1.1.1/bazel-skylib-1.1.1.tar.gz",
```

thii (approved):

apple/repositories.bzl
keith: We've found the mirror to be much flakier than github in general, we actually removed it internally from everything because of that, which is why i left it second

PR #1352, opened 2022-02-03
Workaround swift-stdlib-tool breakage in Xcode 13.3 beta 1.

keith:
In 13.3b1, autodiscovery of the Swift compatibility libraries is broken. Search for possible `swift-*` directories and pass them explicitly with `--source-libraries` as a temporary fix; this is also forward-compatible until we're confident swift-stdlib-tool works correctly again.

PiperOrigin-RevId: 426225192
(cherry picked from commit ee2a905fca87c2f1b085e8987655c32dd5f50618)

### DIFF
--- a/tools/swift_stdlib_tool/swift_stdlib_tool.py
+++ b/tools/swift_stdlib_tool/swift_stdlib_tool.py
@@ -14,6 +14,7 @@
 #
 
 import argparse
+import glob
 import os
 import shutil
 import sys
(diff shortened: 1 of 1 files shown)


keith:
This is just https://github.com/bazelbuild/rules_apple/pull/1352 but I agree we can land it for now

keith:
This isj ust

keith:
This is https://github.com/bazelbuild/rules_apple/pull/1343 but I agree we can land it for now

keith:
the glob here makes it more backwards compat as well

tools/swift_stdlib_tool/swift_stdlib_tool.py
```diff
@@ -28,10 +29,21 @@ def _copy_swift_stdlibs(binaries_to_scan, sdk_platform, destination_path):
   """Copies the Swift stdlibs required by the binaries to the destination."""
   # Rely on the swift-stdlib-tool to determine the subset of Swift stdlibs that
   # these binaries require.
+  _, stdout, _ = execute.execute_and_filter_output(
+      ["xcode-select", "--print-path"], raise_on_failure=True)
+
+  developer_dir = stdout.strip()
```
thii, line 35: Should we get this from the `DEVELOPER_DIR` environment? That would save us one `xcode-select` invocation.

```suggestion
  developer_dir = os.environ["DEVELOPER_DIR"]
```

brentleyjones (approved):

thii (approved):

tools/swift_stdlib_tool/swift_stdlib_tool.py
keith: oops sorry, race. Is that one always going to match the format we need here? I've noticed that xcrun happily accepts the .app dir, or the .app/Contents/Developer dir, at the very least calling that above + this var being set, will normalize that

thii: ah I just realized this is a cherry pick, nevermind.

(`DEVELOPER_DIR` is the one injected by Bazel, so it should be the .app/Contents/Developer.)

keith: but if you launch bazel with that set to one without it, does it rewrite it?

thii: Hmm, I haven't tried that 🤔. I see that wrapped_clang and swift worker use it now, seems like bazel does rewrite it.

keith: yea looks like it gets normalized https://github.com/bazelbuild/rules_apple/pull/1356

PR #1353, opened 2022-02-04
Add support for including headers when importing XCFrameworks assembled with static libraries

thii:
For example, with the following XCFramework:

```
ios_static_xcframework.xcframework
├── Info.plist
├── ios-arm64
│   ├── Headers
│   │   └── shared.h
│   └── ios_static_xcframework.apple_static_library_lipo.a
└── ios-arm64_x86_64-simulator
    ├── Headers
    │   └── shared.h
    └── ios_static_xcframework.apple_static_library_lipo.a
```

Specifying `includes = ["Headers"]` will pass `-Ipath/to/ios_static_xcframework.xcframework/ios-arm64_x86_64-simulator/Headers` to the `{cc,objc}_library` that depends on this, when building for the iOS Simulator platform.


keith (approved):

brentleyjones (approved):

apple/internal/apple_framework_import.bzl
```diff
@@ -616,8 +617,16 @@ def _common_static_framework_import_impl(ctx, is_xcframework):
     providers.append(
         _objc_provider_with_dependencies(ctx, objc_provider_fields, additional_objc_infos),
     )
+
+    includes = []
+    if is_xcframework and not is_framework and ctx.attr.includes:
```
rockwotj, line 622: Why is there a `not is_framework` here?

keith: Note that this has changed slightly at HEAD

rockwotj: It looks to be broken at HEAD? It doesn't look like there is a prefix added but I'm not exactly sure where those are supposed to be relative too: https://cs.github.com/bazelbuild/rules_apple/blob/76752fe84ff0dc5d793033015647c78efbf41309/apple/internal/apple_xcframework_import.bzl#L578

keith: do you have a simple case I can test with? definitely possible something regressed

Issue #1358, opened 2022-02-09
LibreSSL in MacOS 12.3 breaks codesigningtool's identity extraction.


tnek:
As called out in https://developer.apple.com/documentation/macos-release-notes/macos-12_3-release-notes, MacOS 12.3 beta updates the builtin version of LibreSSL to 3.3.5, which changes the default digest type of openssl to SHA256. 

This breaks codesigningtool's mobileprovision identity extraction. We can fix this by just explicitly specifying the output format with `-sha1`.

PR #1359, opened 2022-02-09
Add the -sha1 flag to the openssl command used in codesigningtool

tnek:
As called out in https://developer.apple.com/documentation/macos-release-notes/macos-12_3-release-notes, MacOS 12.3 beta updates the builtin version of LibreSSL to 3.3.5, which changes the default digest type of openssl to SHA256.

This breaks codesigningtool's mobileprovision identity extraction. We can fix this by just explicitly specifying the output format with -sha1.

Linked issue: https://github.com/bazelbuild/rules_apple/issues/1358


keith:
Thanks!

PR #1369, opened 2022-03-01
Add example working UI test

keith:
The key here is the module_name of the test source lib is the same as
the test name so the ProductModuleName in the xctestrun file matches the
module name


brentleyjones (approved):

keith:
we don't really have to merge, we'll see, either way it's an example of a working option for the linked bug report

keith:
https://github.com/bazelbuild/rules_apple/pull/1909#issuecomment-1474037323

Issue #1370, opened 2022-03-02
Support skipping tests


birwin93:
Right now users can only use `--test_filter` to specify which tests to run. The generated `ios_test_runner.sh` file only passes `TESTBRIDGE_TEST_ONLY` to `tests_to_run` in the launch options json file.

It would be helpful to also be able to specify tests to skip.

One possible way we could do this is
1) Prefix tests with `^` when using `--test_filter`. eg `--test_filter=^FeatureTests/skipThisTest`
2) Update `ios_test_runner.template.sh` to conditionally pass tests to `tests_to_run` or `skip_testing` if the prefix is there

keith:
In other cases like `--features` prefixing with a minus is the standard, i wonder if we can do that instead? `--test_filter=-a/b`

birwin93:
👍  that works!

keith:
done in https://github.com/bazelbuild/rules_apple/pull/1371

PR #1371, opened 2022-03-02
Support skipping tests with --test_filter

birwin93:
Addresses https://github.com/bazelbuild/rules_apple/issues/1370

Currently the `ios_test_runner` rule only passes the contents of `--test_filter` to `tests_to_run` when invoking xctestrunner.

This PR allows users to now specify if tests should be passed to `skip_tests` by prefixing the test name with `-`

Example:
```
bazel test //Feature:FeatureTests --test_filter=-FeatureTests/someTest
```

### DIFF
--- a/apple/testing/default_runner/ios_test_runner.template.sh
+++ b/apple/testing/default_runner/ios_test_runner.template.sh
@@ -119,12 +119,44 @@ fi
 
 # Use the TESTBRIDGE_TEST_ONLY environment variable set by Bazel's --test_filter
 # flag to set tests_to_run value in ios_test_runner's launch_options.
+# Any test prefixed with '-' will be passed to "skip_tests". Otherwise the tests 
+# is passed to "tests_to_run"
 if [[ -n "$TESTBRIDGE_TEST_ONLY" ]]; then
   if [[ -n "${LAUNCH_OPTIONS_JSON_STR}" ]]; then
     LAUNCH_OPTIONS_JSON_STR+=","
   fi
-  TESTS="${TESTBRIDGE_TEST_ONLY//,/\",\"}"
-  LAUNCH_OPTIONS_JSON_STR+="\"tests_to_run\":[\"$TESTS\"]"
+
+  IFS=","
+  ALL_TESTS=("$TESTBRIDGE_TEST_ONLY")
+  for TEST in $ALL_TESTS; do
+    if [[ ${TEST:0:1} == "-" ]]; then
+      if [[ -n "$SKIP_TESTS" ]]; then
+        SKIP_TESTS+=",${TEST:1}"
+      else
+        SKIP_TESTS="${TEST:1}"
+      fi
+    else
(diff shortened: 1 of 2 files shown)


test/ios_test_runner_unit_test.sh
```diff
@@ -446,6 +446,32 @@ function test_ios_unit_test_with_multi_filter() {
...
+  expect_log "Test Case '-\[PassingUnitTest testPass3\]' passed"
+  expect_log "Test Suite 'PassingUnitTest' passed"
+  expect_log "Test Suite 'PassingUnitTest.xctest' passed"
+  expect_log "Executed 1 test, with 0 failures"
+}
+
+function test_ios_unit_test_with_multi_skip_filter() {
+  create_sim_runners
+  create_ios_unit_tests
+  do_ios_test --test_filter=-PassingUnitTest/testPass,-PassingUnitTest/testPass2 //ios:PassingUnitTest || fail "should pass"
```
keith, line 465: can you add a case where we have both added and removed in the same arg? maybe include the class but exclude a single case from it?

apple/testing/default_runner/ios_test_runner.template.sh
```diff
@@ -119,12 +119,44 @@ fi
...
   if [[ -n "${LAUNCH_OPTIONS_JSON_STR}" ]]; then
     LAUNCH_OPTIONS_JSON_STR+=","
   fi
-  TESTS="${TESTBRIDGE_TEST_ONLY//,/\",\"}"
-  LAUNCH_OPTIONS_JSON_STR+="\"tests_to_run\":[\"$TESTS\"]"
+
+  IFS=","
+  ALL_TESTS=("$TESTBRIDGE_TEST_ONLY")
+  for TEST in $ALL_TESTS; do
+    if [[ ${TEST:0:1} == "-" ]]; then
```
keith, line 132:
```suggestion
    if [[ $TEST == -* ]]; then
```

keith (approved):
looks great! I think that extra test case would be nice to cover just to be safe tho

test/ios_test_runner_unit_test.sh
birwin93: ah good call! I'll add that

birwin93:
Small downside, looks like `skip_tests` only works for `xcodebuild` so need to have a test host at the moment

keith (approved):
too bad, but not surprising

test/ios_test_runner_unit_test.sh
```diff
@@ -445,7 +445,6 @@ function test_ios_unit_test_with_multi_filter() {
   expect_log "Test Suite 'PassingUnitTest.xctest' passed"
   expect_log "Executed 2 tests, with 0 failures"
 }
-
```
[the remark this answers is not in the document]
birwin93: ✅

keith:
Thanks!

PR #1381, opened 2022-03-05
Fix bundle loader coverage and add tests

keith:
This adds tests covering iOS coverage logic, and also fixes a previously
broken case where coverage for the bundle loader was not collected since
we did not pass the binary to llvm-cov.

This also passes the arch flag to llvm-cov, which is required for multi-arch binaries.
This is a bit complicated to handle rosetta, it first attempts to use the host arch,
and if the binary doesn't contain that, falls back to x86_64. This should be improved.
It also doesn't allow binaries to be different archs, but realistically I don't think we
could easily produce that with rules_apple anyways.

### DIFF
--- a/apple/testing/default_runner/ios_test_runner.template.sh
+++ b/apple/testing/default_runner/ios_test_runner.template.sh
@@ -235,13 +235,19 @@ lcov_args=(
 )
 has_binary=false
 IFS=";"
+arch=$(uname -m)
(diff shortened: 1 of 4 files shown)


brentleyjones (approved):

mattrobmattrob (approved):
Looks good in our configuration for both `ios_unit_test`s with and without `test_host`. 👍  Thanks for the quick fix, @keith!

PR #1382, opened 2022-03-07
Remove --ios_multi_cpus arg for simulator tests

keith:
This changes 2 things:

1. It removes 32 bit simulators for the test suite. I imagine there
   aren't many bugs we'd hit on just 1 arch, and given how old 32 bit
   devices are now I don't think it's a huge deal for us to support, so
   this should lightly improve test performance
2. Support testing on Apple Silicon by using the native simulator
   architecture detection, instead of forcing x86_64


brentleyjones (approved):

keith:
This doesn't work, submitted an alternative https://github.com/bazelbuild/rules_apple/pull/1392

PR #1385, opened 2022-03-09
Fix bazel 5.x support for now

keith:
This hack allows us to fake a bazel version check for now.

### DIFF
--- a/apple/internal/linking_support.bzl
+++ b/apple/internal/linking_support.bzl
@@ -157,12 +157,20 @@ def _register_linking_action(
         linkopts.extend(["-bundle_loader", bundle_loader_file.path])
         link_inputs.append(bundle_loader_file)
 
+    # TODO: This is a hack to support bazel 5.x and 6.x at the same time after
+    # should_lipo was removed from the arguments list, but is still required
+    # before that point. The addition of link_multi_arch_static_library probably
+    # doesn't line up perfectly, but should be good enough
+    kwargs = {"should_lipo": False}
+    if getattr(apple_common, "link_multi_arch_static_library", False):
+        kwargs = {}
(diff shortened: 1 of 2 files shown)


apple/internal/linking_support.bzl
```diff
@@ -157,12 +157,20 @@ def _register_linking_action(
         linkopts.extend(["-bundle_loader", bundle_loader_file.path])
         link_inputs.append(bundle_loader_file)
 
+    # TODO: This is a hack to support bazel 5.x and 6.x at the same time after
+    # should_lipo was removed from the arguments list, but is still required
+    # before that point. The addition of link_multi_arch_static_library probably
+    # doesn't line up perfectly, but should be good enough
```
brentleyjones, line 163:
```suggestion
    # doesn't line up perfectly, but should be good enough.
```

apple/internal/linking_support.bzl
```diff
@@ -157,12 +157,20 @@ def _register_linking_action(
         linkopts.extend(["-bundle_loader", bundle_loader_file.path])
         link_inputs.append(bundle_loader_file)
 
+    # TODO: This is a hack to support bazel 5.x and 6.x at the same time after
```
brentleyjones, line 160: I assume we want to create an issue to have this removed once we remove support for 5.x?

keith: I'll submit a revert as soon as this is in

thii (approved):

apple/internal/linking_support.bzl
keith: https://github.com/bazelbuild/rules_apple/pull/1386

PR #1386, opened 2022-03-09
Revert "Fix bazel 5.x support for now"

keith:
This reverts https://github.com/bazelbuild/rules_apple/pull/1385


keith:
we can leave this open until we're happy to drop 5.x

PR #1388, opened 2022-03-12
Add _lcov_merger attribute to test rules

keith:
As of bazel 5.1 this can be used as a late bound attribute https://github.com/bazelbuild/bazel/commit/14d8be0e8ec07aabb6c4497555f712dac5720a5f

This alleviates the concerns from
https://github.com/bazelbuild/rules_apple/pull/691 and allows users to
use this feature, and avoid having to pass `LCOV_MERGER` in the
environment manually

### DIFF
--- a/apple/internal/rule_factory.bzl
+++ b/apple/internal/rule_factory.bzl
@@ -180,6 +180,13 @@ AppleTestRunnerInfo provider.
         cfg = "exec",
         default = Label("@build_bazel_apple_support//tools:coverage_support"),
     ),
+    "_lcov_merger": attr.label(
+        default = configuration_field(
+            fragment = "coverage",
+            name = "output_generator",
+        ),
+        cfg = "exec",
+    ),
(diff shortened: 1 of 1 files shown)


keith:
Do we need a new release first even? since it's in bazel 5.1+? I think of us supporting LTS meaning the newest in the LTS release version

brentleyjones:
My thoughts are:
- We have over 80 commits since the last release
- This is a breaking change for some people, who can't go to 5.1 for some reason
- So why not at least give them the latest changes as a release? We can even do two releases back to back if we want (maybe the second one with this has the new versioning we've been thinking about)?

It doesn't directly impact me, since I think anyone on 5.x should be on 5.2 right now, but that was my reasoning at least.

keith:
ok since we've done a release since then we can probably merge this one now?

brentleyjones (approved):

PR #1390, opened 2022-03-14
Allow for JSON lcov coverage exports via a flag

maxwellE:
Allows JSON coverage results to be created in
`$TEST_UNDECLARED_OUTPUTS_DIR` 
when passing `--test_env=COVERAGE_PRODUCE_JSON=TRUE`

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -31,3 +31,10 @@ build --nocheck_visibility
 # here forever.
 build --define=RULES_SWIFT_BUILD_DUMMY_WORKER=1
 build --strategy=SwiftCompile=local
+
+# Required to avoid missing coverage runner file errors.
+# GitHub Issue: https://github.com/bazelbuild/rules_apple/issues/1128
+coverage --action_env=LCOV_MERGER=/usr/bin/true
+
+# Use llvm-cov instead of gcov (default).
+coverage --experimental_use_llvm_covmap
--- a/apple/testing/default_runner/ios_test_runner.template.sh
+++ b/apple/testing/default_runner/ios_test_runner.template.sh
@@ -253,7 +253,7 @@ readonly error_file="$TMP_DIR/llvm-cov-error.txt"
 llvm_cov_status=0
 xcrun llvm-cov \
   export \
-  -format lcov
+  -format lcov \
   "${lcov_args[@]}" \
   @"$COVERAGE_MANIFEST" \
   > "$COVERAGE_OUTPUT_FILE" \
@@ -272,7 +272,7 @@ if [[ -n "${COVERAGE_PRODUCE_JSON:-}" ]]; then
   llvm_cov_json_export_status=0
   xcrun llvm-cov \
     export \
-    -format text
+    -format text \
     "${lcov_args[@]}" \
     @"$COVERAGE_MANIFEST" \
     > "$TEST_UNDECLARED_OUTPUTS_DIR/coverage.json"
--- a/apple/testing/default_runner/ios_test_runner.template.sh
+++ b/apple/testing/default_runner/ios_test_runner.template.sh
@@ -228,7 +228,6 @@ readonly profdata="$TMP_DIR/coverage.profdata"
 xcrun llvm-profdata merge "$profraw" --output "$profdata"
 
 lcov_args=(
-  -format lcov
   -instr-profile "$profdata"
   -ignore-filename-regex='.*external/.+'
   -path-equivalence="$ROOT",.
@@ -254,6 +253,7 @@ readonly error_file="$TMP_DIR/llvm-cov-error.txt"
 llvm_cov_status=0
 xcrun llvm-cov \
   export \
+  -format lcov
   "${lcov_args[@]}" \
   @"$COVERAGE_MANIFEST" \
   > "$COVERAGE_OUTPUT_FILE" \
@@ -268,17 +268,11 @@ if [[ -s "$error_file" || "$llvm_cov_status" -ne 0 ]]; then
   exit 1
 fi
 
-llvm_cov_json_export_status=0
-if [[ "$TEST_ENV" =~ "COVERAGE_PRODUCE_JSON" ]]; then
-  mkdir -p "$TEST_UNDECLARED_OUTPUTS_DIR"
-  lcov_args=(
-    -format text
-    -instr-profile "$profdata"
-    -ignore-filename-regex='.*external/.+'
-    -path-equivalence="$ROOT",.
-  )
+if [[ -n "${COVERAGE_PRODUCE_JSON:-}" ]]; then
+  llvm_cov_json_export_status=0
   xcrun llvm-cov \
     export \
+    -format text
     "${lcov_args[@]}" \
     @"$COVERAGE_MANIFEST" \
     > "$TEST_UNDECLARED_OUTPUTS_DIR/coverage.json"
--- a/apple/testing/default_runner/ios_test_runner.template.sh
+++ b/apple/testing/default_runner/ios_test_runner.template.sh
@@ -129,7 +129,7 @@ fi
 
 # Use the TESTBRIDGE_TEST_ONLY environment variable set by Bazel's --test_filter
 # flag to set tests_to_run value in ios_test_runner's launch_options.
-# Any test prefixed with '-' will be passed to "skip_tests". Otherwise the tests 
+# Any test prefixed with '-' will be passed to "skip_tests". Otherwise the tests
 # is passed to "tests_to_run"
 if [[ -n "$TESTBRIDGE_TEST_ONLY" ]]; then
   if [[ -n "${LAUNCH_OPTIONS_JSON_STR}" ]]; then
@@ -266,4 +266,27 @@ if [[ -s "$error_file" || "$llvm_cov_status" -ne 0 ]]; then
   echo "error: while exporting coverage report" >&2
   cat "$error_file" >&2
   exit 1
+fi
+
+llvm_cov_json_export_status=0
+if [[ "$TEST_ENV" =~ "COVERAGE_PRODUCE_JSON" ]]; then
+  mkdir -p "$TEST_UNDECLARED_OUTPUTS_DIR"
+  lcov_args=(
+    -format text
+    -instr-profile "$profdata"
+    -ignore-filename-regex='.*external/.+'
+    -path-equivalence="$ROOT",.
+  )
+  xcrun llvm-cov \
+    export \
+    "${lcov_args[@]}" \
+    @"$COVERAGE_MANIFEST" \
+    > "$TEST_UNDECLARED_OUTPUTS_DIR/coverage.json"
+    2> "$error_file" \
+    || llvm_cov_json_export_status=$?
+  if [[ -s "$error_file" || "$llvm_cov_json_export_status" -ne 0 ]]; then
+    echo "error: while exporting json coverage report" >&2
+    cat "$error_file" >&2
+    exit 1
+  fi
 fi
--- a/apple/testing/default_runner/macos_test_runner.template.sh
+++ b/apple/testing/default_runner/macos_test_runner.template.sh
@@ -151,12 +151,15 @@ xcrun llvm-profdata merge "$profraw" --output "$profdata"
 
 readonly export_error_file="$TEST_TMP_DIR/llvm-cov-export-error.txt"
 llvm_cov_export_status=0
+lcov_args=(
+  -instr-profile "$profdata"
+  -ignore-filename-regex='.*external/.+'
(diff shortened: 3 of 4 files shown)


tinder-maxwellelliott:
> Allows JSON coverage results to be created in
> `$TEST_UNDECLARED_OUTPUTS_DIR` for macOS
> when passing `--test_env=COVERAGE_PRODUCE_JSON=TRUE`

.bazelrc
```diff
@@ -31,3 +31,10 @@ build --nocheck_visibility
 # here forever.
 build --define=RULES_SWIFT_BUILD_DUMMY_WORKER=1
 build --strategy=SwiftCompile=local
+
+# Required to avoid missing coverage runner file errors.
+# GitHub Issue: https://github.com/bazelbuild/rules_apple/issues/1128
+coverage --action_env=LCOV_MERGER=/usr/bin/true
```
maxwellE, line 37: Added these settings here to raise the visibility of the required flags to make coverage work

apple/testing/default_runner/macos_test_runner.template.sh
```diff
@@ -116,9 +116,15 @@ if [[ "${COVERAGE:-}" -eq 1 ]]; then
...
 XCTESTRUN_ENV=""
+COVERAGE_PRODUCE_JSON=false
 for SINGLE_TEST_ENV in ${TEST_ENV//,/ }; do
   IFS== read key value <<< "$SINGLE_TEST_ENV"
   XCTESTRUN_ENV+="<key>$(escape "$key")</key><string>$(escape "$value")</string>"
+  if [[ "${COVERAGE:-}" -eq 1 ]]; then
+    if [[ "$key" = "COVERAGE_PRODUCE_JSON" && "$value" = "TRUE" ]]; then
+      COVERAGE_PRODUCE_JSON=true
+    fi
+  fi
```
keith, line 127: I don't think we should have to parse this, I think this would just be set in the environment with `--test_env`?

apple/testing/default_runner/macos_test_runner.template.sh
```diff
@@ -149,8 +155,8 @@ fi
 readonly profdata="$TEST_TMP_DIR/coverage.profdata"
 xcrun llvm-profdata merge "$profraw" --output "$profdata"
 
-readonly error_file="$TEST_TMP_DIR/llvm-cov-error.txt"
-llvm_cov_status=0
+readonly export_error_file="$TEST_TMP_DIR/llvm-cov-export-error.txt"
```
keith, line 158: we can re-use this error file for both, since we check after each to see if it's empty and error, same with the status code. We can rename the variables as needed though

apple/testing/default_runner/macos_test_runner.template.sh
```diff
@@ -160,13 +166,41 @@ xcrun llvm-cov \
...
   @"$COVERAGE_MANIFEST" \
   > "$COVERAGE_OUTPUT_FILE" \
-  2> "$error_file" \
-  || llvm_cov_status=$?
+  2> "$export_error_file" \
+  || llvm_cov_export_status=$?
+
+llvm_cov_json_export_status=0
+readonly json_export_error_file="$TEST_TMP_DIR/llvm-cov-json-export-error.txt"
+if [ "$COVERAGE_PRODUCE_JSON" = true ]; then
```
keith, line 174: In other cases I'm just checking if the variable in question is set at all so folks don't have to worry about specific values, the downside is if folks want to negate it that doesn't work, wdyt?

apple/testing/default_runner/macos_test_runner.template.sh
```diff
@@ -160,13 +166,41 @@ xcrun llvm-cov \
...
+  if [[ -n "${TEST_UNDECLARED_OUTPUTS_DIR}" ]]; then
```
keith, line 175: Can this ever not be set?

apple/testing/default_runner/macos_test_runner.template.sh
```diff
@@ -160,13 +166,41 @@ xcrun llvm-cov \
...
+      > "$TEST_UNDECLARED_OUTPUTS_DIR/coverage.json"
+        2> "$json_export_error_file" \
+        || llvm_cov_json_export_status=$?
+  fi
+fi
 
 # Error ourselves if lcov outputs warnings, such as if we misconfigure
 # something and the file path of one of the covered files doesn't exist
-if [[ -s "$error_file" || "$llvm_cov_status" -ne 0 ]]; then
+if [[ -s "$export_error_file" || "$llvm_cov_export_status" -ne 0 ]]; then
```
keith, line 194: You can move this up to above the json check for the ability to reuse the variables

apple/testing/default_runner/macos_test_runner.template.sh
```diff
@@ -160,13 +166,41 @@ xcrun llvm-cov \
...
   echo "error: while exporting coverage report" >&2
-  cat "$error_file" >&2
+  cat "$export_error_file" >&2
   exit 1
 fi
+
+if [ "$COVERAGE_PRODUCE_JSON" = true ]; then
+  if [[ -s "$json_export_error_file" || "$llvm_cov_json_export_status" -ne 0 ]]; then
```
keith, line 201: once you move the other block up you can add this inside the if that actually produces it so we don't have to dup the conditional

examples/macos/CommandLineSwift/Sources/main.swift
```diff
@@ -17,5 +17,5 @@ import Foundation
 // Since the Info.plist file gets embedded in the binary, we can access values
 // like the bundle identifier using the NSBundle APIs.
 let bundle = Bundle.main
-NSLog("Hello World from \(bundle.bundleIdentifier ?? "<none>")")
-NSLog("\nHere is the entire Info.plist dictionary: \(bundle.infoDictionary ?? [:])")
+let logger: Logger = .init(log: LogImp())
```
keith, line 20: Are all these test changes intentional? I can't tell what they're specifically testing above coverage but maybe it was just helpful for your testing? I'm working on unit tests in https://github.com/bazelbuild/rules_apple/pull/1381 that ideally we could expand to macOS too

keith (commented):
did you also want to do this for iOS in this PR or separately?

maxwellE:
> did you also want to do this for iOS in this PR or separately?

Will add in this PR

apple/testing/default_runner/macos_test_runner.template.sh
maxwellE: Checking for existence sounds good to me

apple/testing/default_runner/macos_test_runner.template.sh
maxwellE: This was copied from one of the other runner shell scripts, down tor remove it

examples/macos/CommandLineSwift/Sources/main.swift
maxwellE: This was simply an attempt to make a test and get some coverage on logic, there is nothing special and I can remove it

apple/testing/default_runner/ios_test_runner.template.sh
```diff
@@ -267,3 +267,26 @@ if [[ -s "$error_file" || "$llvm_cov_status" -ne 0 ]]; then
   cat "$error_file" >&2
   exit 1
 fi
+
+llvm_cov_json_export_status=0
```
keith, line 271: looks like this can move inside the `if`?

apple/testing/default_runner/ios_test_runner.template.sh
```diff
@@ -267,3 +267,26 @@ if [[ -s "$error_file" || "$llvm_cov_status" -ne 0 ]]; then
...
+if [[ "$TEST_ENV" =~ "COVERAGE_PRODUCE_JSON" ]]; then
```
keith, line 272:
```suggestion
if [[ -n "${COVERAGE_PRODUCE_JSON:-}" ]]; then
```

I think this var is set in the environment so we can check this way?

apple/testing/default_runner/ios_test_runner.template.sh
```diff
@@ -267,3 +267,26 @@ if [[ -s "$error_file" || "$llvm_cov_status" -ne 0 ]]; then
...
+  mkdir -p "$TEST_UNDECLARED_OUTPUTS_DIR"
```
keith, line 273: this seems fine but doese this directory ever not exist?

apple/testing/default_runner/ios_test_runner.template.sh
```diff
@@ -267,3 +267,26 @@ if [[ -s "$error_file" || "$llvm_cov_status" -ne 0 ]]; then
...
+  lcov_args=(
```
keith, line 274: we will have to mirror the args above about the specific binaries to correctly support hosted binaries. It looks to me like we can remove `-format lcov` from the `lcov_args` above, and pass that directly above, and then share all the rest of these arguments which will solve that and remove the need for this separate list?

apple/testing/default_runner/macos_test_runner.template.sh
```diff
@@ -160,13 +160,34 @@ xcrun llvm-cov \
...
-if [[ -s "$error_file" || "$llvm_cov_status" -ne 0 ]]; then
+if [[ -s "$export_error_file" || "$llvm_cov_export_status" -ne 0 ]]; then
   echo "error: while exporting coverage report" >&2
-  cat "$error_file" >&2
+  cat "$export_error_file" >&2
   exit 1
 fi
+
+llvm_cov_json_export_status=0
+if [[ "$TEST_ENV" =~ "COVERAGE_PRODUCE_JSON" ]]; then
```
keith, line 175: similar comments as above

keith (commented):
I submitted https://github.com/bazelbuild/rules_apple/pull/1405 which along with ios_coverage_test.sh should allow you to write unit tests that validate this behavior

keith:
bump since this would solve https://github.com/bazelbuild/rules_apple/pull/1411 as well and serve as a template for doing the same for HTML

tinder-maxwellelliott:
> bump since this would solve #1411 as well and serve as a template for doing the same for HTML

Updating now!

test/ios_coverage_test.sh
```diff
@@ -166,6 +166,14 @@ function test_standalone_unit_test_coverage() {
   assert_contains "SharedLogic.m:-\[SharedLogic doSomething\]" "test-testlogs/app/standalone_test/coverage.dat"
 }
 
+function test_standalone_unit_test_coverage_json() {
+  create_common_files
+  do_coverage ios --test_output=errors --ios_minimum_os=9.0 --experimental_use_llvm_covmap --test_env=COVERAGE_PRODUCE_JSON=1 //app:standalone_test || fail "Should build"
+  cd "test-testlogs/app/standalone_test/test.outputs" || fail "Should unzip"
+  unzip "outputs.zip"
+  assert_contains "\"name\":\"SharedLogic.m:-\[SharedLogic doSomething\]\"" "coverage.json"
```
keith, line 174:
```suggestion
  unzip_single_file "test-testlogs/app/standalone_test/test.outputs/outputs.zip" coverage.json \
      grep '"name":"SharedLogic.m:-\[SharedLogic doSomething\]"'
```

I think this is the same? similar to below? if the grep fails it should fail the test. not sure how much of a difference it makes but it should be more ephemeral than the current version

keith (approved):
Awesome, thanks!

wendyliga:
@maxwellE how do you access `$TEST_UNDECLARED_OUTPUTS_DIR` outside of bazel test? how do you fetch the file after bazel coverage, do you use a manual path?

do bazel has any API to fetch this dir?

keith:
After the build we access it through the bazel-testlogs symlink

tinder-maxwellelliott:
For files like `test.lcov` you will get those directly back in the Build Event Protocol format listed under `testResult`'s `testActionOutput`s. You can construct paths relative to those paths to discover files created under `$TEST_UNDECLARED_OUTPUTS_DIR`. This is all without RBE though, unsure if that affects this

wendyliga:
bazel-testlogs could work if we know the exact target.
i think utilizing `testResult's` on [Build Event Protocol](https://docs.bazel.build/versions/main/build-event-protocol.html) is the right approach, and then construct it to the json coverage file

testResult: 
`/bazel-out/darwin_arm64-dbg/testlogs/myTest/myTest/coverage.dat`

json filepath:  
`/bazel-out/darwin_arm64-dbg/testlogs/myTest/myTest/test.outputs/coverage.json`

wendyliga:
new comer question, couldn't we put coverage.json on `testResult` too? can `testResult` has multiple coverage result?

PR #1391, opened 2022-03-14
Update runfiles documentation in common_info

maxwellE:
Provides a more realistic usecase for using runfiles
in tests. Also adds a note about an odd bazel
bug that breaks this approach in certain
situations


keith (approved):
thanks!

PR #1392, opened 2022-03-15
Add sim_arm64 to ios_multi_cpus in tests

keith:
This makes it so whatever host arch we're running on there is a matching
slice. We cannot remove this flag and just take the default because test
implementations depend on parsing the ios_multi_cpus flag to get the
archs. Without this tests that try to actually run fail on M1 machines
because they were built for x86_64.


thii (approved):

keith:
This breaks the consumers of current_archs because it would need to strip the `sim_` prefix

PR #1393, opened 2022-03-16
Add a test to verify that `libswift_Concurrency.dylib` is bundled if app uses Swift Concurrency and targets iOS >= 13 and < 15.

thii:
Also verify that `libswiftCore.dylib` should not get bundled in this
case, which is broken with Xcode 13.3.


thii:
BazelCI uses Xcode 13.0 so this doesn't build here yet.

thii:
https://github.com/bazelbuild/continuous-integration/issues/1348

PR #1396, opened 2022-03-18
Don't pass path of `swift-5.0` stdlib to `swift-stdlib-tool` when not neeeded

thii:
Workaround for https://bugs.swift.org/browse/SR-16010:
`libswift_Concurrency.dylib` bundled with Xcode 13.3 refenrences
`@rpath/libswiftCore.dylib` instead of `/usr/lib/libswiftCore.dylib`,
which causes `libswiftCore.dylib` to unnecessarily get copied into the
product bundle.

Fixes https://github.com/bazelbuild/rules_apple/issues/1397.

### DIFF
--- a/tools/swift_stdlib_tool/swift_stdlib_tool.py
+++ b/tools/swift_stdlib_tool/swift_stdlib_tool.py
@@ -25,15 +25,16 @@ from build_bazel_rules_apple.tools.wrapper_common import execute
 from build_bazel_rules_apple.tools.wrapper_common import lipo
 
 
-def _copy_swift_stdlibs(binaries_to_scan, sdk_platform, destination_path):
+def _copy_swift_stdlibs(binaries_to_scan, swift_dylibs_paths, sdk_platform,
+                        destination_path):
   """Copies the Swift stdlibs required by the binaries to the destination."""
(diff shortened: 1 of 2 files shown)


tools/swift_stdlib_tool/swift_stdlib_tool.py
```diff
@@ -25,15 +25,16 @@
...
   # Rely on the swift-stdlib-tool to determine the subset of Swift stdlibs that
   # these binaries require.
   developer_dir = os.environ["DEVELOPER_DIR"]
-  swift_dylibs_root = "Toolchains/XcodeDefault.xctoolchain/usr/lib/swift-*"
-  swift_library_dir_pattern = os.path.join(developer_dir, swift_dylibs_root,
-                                           sdk_platform)
-  swift_library_dirs = glob.glob(swift_library_dir_pattern)
+  swift_library_dirs = [
+    os.path.join(developer_dir, x, sdk_platform)
+    for x in swift_dylibs_paths
```
keith, line 36:
```suggestion
    os.path.join(developer_dir, dylibs_path, sdk_platform)
    for dylibs_path in swift_dylibs_paths
```

keith (approved):

Issue #1397, opened 2022-03-18
lipo Thinning unable to find arm64 on apple silicon xcode 13.3


wendyliga:
environment:
- Xcode 13.3
- MacOS 12.3
- min ios 11.0

how to repro
- clone https://github.com/wendyliga/simple_bazel
- `bazel build //ios/DummyApplication`

when using async/await and GroupActivities SDK from apple, `build_bazel_rules_apple/tools/swift_stdlib_tool/swift_stdlib_tool.py` will return 2 libs
- /Applications/Xcode_13.3.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/swift-5.0/iphonesimulator/libswiftCore.dylib
- /Applications/Xcode_13.3.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/swift-5.5/iphonesimulator/libswift_Concurrency.dylib

but if we check the arch on `libswiftCore.dylib`, It only contains x86_64 and i386.
<details>
  <summary>Spoiler</summary>
<img src="https://user-images.githubusercontent.com/16457495/158939528-d89603dd-f7d0-4495-99fe-b9d14d3872f9.png"/>
</details>

This will make `swift_stdlib_tool.py` unable to thin `arm64`, and will return error
<details>
  <summary>Spoiler</summary>
ERROR: /Users/wendy.liga/Developer/simple_bazel/ios/DummyApplication/BUILD:10:16: SwiftStdlibCopy ios/DummyApplication/DummyApplication-intermediates/swiftlibs failed: (Exit 1): swift_stdlib_tool failed: error executing command bazel-out/darwin_arm64-opt-exec-2B5CBBC6-ST-15bb1cbb5e86/bin/external/build_bazel_rules_apple/tools/swift_stdlib_tool/swift_stdlib_tool --platform iphonesimulator --output_path ... (remaining 4 arguments skipped)
Traceback (most recent call last):
  File "/private/var/tmp/_bazel_wendy.liga/8284c4a94ccbbb579b9e33d50d39273c/execroot/__main__/bazel-out/darwin_arm64-opt-exec-2B5CBBC6-ST-15bb1cbb5e86/bin/external/build_bazel_rules_apple/tools/swift_stdlib_tool/swift_stdlib_tool.runfiles/build_bazel_rules_apple/tools/swift_stdlib_tool/swift_stdlib_tool.py", line 132, in <module>
    sys.exit(main())
[...]

PR #1400, opened 2022-03-22
Fix error when building for simulator on Apple silicon with Xcode 13.3

thii:
Additional workaround for https://bugs.swift.org/browse/SR-16010.

There's no such a simulator on Apple silicon Macs with a minimum OS
version before ABI stability, so it's unnecessary to bundle the original
Swift runtime here. This avoids error when we thin the dylibs later,
because the original Swift dylibs don't contain arm64 slice for
simulator.

Fixes https://github.com/bazelbuild/rules_apple/issues/1397

### DIFF
--- a/apple/internal/partials/swift_dylibs.bzl
+++ b/apple/internal/partials/swift_dylibs.bzl
@@ -128,6 +128,16 @@ def _swift_dylib_action(
         xcode_path_wrapper = platform_prerequisites.xcode_path_wrapper,
     )
 
+def _target_platform_is_arm_simulator(platform_prerequisites):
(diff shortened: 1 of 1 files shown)


keith:
Is this what Xcode does on m1s as well

keith (approved):
it would be nice to know, but seems good to unblock

thii:
Unless it's an archive action, Xcode's copying logic is not based on the minimum OS version, but the OS version of the actual device you're selecting. AFAICT, the minimum iOS simulator version on M1s is 12.4, so Xcode never has to copy the original Swift dylibs for M1 simulator builds.

PR #1405, opened 2022-03-23
Fix macOS bundle loader coverage and add tests

keith:
This is the same as https://github.com/bazelbuild/rules_apple/pull/1381 but for macOS


keith:
blocked on https://github.com/bazelbuild/continuous-integration/issues/1273

PR #1408, opened 2022-03-24
Add config_setting for tvos_sim_arm64

dflems:
Added to bazel here: https://github.com/bazelbuild/bazel/commit/207c31be57b9863f5c1aa17361a4ed68e6eaf041

Just landed in Bazel 5.1

### DIFF
--- a/apple/BUILD
+++ b/apple/BUILD
@@ -22,6 +22,7 @@ licenses(["notice"])
         "ios_sim_arm64",
         "tvos_x86_64",
         "tvos_arm64",
+        "tvos_sim_arm64",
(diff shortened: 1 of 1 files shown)


keith (approved):

brentleyjones:
Does this mean `master` stops supporting Bazel 5.0?

keith:
AFAIUI this is just a config setting that you can use, it just wouldn't ever hit before without that change in bazel

dflems:
@brentleyjones Adding a `ios_sim_arm64` config setting to this same file didn't break older versions of bazel so it should be fine

PR #1411, opened 2022-03-30
Use `LLVM_PROFILE_FILE` from env if exist

wendyliga:
recently there're changes to generate coverage data automatically if requested on `ios_unit_test`
add an option to use `LLVM_PROFILE_FILE` given from `--test_env`

if none, will use $TMP_DIR instead

### DIFF
--- a/apple/testing/default_runner/ios_test_runner.template.sh
+++ b/apple/testing/default_runner/ios_test_runner.template.sh
@@ -105,12 +105,24 @@ else
   TEST_ENV="TEST_SRCDIR=$TEST_SRCDIR"
 fi
 
-readonly profraw="$TMP_DIR/coverage.profraw"
+# default profraw
+profraw="$TMP_DIR/coverage.profraw"
+# check if LLVM_PROFILE_FILE has been set on --test_env, if exist, will use the given path instead
+IFS=',' read -ra ENV <<< "$TEST_ENV"
+for i in "${ENV[@]}"; do
+  IFS='=' read -ra VALUE <<< "$i"
+  for j in "${VALUE[@]}"; do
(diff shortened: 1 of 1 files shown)


wendyliga:
@keith to access the profraw and process the coverage manually, right now , with current approach, the bazel coverage will overwrite the LLVM_PROFILE_FILE

wendyliga:
On my case i want to process the profraw, and use json insted of lcov

also i want to use llvm cov to generate html from the profraw

keith:
If the test runner produced those as outputs as well, would that work for your case? For example similar to https://github.com/bazelbuild/rules_apple/pull/1390

wendyliga:
@keith 
> If the test runner produced those as outputs as well, would that work for your case? For example similar to https://github.com/bazelbuild/rules_apple/pull/1390

that could solve one issue.
the problem is just, with the new update, we don't have a way to access the raw `profraw`. 

I don't see a problem processing the `profraw`, to `lcov` and `json`, but because we don't have any access to `profraw`, there's no flexibility to handle the code coverage manually, somehow on my end, I have a different way to process the `profraw`.

the effect of this PR is just to add ability to access the profraw if you so choose, but doesn't effect the general code coverage output of lcov dat, or json in the future.

keith:
Yea that makes sense, but what else besides json / html do you need the profraw for? If it's just those 2 I'd rather the test runner produce those files for everyone than expose this implementation detail. I'm also loosely worried about how it's up to you to make sure to workaround the sandbox with this file, it just sounds error prone.

wendyliga:
@keith this is the HTML implementation https://github.com/bazelbuild/rules_apple/pull/1490

PR #1416, opened 2022-04-08
Fix dSYM filename in case custom executable names are used

rsahara:
Fixes the dSYM filename in case a custom executable name is used, like this:
- before this fix `bundle_name.app.dSYM/bundle_name`;
- after this fix `bundle_name.app.dSYM/executable_name`.

Changes in the code:
- Fixed the filename by using the executable name instead of the bundle name when it's specified (i.e. use the value returned by `bundling_support.executable_name`).
- Fixed the dSYM test `ios_application_custom_executable_name_dsyms_test` to support custom executable names. (Basically a revert, the last commits made this test ignore the custom executable name argument.)

### DIFF
--- a/apple/internal/partials/debug_symbols.bzl
+++ b/apple/internal/partials/debug_symbols.bzl
@@ -115,21 +115,20 @@ def _collect_linkmaps(
 def _copy_dsym_into_declared_bundle(
         *,
         actions,
-        debug_output_filename,
         dsym_binaries,
         dsym_bundle_name,
+        dsym_output_filename,
         _executable_name,
         platform_prerequisites):
     """Declares the dSYM binary file and copies it into the preferred .dSYM bundle location.
 
     Args:
       actions: The actions provider from `ctx.actions`.
-      debug_output_filename: The base file name to use for this debug output, which will be followed
-        by the architecture with an underscore to make the dSYM binary file name or with the bundle
-        extension following it for the dSYM bundle file name.
       dsym_binaries: A mapping of architectures to Files representing dsym binary outputs for each
         architecture.
       dsym_bundle_name: The full name of the dSYM bundle, including its extension.
+      dsym_output_filename: The base file name to use for this debug output, which will be followed
+        by the architecture with an underscore to make the dSYM binary file name.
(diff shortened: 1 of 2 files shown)


rsahara:
This is for fixing https://github.com/bazelbuild/rules_apple/pull/1271#discussion_r844719844.

@thii Would appreciate a review.
It worked for my project, but I'm not sure if it treated debug_discriminator correctly (I think I'm not using it).

apple/internal/partials/debug_symbols.bzl
```diff
@@ -300,13 +300,17 @@ def _debug_symbols_partial_impl(
 
     if platform_prerequisites.cpp_fragment:
         if platform_prerequisites.cpp_fragment.apple_generate_dsym:
+            dsym_output_filename = executable_name
+            if debug_discriminator:
+                dsym_output_filename += "_" + debug_discriminator
```
thii, line 305: I think if you make the following change to line 284

```diff
-debug_output_filename = bundle_name
+debug_output_filename = executable_name if executable_name else bundle_name 
```

https://github.com/bazelbuild/rules_apple/blob/a93ad60b28877b0cfa16abafc6bdd3fce0e66c76/apple/internal/partials/debug_symbols.bzl#L284-L286

there wouldn't be a need for a new `dsym_output_filename` parameter.

rsahara: I tried that change at first, but the output filename was
executable_name.app.dSYM/executable_name,
and the test (ios_application_custom_executable_name_dsyms_test) was expecting 
bundle_name.app.dSYM/executable_name
so I tried to separate the bundle name and filename.

Is executable_name.app.dSYM/executable_name OK too?

Also, I checked that the executable_name here is never None because we always pass the result of bundling_support.executable_name to this function, and basically it's equal to `executable_name if executable_name else bundle_name`

thii: Sorry I didn't realize that.

> Is executable_name.app.dSYM/executable_name OK as output?

I think `bundle_name.app.dSYM/executable_name` is the expected output.

thii (approved):

PR #1417, opened 2022-04-08
Emit resultBundlePath for macos_test_runner.

sebastianv1:
Supplies a `resultBundlePath` to the macos test runner xcodebuild invocation. This matches the behavior and filename used by the ios test runner.

### DIFF
--- a/apple/testing/default_runner/macos_test_runner.template.sh
+++ b/apple/testing/default_runner/macos_test_runner.template.sh
@@ -140,7 +140,7 @@ fi
 rm -rf "$TEST_UNDECLARED_OUTPUTS_DIR/test.xcresult"
 xcodebuild test-without-building \
     -destination "platform=macOS" \
-    -resultBundlePath "${TEST_UNDECLARED_OUTPUTS_DIR}/test.xcresult" \
+    -resultBundlePath "$TEST_UNDECLARED_OUTPUTS_DIR/test.xcresult" \
     -xctestrun "$XCTESTRUN"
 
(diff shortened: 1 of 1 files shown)


apple/testing/default_runner/macos_test_runner.template.sh
```diff
@@ -137,8 +137,10 @@ fi
 # Run xcodebuild with the xctestrun file just created. If the test failed, this
 # command will return non-zero, which is enough to tell bazel that the test
 # failed.
+mkdir -p "${TEST_UNDECLARED_OUTPUTS_DIR}"
```
keith, line 140:
```suggestion
rm -rf "$TEST_UNDECLARED_OUTPUTS_DIR/test.xcresult"
```

We shouldn't need to make this directory ahead of time, but there is an issue where it isn't cleaned if you have a flaky test that is rerun multiple times, and xcodebuild will fail if the path already exists

apple/testing/default_runner/macos_test_runner.template.sh
```diff
@@ -137,8 +137,10 @@ fi
...
 xcodebuild test-without-building \
     -destination "platform=macOS" \
+    -resultBundlePath "${TEST_UNDECLARED_OUTPUTS_DIR}/test.xcresult" \
```
keith, line 143:
```suggestion
    -resultBundlePath "$TEST_UNDECLARED_OUTPUTS_DIR/test.xcresult" \
```

keith (commented):
2 small comments, looks good!

keith (approved):
thanks!

Issue #1418, opened 2022-04-08
Generate Swift source for CoreML models


a2:
(I'm interested in working on Swift source generation for CoreML models. I saw in the [contributing guidelines](https://github.com/bazelbuild/rules_apple/blob/a93ad60b28877b0cfa16abafc6bdd3fce0e66c76/CONTRIBUTING.md) that I should file an issue to "claim" this work item.)

I noticed that the `coremlc generate` subcommand now — at least as of version 1127.100.30, which comes bundled with Xcode 13.3 — supports generating Swift source files with the public access modifiers (`--public-access`):

<details>
<summary>Output from <code>xcrun coremlc generate --help</code></summary>

```
$ xcrun coremlc generate --help
usage: coremlc <command> <inputdocument> <outputpath> [options ...]

Examples:

[...]

Commands:

[...]

   generate <model.{mlmodel,mlpackage}> </desired/path/to/output/directory>
      Generate source code interface for supplied CoreML .mlmodel or .mlpackage with the following options:
      --outputpath </desired/path/to/output/directory> : where to place the generated interface files.
      --language <Swift|Objective-C> : interface language.
      --swift-version x.y : generated interface requires Swift version x.y (swift only).
      --public-access : classes and members get public access specifier (swift only).
      --globalmodule : generate objc compatible model classes (swift only).
      --sdkroot : filename base for System/Library/CoreServices/SystemVersion.plist (internal, swift only). Requires --platform.
      --platform <macos|ios|watchos|tvos|maccatalyst> : check mlmodel version compatability for this platform.
      --deployment-target xx.yy.zz : check mlmodel version compatability for this version.
      --container <bundle-resources|swift-package> : code generation variants (internal).
      --no-documentation : suppress inline documentation.
      --dry-run : don't write generated code to output directory.

[...]
```
</details>

As for implementing this, and because naming is one of the two most difficult problems in programming (along with cache invalidation and off-by-one errors), I see that there is are `objc_intent_library` and `swift_intent_library` functions (Macros? I'm new to Bazel) in [_apple/resources.bzl_](https://github.com/bazelbuild/rules_apple/blob/a93ad60b28877b0cfa16abafc6bdd3fce0e66c76/apple/resources.bzl).

Should `apple_core_ml_library` be renamed to something like `objc_apple_core_ml_library` to make room for `swift_apple_core_ml_library`?

keith:
https://github.com/bazelbuild/rules_apple/pull/1419

PR #1419, opened 2022-04-09
Generate Swift source for CoreML models

a2:
Implementation of #1418

> I noticed that the `coremlc generate` subcommand now — at least as of version 1127.100.30, which comes bundled with Xcode 13.3 — supports generating Swift source files with the public access modifiers (`--public-access`):
> 
> <details>
> <summary>Output from <code>xcrun coremlc generate --help</code></summary>
> 
> ```
> $ xcrun coremlc generate --help
> usage: coremlc <command> <inputdocument> <outputpath> [options ...]
> 
> Examples:
> 
> [...]
> 
> Commands:
> 
> [...]
> 
>    generate <model.{mlmodel,mlpackage}> </desired/path/to/output/directory>
>       Generate source code interface for supplied CoreML .mlmodel or .mlpackage with the following options:
>       --outputpath </desired/path/to/output/directory> : where to place the generated interface files.
>       --language <Swift|Objective-C> : interface language.
>       --swift-version x.y : generated interface requires Swift version x.y (swift only).
>       --public-access : classes and members get public access specifier (swift only).
>       --globalmodule : generate objc compatible model classes (swift only).
>       --sdkroot : filename base for System/Library/CoreServices/SystemVersion.plist (internal, swift only). Requires --platform.
>       --platform <macos|ios|watchos|tvos|maccatalyst> : check mlmodel version compatability for this platform.
>       --deployment-target xx.yy.zz : check mlmodel version compatability for this version.
>       --container <bundle-resources|swift-package> : code generation variants (internal).
>       --no-documentation : suppress inline documentation.
>       --dry-run : don't write generated code to output directory.
> 
> [...]
> ```
> </details>
> 
> As for implementing this, and because naming is one of the two most difficult problems in programming (along with cache invalidation and off-by-one errors), I see that there is are `objc_intent_library` and `swift_intent_library` functions (Macros? I'm new to Bazel) in [_apple/resources.bzl_](https://github.com/bazelbuild/rules_apple/blob/a93ad60b28877b0cfa16abafc6bdd3fce0e66c76/apple/resources.bzl).

### DIFF
--- a/apple/internal/resource_rules/apple_core_ml_library.bzl
+++ b/apple/internal/resource_rules/apple_core_ml_library.bzl
@@ -31,10 +31,6 @@ load(
     "platform_support",
 )
 load(
-    "@build_bazel_rules_apple//apple/internal:swift_support.bzl",
-    "swift_support",
-)
-load(
     "@bazel_skylib//lib:dicts.bzl",
     "dicts",
 )
@@ -48,14 +44,22 @@ def _apple_core_ml_library_impl(ctx):
     actions = ctx.actions
     basename = paths.replace_extension(ctx.file.mlmodel.basename, "")
 
-    deps = getattr(ctx.attr, "deps", None)
-    uses_swift = swift_support.uses_swift(deps) if deps else False
+    is_swift = ctx.attr.language == "Swift"
 
-    coremlc_source = actions.declare_file(
-        "{}.m".format(basename),
-        sibling = ctx.outputs.source,
-    )
-    coremlc_header = actions.declare_file("{}.h".format(basename), sibling = coremlc_source)
+    if is_swift and not ctx.attr.swift_source:
+        fail("Attribute `swift_source` is mandatory when generating Swift.")
+    if not is_swift and (not ctx.attr.objc_header or not ctx.attr.objc_source):
+        fail("Attributes `objc_header` and `objc_source` are mandatory when generating Objective-C.")
+
+    swift_output_src = None
+    objc_output_src = None
+    objc_output_hdr = None
+
+    if is_swift:
+        swift_output_src = actions.declare_file("{}.swift".format(basename))
+    else:
+        objc_output_src = actions.declare_file("{}.m".format(basename))
+        objc_output_hdr = actions.declare_file("{}.h".format(basename))
 
     # TODO(b/168721966): Consider if an aspect could be used to generate mlmodel sources. This
     # would be similar to how we are planning to use the resource aspect with the
@@ -71,7 +75,7 @@ def _apple_core_ml_library_impl(ctx):
         features = ctx.features,
         objc_fragment = None,
         platform_type_string = str(ctx.fragments.apple.single_arch_platform.platform_type),
-        uses_swift = uses_swift,
+        uses_swift = is_swift,
         xcode_path_wrapper = ctx.executable._xcode_path_wrapper,
         xcode_version_config = ctx.attr._xcode_config[apple_common.XcodeVersionConfig],
     )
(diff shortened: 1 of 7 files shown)


a2:
I ended up reverting to separate implementations of `generate_mlmodel_sources` and `generate_swift_mlmodel_sources` due to the way `ctx.outputs.source` ([Obj-C](https://github.com/a2/rules_apple/blob/3e83f3455f733520dc9853a4d61dcf56bbc11fba/apple/internal/resource_rules/apple_core_ml_library.bzl#L131-L133), [Swift](https://github.com/a2/rules_apple/blob/3e83f3455f733520dc9853a4d61dcf56bbc11fba/apple/internal/resource_rules/apple_core_ml_library.bzl#L222-L224)) is used / expects to be used by the macro implementation. Although the tests passed locally (minus the documentation issue which I fixed in the latest commit), I noticed an issue with resolving the model files in my own project.

apple/internal/resource_rules/apple_core_ml_library.bzl
```diff
@@ -153,8 +153,76 @@ as `#import my/package/MyModel.h`.
 
 This rule will also compile the `mlmodel` into an `mlmodelc` and propagate it
 upstream so that it is packaged as a resource inside the top level bundle.
+""",
+)
+
+def _swift_apple_core_ml_library_impl(ctx):
```
keith, line 159: Looking at the differences between this and the existing `_apple_core_ml_library_impl` it seems like there are very few changes. instead of creating a separate rule for this what do you think about having a parameter on that internal rule for whether it generates swift or not, I guess just something like `generates_swift = True`, defaulting to false for keeping the current behavior. And then doing the same in the macro, to either create a objc_library or swift_library as the final target.

I think overall that would be a nicer user interface, and you'd dodge your naming concern. Thoughts?

apple/internal/resource_rules/apple_core_ml_library.bzl
```diff
@@ -114,7 +114,7 @@ apple_core_ml_library = rule(
             allow_single_file = ["mlmodel"],
             mandatory = True,
             doc = """
-Label to a single `.mlmodel` file from which to generate sources and compile into mlmodelc files.
+Label to a single `.mlmodel` or `.mlpackage` file from which to generate sources and compile into mlmodelc files.
```
keith, line 117: could you split out the mlpackage changes into a separate PR and add a fixture like the one that currently exists for mlmodel?

apple/internal/resource_rules/apple_core_ml_library.bzl
a2: I had an implementation like this in one of my commits on this branch. I wasn't able to make it work in practice, though, due to a larger, critical detail between `outputs = { "source": "%{name}.swift" }` (in the Swift version) and `outputs = { "source": "%{name}.m" }` (in the Obj-C version).

is there a way to declare a rule that supports something like this? Otherwise I agree that having a "language" parameter, similar to the `apple_intent_library` rule, would be ideal.

keith: If I understand the use case you don't have to predeclare outputs like this (it's actually discouraged now). You can just call ctx.actions.declare_file with the name instead.

a2: Interesting! I'll take another look tomorrow then 😃

keith (approved):
this looks great! one small comment, and you might need to rebase since I merged your other PR?

apple/internal/resource_rules/apple_core_ml_library.bzl
```diff
@@ -71,7 +75,7 @@ def _apple_core_ml_library_impl(ctx):
         features = ctx.features,
         objc_fragment = None,
         platform_type_string = str(ctx.fragments.apple.single_arch_platform.platform_type),
-        uses_swift = uses_swift,
+        uses_swift = is_swift,
```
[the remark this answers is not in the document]
keith: ah yea probably need to do:

```suggestion
        uses_swift = is_swift or (swift_support.uses_swift(deps) if deps else False),
```

keith:
the test failures appear related

keith:
Awesome, great work! Thanks!

PR #1422, opened 2022-04-16
Enable additional support for CoreML models in mlpackage format

a2:
Making a separate PR for this change per the discussion in #1419, which adds supplementary support for CoreML models in the `.mlpackage` bundle format. (Previously only the `.mlmodel` binary format was supported.)

Note that one of the two PRs will need to be rebased if the other is merged first.


keith (approved):
nice, thanks!

PR #1423, opened 2022-04-19
Ensure `[AppleBundleInfo].binary` exists for coverage

mattrobmattrob:
**Summary:**
When `AppleBundleInfo` is created without a `binary`, allow coverage to still be gathered from the other `direct_binaries`. Example that creates the provider with a non-populated/`None` binary:

https://github.com/bazel-ios/rules_ios/blob/57edfbb4205fcfaae82f7c42c32e137d9c77634a/rules/internal/framework_middleman.bzl#L74-L82


keith (approved):

keith:
not sure we should encourage binary-less AppleBundleInfo, but this change is small enough that I think it's fine

PR #1424, opened 2022-04-23
Fix linting for plisttool and unit tests.

keith:
PiperOrigin-RevId: 424391499
(cherry picked from commit a9fd9c497635ee125d4aa7320a183a8fc7942b12)


keith:
diff best viewed without whitespace

PR #1427, opened 2022-04-23
Fix visibility formatting

keith:
RELNOTES: None
PiperOrigin-RevId: 432245920
(cherry picked from commit 56b79f2d6f2cfc4506e8ce9f5ee73d6e06110b22)


keith:
I think this might break buttons on CI, and i'll just revert that bit

PR #1443, opened 2022-04-25
Add upstream xcframework improvements

keith:
See each commit for details. This depends on https://github.com/bazelbuild/bazel/commit/877845584186301e53ed01eaecf3166db4ddcec7 in bazel, which is not in the 5.x branches

### DIFF
--- a/doc/rules-apple.md
+++ b/doc/rules-apple.md
@@ -132,6 +132,32 @@ objc_library(
 | <a id="apple_static_framework_import-weak_sdk_frameworks"></a>weak_sdk_frameworks |  Names of SDK frameworks to weakly link with. For instance, <code>MediaAccessibility</code>. In difference to regularly linked SDK frameworks, symbols from weakly linked frameworks do not cause an error if they are not present at runtime.   | List of strings | optional | [] |
 
 
+<a id="#apple_static_xcframework"></a>
+
+## apple_static_xcframework
+
+<pre>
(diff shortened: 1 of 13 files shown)


thii (approved):

jpsim (approved):
`apple_static_xcframework` appears to be working well in my tests here: https://github.com/envoyproxy/envoy-mobile/pull/2217

apple/internal/apple_framework_import.bzl
```diff
@@ -810,6 +810,8 @@ objc_library(
     ],
 )
 ```
+
+NOTE: This is only support on bazel 6.0+
```
jpsim, line 814:
```suggestion
NOTE: This is only supported on bazel 6.0+
```

keith: oof thanks

doc/rules-apple.md
```diff
@@ -117,6 +117,8 @@ objc_library(
 )
 ```
 
+NOTE: This is only supported on bazel 6.0+
```
brentleyjones, line 120: Does this only apply to `apple_static_framework_import`? What about `apple_static_xcframework`?

keith: actually i think this one is fine, just not in our tests, because we generate one to tetst it

brentleyjones (approved):

PR #1446, opened 2022-04-25
Add missing Mnemonic

keith:
RELNOTES: None
PiperOrigin-RevId: 439913077
(cherry picked from commit f85aee527879a85edaedc04e09efef379a7fd11f)

### DIFF
--- a/apple/internal/partials/debug_symbols.bzl
+++ b/apple/internal/partials/debug_symbols.bzl
@@ -153,7 +153,7 @@ def _copy_dsym_into_declared_bundle(
         actions.run_shell(
             inputs = [dsym_binary],
             outputs = [output_binary],
-            mnemonic = "DsymDwarf",
+            mnemonic = "CopyDWARF2dSYM",
             progress_message = "Copy DWARF into dSYM `%s`" % dsym_binary.short_path,
             command = """
 if [[ $OSTYPE == darwin* ]]; then


apple/internal/partials/debug_symbols.bzl
```diff
@@ -153,7 +153,7 @@ def _copy_dsym_into_declared_bundle(
         actions.run_shell(
             inputs = [dsym_binary],
             outputs = [output_binary],
-            mnemonic = "DsymDwarf",
+            mnemonic = "CopyDWARF2dSYM",
```
brentleyjones, line 156: This is a breaking change, do we want to take it?

keith: I don't feel strongly. It seemed like a pretty minor one so I wasn't worried about it.

brentleyjones: I know there are people excluding this mnemonic, so it would be annoying if it just started being unexcluded. I vote not to take it.

keith: Yea, it's probably not worth breaking people. But in general we should be able to make changes like this as long as we mention them in the notes. In this example I think the upstream mnemonic is more descriptive, which is nicer.

BalestraPatrick: We reference `DsymDwarf` right now so for us this would be a breaking change. As long as it's clearly noted in the release notes, I'm fine with it.

PR #1449, opened 2022-04-25
Add logic to reference the lipoed dSYM bundles instead of the non-standard, non-lipoed dSYM bundles as default & config group outputs when the defined variable apple.tree_artifact_dsym_files is set.

keith:
PiperOrigin-RevId: 441794931
(cherry picked from commit ea9f128ccc52c3b89cef6a86a3263b5daba2259f)


keith:
needs a diff cherry pick that I haven't submitted yet

keith:
moved to that PR https://github.com/bazelbuild/rules_apple/pull/1477/

PR #1456, opened 2022-04-28
Return the correct entitlements in `AppleBundleInfo`

brentleyjones:
e6d792e2916577845571a99112c7595f013b3fad and 8764a6ea5fc508f6426e6e162aa7bab744829207 changed which entitlements file was returned in `AppleBundleInfo`. It used to be the final processed entitlements, but now it's just the one used to code sign, which isn't as useful to consumers of the provider.

Also adds some tests to ensure this doesn't regress in the future.


keith (approved):
worth a starlark test so we don't regress since this is pretty subtle

brentleyjones:
Added some tests!

PR #1464, opened 2022-05-02
Use rolling releases for local development

keith:
Since we broke 5.x compat for some cases, this is nicer to use so you
can run tests over everything.

### DIFF
--- a/.bazelversion
+++ b/.bazelversion
@@ -1,0 +1,1 @@
+rolling


keith:
hrm does this file affect CI or is it ignored?

keith:
looks like it's ignored, which is good

brentleyjones (approved):

PR #1478, opened 2022-05-24
Give all static libraries produced by the apple_static_xcframework rule file names based directly on the inferred bundle name.

keith:
PiperOrigin-RevId: 448546861
(cherry picked from commit e9a882304d2131d12e5fd32b968b33a0752e1d74)


keith:
There were some test conflicts here which I picked the right version of for now but will have to be re-updated with https://github.com/bazelbuild/rules_apple/pull/1447, it's fine they will fail because the numbers like `2` and `7` become strings instead

segiddins (approved):

PR #1481, opened 2022-05-24
Re-enable sandboxing in tests

keith:


keith:
i don't think this should cause issues, but clearly it did in the past

brentleyjones (approved):

PR #1484, opened 2022-05-26
Add compression for xcframework archives

keith:
Previously all zips created by bundletool would not compress their
contents. For envoy-mobile the uncompressed zip is ~2.3gb where the
compressed zip is ~434mbs. Since this rule is always meant for
distribution, I think compressing their outputs is worth the extra CPU
time.

ios_static_framework gets compression through a different codepath since
it relies on the processor to create the zip instead.

### DIFF
--- a/apple/internal/xcframework_rules.bzl
+++ b/apple/internal/xcframework_rules.bzl
@@ -439,6 +439,7 @@ def _create_xcframework_bundle(
         bundle_merge_zips = framework_archive_merge_zips,
         bundle_path = bundle_name + ".xcframework",
         output = output_archive.path,
+        compress = True,
     )
     actions.write(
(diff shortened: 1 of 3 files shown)


brentleyjones (approved):

keith:
macOS unzip doesn't support `zipfile.ZIP_BZIP2` it seems, going with `ZIP_DEFLATED` instead, which increases my test case size to 434mb

jpsim (approved):

keith:
for future reference lzma also doesn't work with macOS unzip, but the archive in this case is 168mb, so massive improvement there. I don't like the idea of customizing this but folks might still want to unzip / rezip in the case they can support that in their toolchain

PR #1490, opened 2022-06-09
Allow for HTML lcov coverage exports via a flag

wendyliga:
extending https://github.com/bazelbuild/rules_apple/pull/1390

Allows HTML coverage results to be created in$TEST_UNDECLARED_OUTPUTS_DIR
when passing --test_env=COVERAGE_PRODUCE_HTML=TRUE

### DIFF
--- a/apple/testing/default_runner/ios_test_runner.template.sh
+++ b/apple/testing/default_runner/ios_test_runner.template.sh
@@ -284,3 +284,21 @@ if [[ -n "${COVERAGE_PRODUCE_JSON:-}" ]]; then
     exit 1
   fi
 fi
+
+if [[ -n "${COVERAGE_PRODUCE_HTML:-}" ]]; then
+  llvm_cov_html_export_status=0
+  xcrun llvm-cov \
+    show \
+    -format html \
+    -use-color \
+    "${lcov_args[@]}" \
+    @"$COVERAGE_MANIFEST" \
+    > "$TEST_UNDECLARED_OUTPUTS_DIR/coverage.html"
+    2> "$error_file" \
+    || llvm_cov_html_export_status=$?
+  if [[ -s "$error_file" || "$llvm_cov_html_export_status" -ne 0 ]]; then
+    echo "error: while exporting html coverage report" >&2
+    cat "$error_file" >&2
+    exit 1
+  fi
+fi
--- a/apple/testing/default_runner/ios_test_runner.template.sh
+++ b/apple/testing/default_runner/ios_test_runner.template.sh
@@ -287,12 +287,13 @@ fi
 
 if [[ -n "${COVERAGE_PRODUCE_HTML:-}" ]]; then
   llvm_cov_html_export_status=0
+
+  # we couldn't use $COVERAGE_MANIFEST", as it will result an empty html
   xcrun llvm-cov \
     show \
     -format html \
     -use-color \
     "${lcov_args[@]}" \
-    @"$COVERAGE_MANIFEST" \
     > "$TEST_UNDECLARED_OUTPUTS_DIR/coverage.html"
     2> "$error_file" \
     || llvm_cov_html_export_status=$?


apple/testing/default_runner/ios_test_runner.template.sh
```diff
@@ -284,3 +284,22 @@ if [[ -n "${COVERAGE_PRODUCE_JSON:-}" ]]; then
     exit 1
   fi
 fi
+
+if [[ -n "${COVERAGE_PRODUCE_HTML:-}" ]]; then
+  llvm_cov_html_export_status=0
+
+  # we couldn't use $COVERAGE_MANIFEST", as it will result an empty html
```
keith, line 291: After adding this here like this:

```
diff --git a/apple/testing/default_runner/ios_test_runner.template.sh b/apple/testing/default_runner/ios_test_runner.template.sh
index 67b44514..818d32d7 100644
--- a/apple/testing/default_runner/ios_test_runner.template.sh
+++ b/apple/testing/default_runner/ios_test_runner.template.sh
@@ -294,6 +294,7 @@ if [[ -n "${COVERAGE_PRODUCE_HTML:-}" ]]; then
     -format html \
     -use-color \
     "${lcov_args[@]}" \
+    @"$COVERAGE_MANIFEST" \
     > "$TEST_UNDECLARED_OUTPUTS_DIR/coverage.html"
     2> "$error_file" \
     || llvm_cov_html_export_status=$?
```

your test still passes for me, are you seeing something different? i think we always need to limit this otherwise you might get tons of transitive dependency noise in your html reports if you statically link your dependencies

wendyliga: you are right, but adding `@"$COVERAGE_MANIFEST"` will result an empty HTML.

wendyliga: @keith any update?

keith: I think the problem is that you need to `cd "$ROOT"` before doing this, does that work?

wendyliga: @keith nice catch. so CMIIW, the `@COVERAGE_MANIFEST` will force the coverage to only include the files that are described there right?

I couldn't find any reference for `COVERAGE_MANIFEST` instead of this file, how do Bazel create this file?

keith: yes that sounds right, i don't see anything in the docs for it either. but you can `cat` it in the test script for debugging. IIRC the reason I hit this `cd` workaround before was because the relative paths were somehow invalid on disk from the default PWD in the test runner, which didn't work for `llvm-cov`. does

wendyliga: yes, but i need to change the path for the xctest executable here, to get the full path on bazel-out
```swift
def _get_coverage_execution_environment(_ctx, covered_binaries):
    - covered_binary_paths = [f.short_path for f in covered_binaries.to_list()]
    + covered_binary_paths = [f.path for f in covered_binaries.to_list()]
```

wendyliga: also looks like the `instrument_files` that is being used by `llvm-cov`, only contain file inside the unit test, so the HTML only consist of the source code for the unit test

keith: are you setting `--instrumentation_filter` to something that should include more binaries?

wendyliga: @keith nope, just default bazel coverage

keith: I see the same issue here generally. Interestingly our internal test runner works with this just fine:

```sh
xcrun llvm-cov \
  show \
  -format html \
  -instr-profile "$profdata" \
  -path-equivalence=.,"$PWD" \
  -output-dir="$TEST_UNDECLARED_OUTPUTS_DIR/html" \
  "$test_tmp_dir/$test_bundle_name.xctest/$test_bundle_name" \
  @"$COVERAGE_MANIFEST"
```

so I'm still not sure why there's a difference.

Lets merge without the coverage manifest limiting today and try to fix it up later. Only other thing, should we do the html directory instead of a single file like my example here? that way you don't have 1 huge html file but many small ones and an index

wendyliga: @keith regarding HTML file, AFAIK the `llvm-cov` only generate single HTML file.

keith: My example above outputs a directory

wendyliga: @keith will update the PR

wendyliga: @keith worked with `output-dir`. will it be better if we add an option to generate a single html file too?

keith: I'm not really sure. IMO it seems like folks with any non-trivial module size would want the directory instead, what do you think? looks like the test needs an update if we keep this route

wendyliga: Agree, I think we can add it later if necessary.

will update the test

wendyliga: @keith please help re review

keith: looks good but tests seem to be failing here

wendyliga: is there any way to `cat` the log, to see the error detail?

thii: https://buildkite.com/bazel/rules-apple-darwin/builds/5897#01823cc1-6110-4617-9429-ad4db90aa7d3

wendyliga: @thii yup, but the error log is located on a specific file

Screen Shot 2022-07-27 at 16 17 12
e

wendyliga: running it on my end seems to work, but I will try to mimic ci's command

wendyliga: works on my end, but looks like on the ci
it tries to refer to `/app` and not `app`.

do you guys have any idea about it

Screen Shot 2022-07-28 at 15 28 44

wendyliga:
I think we can refer to https://wiki.documentfoundation.org/Development/Lcov 
to convert lcov to HTML.

I think we can close this PR

PR #1495, opened 2022-06-13
Apply the same configuration to `resources` as `deps`

brentleyjones:
This ensures that resources included via a top-level target's `resources` attribute have the same transition applied to them as those included transitively through `deps`'d `data`.

Before this change, a resource bundle included via `resources` has this path for it's generated plist:

```
applebin_ios-ios_sim_arm64-dbg-ST-dd300a2e2c23/bin/ExampleNestedResources/ExampleNestedResources-intermediates/ExampleNestedResources.bundle/Info.plist
```

and this path when included via a dependency:

```
ios-sim_arm64-min15.0-applebin_ios-ios_sim_arm64-dbg-ST-dd300a2e2c23/bin/ExampleNestedResources/ExampleNestedResources-intermediates/ExampleNestedResources.bundle/Info.plist
```

Now both have the same path.

### DIFF
--- a/apple/internal/rule_factory.bzl
+++ b/apple/internal/rule_factory.bzl
@@ -298,7 +298,7 @@ bundle.
         },
     )
 
-def _get_common_bundling_attributes(rule_descriptor):
+def _get_common_bundling_attributes(deps_cfg, rule_descriptor):
     """Returns a list of dictionaries with attributes common to all bundling rules."""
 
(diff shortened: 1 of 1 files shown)


keith:
hrm i wonder why this wasn't done before. maybe the thought was top level resources like this shouldn't need the transition?

brentleyjones:
Yeah, they don't "need" it, it's more for consistency. They get a version of a transition from the top-level rule itself, but that is slightly different from the one used on `deps`.

PR #1496, opened 2022-06-13
Add support for UUID in profile finder

mattrobmattrob:
Add the ability to use UUID (in addition to name) to identify provisioning profiles on the users system. This helps when you want 100% specificity around which profile to use. Use `expected_specifier` as the param name to match Xcode's `PROVISIONING_PROFILE_SPECIFIER` setting name.

This allows for both to be used:
```Starlark
    provisioning_profile = select({
        "//:Debug": "bb393eb3-d030-4796-8a5d-b67243fa1234",
        "//:AppStore": "AppStoreProfile_2022-06-01",
    }),
```

### DIFF
--- a/apple/internal/local_provisioning_profiles.bzl
+++ b/apple/internal/local_provisioning_profiles.bzl
@@ -116,7 +116,7 @@ def _local_provisioning_profile(ctx):
 local_provisioning_profile = rule(
     attrs = {
         "profile_name": attr.string(
-            doc = "Name (or UUID) of the profile to use, if it's not provided the name of the rule is used",
+            doc = "Name of the profile to use, if it's not provided the name of the rule is used",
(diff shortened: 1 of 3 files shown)


brentleyjones (approved):

brentleyjones:
@keith What commands need to be run to update the docs correctly here?

keith:
`bazel run doc:update`

keith:
can you push an empty commit 🤔

mattrobmattrob:
Done, @keith 👍

keith:
something is up with stardoc, it's not the first time, I guess just drop the docs changes here and if that passes we can merge and then deal with it when it fails

mattrobmattrob:
And so it was. 👋

PR #1497, opened 2022-06-17
Use `path` instead of `short_path` in `AppleResourceInfo`

brentleyjones:
This is mainly to make `AppleResourceInfo.processed_origins` useful outside of the internal `_deduplication` function. With this change, one can use `processed_origins` to find the resource file behind processed files, such as the original Info.plist uses.

This isn't an API change, since the docs never specify if the paths are short or not, but it's s slight breaking change, if for example someone reimplemented `_deduplicate`. As long as we mention it in the release notes, we should be fine.

### DIFF
--- a/apple/internal/partials/resources.bzl
+++ b/apple/internal/partials/resources.bzl
@@ -168,15 +168,15 @@ def _deduplicate(
           duplicating files referenced by library targets and top level targets.
 
     Returns:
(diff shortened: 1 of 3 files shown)


brentleyjones:
Hmm, I don't care enough about this to try to fix that error right now.

PR #1502, opened 2022-06-17
Add release version scheme note

keith:


README.md
```diff
@@ -98,3 +98,23 @@ ios_application(
...
+release on the [releases
+page](https://github.com/bazelbuild/rules_apple/releases).
+
+Besides these constraint this repo follows [semver](https://semver.org/)
+as best as we can since the 1.0.0 release.
+
+| Bazel release | Minimum supported rules version | Final supported rules version|
+|:-------------------:|:-------------------:|:-------------------------:|
+| 6.x (most recent rolling) | 0.34.0 | current |
+| 5.x | 0.33.0 | current |
```
brentleyjones, line 118: Technically did we support 5.x before 0.33.0? Or did we need to make an update just to support it? Seems like there would be some overlap in these version ranges.

keith: yea these aren't perfect, this was the first version where we mentioned the release was tested with it, but it _might_ have been supported sooner, hard to say without testing. we can fix these up if folks find they're inaccurate and need to show that

brentleyjones (approved):

PR #1503, opened 2022-06-17
Document becoming a maintainer

keith:
This process has been the case since we took over maintainership of the
rules, but it hasn't been written down. We want this on ramp to be clear
in the case that folks are interested.

### DIFF
--- a/MAINTENANCE.md
+++ b/MAINTENANCE.md
@@ -11,6 +11,21 @@ rules repos (this doc is not duplicated across the repos):
 The current maintainers can be seen in the [CODEOWNERS](CODEOWNERS)
 file.
 
+## How to become a maintainer
+
+These rules are community maintained. Any community member can become a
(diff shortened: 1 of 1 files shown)


brentleyjones (approved):

MAINTENANCE.md
```diff
@@ -11,6 +11,21 @@ rules repos (this doc is not duplicated across the repos):
 The current maintainers can be seen in the [CODEOWNERS](CODEOWNERS)
 file.
 
+## How to become a maintainer
+
+These rules are community maintained. Any community member from any
+company can become a maintainer. Ideally we have a large set of
```
ob, line 17: I'd leave it at "community member" no need to be employed? Feels more inclusive.

ob (approved):
Short and sweet. I likes it.

MAINTENANCE.md
keith: ah yep, sounds good thanks

PR #1509, opened 2022-06-30
Expose `apple_intent_library` as public API

brentleyjones:
This allows using the generated sources directly in a library, similar to how Xcode is setup by default.


thii (approved):

brentleyjones:
Last green failures are us not adapting to latest bazel in some way.

keith (approved):

PR #1521, opened 2022-07-07
Add transition changes from upstream commit

keith:
This is a partial commit of
https://github.com/bazelbuild/rules_apple/pull/1447 which includes some
more changes removing macros that would break bazel 5.x compat. So with
this change we retain the `_static_framework_transition` for now, and we
can delete it as part of merging that PR


keith:
merging https://github.com/bazelbuild/rules_apple/pull/1447 instead

PR #1525, opened 2022-07-07
Remove 5.x CI on HEAD

keith:


.bazelci/presubmit.yml
```diff
@@ -57,7 +31,7 @@ tasks:
 
   doc_tests:
     name: "Latest Bazel"
```
brentleyjones, line 33: Should we change the title?

brentleyjones (approved):

PR #1534, opened 2022-07-07
Remove bitcode support

keith:
Now that Xcode 14.1 is the minimum supported to the App Store, and Xcode
14.0 dropped support for bitcode, we don't need to handle these flags
anymore.

Closes https://github.com/bazelbuild/rules_apple/issues/1956


keith:
we can merge this eventually

keith:
https://github.com/bazelbuild/bazel/issues/18321

mattrobmattrob (approved):
What's the plan for things like `@build_bazel_rules_apple//apple/internal:bitcode_support.bzl`? Shouldn't all callers of something like `bitcode_support.bitcode_mode_string(...)` just always assume it's `none` now and strip bitcode, if it's present?

keith:
yea probably, just wanted to do the minimum to get us green again

PR #1536, opened 2022-07-07
Move CI back to last_green

keith:


keith:
needs https://github.com/bazelbuild/rules_apple/pull/1535

PR #1540, opened 2022-07-08
Fix codesign output format in tests

keith:
I'm not sure if it's OS version, Xcode version, or M1, but the default
output for this case is no longer xml which is what the tests were
expecting.


brentleyjones (approved):

brentleyjones:
I thought there was an upstream commit for this as well, but I might be recalling that incorrectly.

keith:
looks like this must be an xcode / os thing because this flag isn't supported on CI

brentleyjones:
https://github.com/bazelbuild/rules_apple/commit/473875c6a344fce0f37fede8858da7ed6a28e75f

keith:
https://github.com/bazelbuild/rules_apple/pull/1542

PR #1553, opened 2022-07-08
Take upstream apple_dynamic_xcframework_import rule

keith:
This change adds the first version of the XCFramework import rule for
XCFrameworks containing Apple frameworks with dynamic libraries.

PiperOrigin-RevId: 452564946
(cherry picked from commit a973b75ed28298019c146d192357677c36f7bb69)

### DIFF
--- a/apple/internal/framework_import_support.bzl
+++ b/apple/internal/framework_import_support.bzl
@@ -36,6 +36,7 @@ def _cc_info_with_dependencies(
         grep_includes,
         header_imports,
         label,
+        swiftmodule_imports,
         includes = [],
         is_framework = True):
     """Returns a new CcInfo which includes transitive Cc dependencies.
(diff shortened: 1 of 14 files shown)


apple/internal/framework_import_support.bzl
```diff
@@ -155,28 +151,73 @@ def _classify_framework_imports(config_vars, framework_imports):
...
+        is_bundle_root_file = parent_dir_name.endswith(".framework")
+        if is_bundle_root_file:
+            bundle_name, _ = paths.split_extension(parent_dir_name)
+            if file.basename == bundle_name:
+                binary_imports.append(file)
+                continue
+
         bundling_imports.append(file)
 
+    # TODO: Enable these checks once static library support works with them
```
brentleyjones, line 200: Is this Google's todo or ours?

keith: i added this one, right now since this is shared with our old static impl this check fails, but i will fix that as more cherry picks land

brentleyjones (approved):

PR #1554, opened 2022-07-08
Remove unused apple_xcframework attr.

keith:
This dates back to the early architecture of the API, which was based on prelim understandings of what Swift Package Manager could accept for XCFrameworks.

In practice it only accepts three types; (dynamic) frameworks, dynamic libraries (which are not useful outside of macOS or testing flows for [i|tv|watch]OS), and static libraries.

PiperOrigin-RevId: 457579889
(cherry picked from commit f6a8e506040a5be9347f31bb713117779a60fe19)


brentleyjones:
The commit comment is wrong with our version, right? Maybe worth adding extra commentary?

brentleyjones:
We guess it's talking about SPM.

brentleyjones (approved):

PR #1555, opened 2022-07-08
Migrate ios_framework_import tests to Starlark

keith:
- Deprecates shell tests for iOS framework import rules to starlark_tests.
- Updated generate_framework to be able to use swift_library outputs to
generate a single architecture Swift static framework.
- Updated binary symbols contains to be able to match partial symbols

PiperOrigin-RevId: 453265762
(cherry picked from commit af389568de912f4e48f5baa50d8e77b4d99e718a)


keith:
this is missing the test case added in https://github.com/bazelbuild/rules_apple/pull/894/files, I will add it manually once https://github.com/bazelbuild/rules_apple/pull/1558 lands

keith:
had to move into https://github.com/bazelbuild/rules_apple/pull/1564 because there was so much overlap, merging there

PR #1557, opened 2022-07-08
Migrate ObjcConfiguration usage to equivalent info in CppConfiguration

keith:
We would like to migrate all such usage to CppConfiguration so that we
can delete the info in ObjcConfiguration.

PiperOrigin-RevId: 457500060
(cherry picked from commit 1cddb75cc00ca6a580a77f0e3e22a2827b1f7cb9)


keith:
considering we've dropped 5.x on this brnach i think we can take this without conditional hacks

brentleyjones (approved):

PR #1564, opened 2022-07-09
Take upstream apple_static_xcframework_import rule

keith:
This rule replaces our previous rule for the same thing, it's a bit more
sophisticated in discovering library identifiers

(cherry picked from commit 01974e8c23486887641a5fd1681a4fde5f580114)


apple/internal/apple_xcframework_import.bzl
```diff
@@ -299,6 +476,119 @@ def _apple_dynamic_xcframework_import_impl(ctx):
...
+    has_swift = ctx.attr.has_swift
+    label = ctx.label
+    linkopts = ctx.attr.linkopts
+    xcframework_imports = ctx.files.xcframework_imports
+    xcode_config = ctx.attr._xcode_config[apple_common.XcodeVersionConfig]
+
+    xcframework = _classify_xcframework_imports(ctx.var, xcframework_imports)
+    target_triplet = cc_toolchain_info_support.get_apple_clang_triplet(cc_toolchain)
+
+    # if xcframework.bundle_type == _BUNDLE_TYPE.frameworks:
```
brentleyjones, line 499: Is the thought that a comment here makes cherry-picking easier? If so, can we leave a comment to that effect?

test/starlark_tests/apple_static_xcframework_import_tests.bzl
```diff
@@ -0,0 +1,70 @@
...
+load(":rules/common_verification_tests.bzl", "archive_contents_test")
+
+def apple_static_xcframework_import_test_suite(name):
+    """Test suite for apple_static_xcframework_import.
+
+    Args:
+      name: the base name to be used in things created by this macro
+    """
+
+    # # Verify importing XCFramework with static frameworks (i.e. not libraries) fails.
```
brentleyjones, line 27: Same, a comment here?

brentleyjones (approved):

apple/internal/apple_xcframework_import.bzl
keith: no i think we should toss this once we're sure it works

test/starlark_tests/apple_static_xcframework_import_tests.bzl
keith: i will toss this one as well, i just want to make sure we have sufficient coverage in this path, which i think we do from our existing rule, but i need to verify

keith:
depends on test helpers from https://github.com/bazelbuild/rules_apple/pull/1555/

Issue #1537, opened 2022-07-07
Created static xcframeworks don't include resources


keith:
If you create a static xcframework with some resources on the `data` attribute of a transitive, those are currently silently ignored. I imagine the initial implementation did this because Xcode doesn't support including resources from static xcframeworks out of the box, but it sounds like CocoaPods does, and bazel definitely does, so if they're there we should include them.

PR #1577, opened 2022-07-15
Add resources support to apple_static_xcframework

keith:
Fixes https://github.com/bazelbuild/rules_apple/issues/1537


keith:
there's quite a bit of overhead here since this rule don't use the standard infra with the processor and all. shouldn't change much though.

apple/internal/xcframework_rules.bzl
```diff
@@ -1026,6 +1026,75 @@ def _apple_static_xcframework_impl(ctx):
...
+            environment_plist = ctx.file._environment_plist_ios,
+            executable_name = executable_name,
+            launch_storyboard = None,
+            output_discriminator = library_identifier,
+            platform_prerequisites = platform_prerequisites,
+            resource_deps = resource_deps,
+            rule_descriptor = rule_descriptor,
+            rule_label = label,
+            version = None,
+        ))
```
keith, line 1073: above this is mostly copied from the dynamic xcfw implementation, same with the attrs, the loop below is the unique part

thii (approved):

PR #1584, opened 2022-07-16
Add generate_static_framework_xcframework test rule

keith:
Previously we could generate dynamic framework or static library based
xcframeworks, but not static framework based xcframeworks. Right now we
do a lot of generation of those manually, but we can replace them with
this.


keith:
mostly copied from the dynamic rule just swapping the call from dynamic to static

brentleyjones (approved):

PR #1587, opened 2022-07-18
Work around a bug in swift-stdlib-tool that results in the Swift runtime being bundled whenever the Swift concurrency back-deploy runtime library is bundled

keith:
The version of `swift-stdlib-tool` that ships with Xcode 13.4.1 bundles an unnecessary copy of the Swift runtime whenever it bundles the back-deploy version of the Swift concurrency runtime. This is because the back-deploy version of the Swift concurrency runtime contains an `@rpath`-relative reference to the Swift runtime due to being built with a deployment target that predates the Swift runtime being shipped with operating system. This `@rpath`-relative reference to the runtime is sufficient for `swift-stdlib-tool` to believe the runtime also needs to be bundled with the application.

We work around this within `swift_stdlib_tool.py` by checking whether any of the binaries that we asked `swift-stdlib-tool` to process have a deployment target that is low enough to require bundling of the Swift runtime. If not, we remove the Swift runtime from the output directory populated by `swift-stdlib-tool`.

This is a workaround for https://github.com/apple/swift/issues/58271

PiperOrigin-RevId: 461684127
(cherry picked from commit 65852aab324b4862e613566e59ef7215ce9eaafb)

### DIFF
--- a/tools/swift_stdlib_tool/swift_stdlib_tool.py
+++ b/tools/swift_stdlib_tool/swift_stdlib_tool.py
@@ -16,6 +16,7 @@
 import argparse
 import glob
 import os
+import re
 import shutil
 import sys
 import tempfile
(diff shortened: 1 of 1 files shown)


keith:
xcode uses an internal version of this tool, so they sidestep this issue, their implementation must have more knowledge of platform specifics, swift-stdlib-tool entirely relies on the rpaths, which are clearly subject to this type of build issue

thii:
Related: https://github.com/bazelbuild/rules_apple/pull/1396
Maybe revert that first?

thii:
https://github.com/bazelbuild/rules_apple/pull/1619

thii (approved):

PR #1588, opened 2022-07-20
Make testrunner attributes public

jerrymarino:
In rules_ios we're building out Apple Silicon based ephemeral VM test
runners to improve running in remote execution and locally. Currently,
we use rules_apple xctestrunner to run the tests under Bazel directly.
You can find this PR here:
https://github.com/bazel-ios/rules_ios/pull/509

Today if you'd like to make a different test runner, you have to copy
the rule and making the attributes public would help hook into them
better. The archicture as it stands is quite nice. Ideally we could make
`sim_creator` an implementation detail of the testrunner / template
combo: or feed that in a different way but leaving as-is for now

### DIFF
--- a/apple/testing/default_runner/ios_test_runner.bzl
+++ b/apple/testing/default_runner/ios_test_runner.bzl
@@ -45,13 +45,13 @@ def _ios_test_runner_impl(ctx):
     device_type = ctx.attr.device_type or ctx.fragments.objc.ios_simulator_device or ""
 
     ctx.actions.expand_template(
-        template = ctx.file._test_template,
+        template = ctx.file.test_template,
         output = ctx.outputs.test_runner_template,
         substitutions = _get_template_substitutions(
             device_type = device_type,
             os_version = os_version,
             simulator_creator = ctx.executable._simulator_creator.short_path,
-            testrunner = ctx.executable._testrunner.short_path,
+            testrunner = ctx.executable.testrunner.short_path,
         ),
     )
     return [
@@ -65,7 +65,7 @@ def _ios_test_runner_impl(ctx):
         ),
         DefaultInfo(
             runfiles = ctx.attr._simulator_creator[DefaultInfo].default_runfiles
-                .merge(ctx.attr._testrunner[DefaultInfo].default_runfiles),
+                .merge(ctx.attr.testrunner[DefaultInfo].default_runfiles),
         ),
     ]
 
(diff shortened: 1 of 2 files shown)


keith:
thanks for the PR. so I see 2 downsides here worth discussing:

1. this makes all the replacements to this script public API. for example if we wanted to drop xctestrunner here, which was recently discussed in this slack thread https://bazelbuild.slack.com/archives/CD3QY5C2X/p1656459247671289, we would then break your downstream test runner script. we could handle this like normal public API changes and try to do semver bumps etc, but that mightbe a pain for you still
2. the second you need changes to the bzl file to add a new argument or something for replacements, I don't see an easily generalizable way to do that? It looks like you have some differences in the bzl file now, but the ones you have so far look compatible here, but do you foresee others?

For reference here's some previous discussion from when google owned the rules about the same case https://github.com/bazelbuild/rules_apple/pull/410

jerrymarino:
> for example if we wanted to drop xctestrunner here, which was recently discussed in this slack thread https://bazelbuild.slack.com/archives/CD3QY5C2X/p1656459247671289, we would then break your downstream test runner script.

In my useage of this code, I'm using `xctestrunner` as an implementation detail: I found this doesn't really care what `xctestrunner` is - in my case it is a VM based runner. Concretely, `xctestrunner` is wrapped inside of [another layer](https://github.com/bazel-ios/rules_ios/pull/509/files#diff-9fde3df203a16ce0b2e5a6dab58dac4ddfa4e73228f62e6d77a8d25566dded08R21). I am still using this runner, [but in layers away](https://github.com/bazel-ios/rules_ios/pull/509/files#diff-f419243332dcbde98e619a59adba8701826d3a5daa00a2649de139dde54d377cR6). Specifically it should be the same file:

> we would then break your downstream test runner

Well rules_ios having to make a copy of this doesn't make it any easier to maintain - for rules_ios 🙃 .

I think the issue with the code here is it's a single bash script, so people have to copy a decent chunk of code to make improvements or change it: the "testrunner" does a lot of stuff and also mixes concerns like code coverage and global simulator reusing. It looks like a lot of this needs fundamental changes if people want to extend it.

jerrymarino:
Thanks for taking the time out of your day to look at this @keith 🙇‍♂️  - going to close this out for now.

keith:
Having thought about this some more I think we should merge this as is, with the understanding that it's not really public API, in that it's more likely to break than standard public API, but it's probably still nicer than duplication for super simple cases. I do assume that your case might be complex enough that over time you'll want to modify the bzl file more, at which point I think you'll have to fork it.

acecilia:
Linking here a very similar PR for extra context: https://github.com/bazelbuild/rules_apple/pull/777

keith:
yea if you want to reopen that one instead that's fine

acecilia:
I agree with your initial rationale, in which that current code is very tightly coupled with xctestrunner binary, and if implementing a custom runner you will quickly need to fork existing code and modify it

PR #1590, opened 2022-07-20
Add sanitizer support for iOS tests

keith:
Previously the tests would crash on launch because the sanitizer
libraries have to be injected early with DYLD_INSERT_LIBRARIES


segiddins (approved):
Nice! Does it make sense to get a test in that uses a sanitizer?

keith:
yea definitely

keith:
pushed a test

PR #1612, opened 2022-08-05
Add xctestrun based iOS unit test runner

keith:
This test runner is a drop in replacement for the default iOS test
runner that you can use by setting it on the `runner` attribute of an
`ios_unit_test`. It uses the `xctest` CLI directly for logic tests, and
uses `xcodebuild test-without-building` with a `xctestrun` file for
hosted tests. It supports running tests in random order, and generating
`xcresult` bundles from any type of test. It also supports running
x86_64 test bundles on arm64 hosts. Examples can be found in the docs.

### DIFF
--- a/apple/testing/default_runner/ios_xctestrun_runner.bzl
+++ b/apple/testing/default_runner/ios_xctestrun_runner.bzl
@@ -34,7 +34,7 @@ def _ios_xctestrun_runner_impl(ctx):
     os_version = str(ctx.attr.os_version or ctx.fragments.objc.ios_simulator_version or
                      ctx.attr._xcode_config[apple_common.XcodeProperties].default_ios_sdk_version)
 
-    # TODO: Ideally we would be smarter about picking a device, but we don't know what the versoin of Xcode supports
+    # TODO: Ideally we would be smarter about picking a device, but we don't know what the current version of Xcode supports
     device_type = ctx.attr.device_type or ctx.fragments.objc.ios_simulator_device or "iPhone 12"
 
     if not os_version:
--- a/apple/testing/default_runner/ios_xctestrun_runner.bzl
+++ b/apple/testing/default_runner/ios_xctestrun_runner.bzl
@@ -1,5 +1,5 @@
 """
-A iOS test runner rule that uses xctestrun files to run unit test bundles on
+An iOS test runner rule that uses xctestrun files to run unit test bundles on
 simulators. This rule currently doesn't support UI tests or running on device.
 """
 
--- a/apple/testing/default_runner/ios_xctestrun_runner.bzl
+++ b/apple/testing/default_runner/ios_xctestrun_runner.bzl
@@ -1,0 +1,163 @@
+"""
+A iOS test runner rule that uses xctestrun files to run unit test bundles on
+simulators. This rule currently doesn't support UI tests or running on device.
+"""
+
+load("@build_bazel_rules_apple//apple/testing:apple_test_rules.bzl", "AppleTestRunnerInfo")
+
+def _get_template_substitutions(
+        *,
+        device_type,
+        os_version,
+        simulator_creator,
+        random,
+        xctestrun_template):
+    substitutions = {
+        "device_type": device_type,
+        "os_version": os_version,
+        "simulator_creator.py": simulator_creator,
+        # "ordered" isn't a special string, but anything besides "random" for this field runs in order
+        "test_order": "random" if random else "ordered",
+        "xctestrun_template": xctestrun_template,
+    }
+
+    return {"%({})s".format(key): value for key, value in substitutions.items()}
+
+def _get_execution_environment(ctx):
+    xcode_version = str(ctx.attr._xcode_config[apple_common.XcodeVersionConfig].xcode_version())
+    if not xcode_version:
+        fail("error: No xcode_version in _xcode_config")
+
+    return {"XCODE_VERSION_OVERRIDE": xcode_version}
+
+def _ios_xctestrun_runner_impl(ctx):
+    os_version = str(ctx.attr.os_version or ctx.fragments.objc.ios_simulator_version or
+                     ctx.attr._xcode_config[apple_common.XcodeProperties].default_ios_sdk_version)
+
+    # TODO: Ideally we would be smarter about picking a device, but we don't know what the versoin of Xcode supports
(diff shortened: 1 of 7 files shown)


apple/testing/default_runner/ios_xctestrun_runner.bzl
```diff
@@ -0,0 +1,163 @@
+"""
+A iOS test runner rule that uses xctestrun files to run unit test bundles on
```
brentleyjones, line 2:
```suggestion
An iOS test runner rule that uses xctestrun files to run unit test bundles on
```

apple/testing/default_runner/ios_xctestrun_runner.bzl
```diff
@@ -0,0 +1,163 @@
...
+        os_version,
+        simulator_creator,
+        random,
+        xctestrun_template):
+    substitutions = {
+        "device_type": device_type,
+        "os_version": os_version,
+        "simulator_creator.py": simulator_creator,
+        # "ordered" isn't a special string, but anything besides "random" for this field runs in order
+        "test_order": "random" if random else "ordered",
```
brentleyjones, line 20: Would it be better to be a bool then?

apple/testing/default_runner/ios_xctestrun_runner.bzl
```diff
@@ -0,0 +1,163 @@
...
+    if not xcode_version:
+        fail("error: No xcode_version in _xcode_config")
+
+    return {"XCODE_VERSION_OVERRIDE": xcode_version}
+
+def _ios_xctestrun_runner_impl(ctx):
+    os_version = str(ctx.attr.os_version or ctx.fragments.objc.ios_simulator_version or
+                     ctx.attr._xcode_config[apple_common.XcodeProperties].default_ios_sdk_version)
+
+    # TODO: Ideally we would be smarter about picking a device, but we don't know what the versoin of Xcode supports
```
brentleyjones, line 37:
```suggestion
    # TODO: Ideally we would be smarter about picking a device, but we don't know what the version Xcode supports
```

apple/testing/default_runner/ios_xctestrun_runner.bzl
```diff
@@ -0,0 +1,163 @@
...
+    device_type = ctx.attr.device_type or ctx.fragments.objc.ios_simulator_device or "iPhone 12"
+
+    if not os_version:
+        fail("error: os_version must be set on ios_xctestrun_runner, or passed with --ios_simulator_version")
+    if not device_type:
+        fail("error: device_type must be set on ios_xctestrun_runner, or passed with --ios_simulator_device")
```
brentleyjones, line 43: Do we still want the "iPhone 12" default?

brentleyjones: Seems both of these checks can't currently be hit.

apple/testing/default_runner/ios_xctestrun_runner.bzl
maxwellE:
```suggestion
    # TODO: Ideally we would be smarter about picking a device, but we don't know what the version of Xcode supports
```

apple/testing/default_runner/ios_xctestrun_runner.template.sh
```diff
@@ -0,0 +1,254 @@
...
+
+  xcodebuild test-without-building "${args[@]}" \
+    2>&1 | tee -i "$testlog" | (grep -v "One of the two will be used" || true) \
+    || test_exit_code=$?
+else
+  platform_developer_dir="$(xcode-select -p)/Platforms/iPhoneSimulator.platform/Developer"
+  xctest_binary="$platform_developer_dir/Library/Xcode/Agents/xctest"
+  test_file=$(file "$test_tmp_dir/$test_bundle_name.xctest/$test_bundle_name")
+  if [[ "$intel_simulator_hack" == true ]]; then
+    sliced_xctest_binary=/tmp/xctest_intel
```
brentleyjones, line 151: Can this be in the output_base instead, similar to global module caches? Seems odd to write outside of it.

apple/testing/default_runner/ios_xctestrun_runner.bzl
keith: from the user perspective it is a bool

apple/testing/default_runner/ios_xctestrun_runner.bzl
keith:
```suggestion
    # TODO: Ideally we would be smarter about picking a device, but we don't know what the current version of Xcode supports
```

apple/testing/default_runner/ios_xctestrun_runner.bzl
keith: we have to have some default, i picked one that should be supported by xcode 12 and 13+. the os_version check supposedly can, although i think that could only happen if you provided your own, potentially invalid xcode version config file. the device_type one can't, but I figured we should leave it in case we break it?

apple/testing/default_runner/ios_xctestrun_runner.template.sh
```diff
@@ -148,11 +146,8 @@ else
   xctest_binary="$platform_developer_dir/Library/Xcode/Agents/xctest"
   test_file=$(file "$test_tmp_dir/$test_bundle_name.xctest/$test_bundle_name")
   if [[ "$intel_simulator_hack" == true ]]; then
-    sliced_xctest_binary=/tmp/xctest_intel
-    if [[ ! -x "$sliced_xctest_binary" ]]; then
-      lipo -thin x86_64 -output "$sliced_xctest_binary" "$xctest_binary"
-    fi
-
+    sliced_xctest_binary="$test_tmp_dir/xctest_intel_bin"
```
brentleyjones, line 150: Is `test_tmp_dir` per test? How expensive is the slicing?

keith: correct.

```
% time lipo -thin x86_64 -output /tmp/slice $(xcode-select -p)/Platforms/iPhoneSimulator.platform/Developer/Library/Xcode/Agents/xctest
lipo -thin x86_64 -output /tmp/slice   0.00s user 0.00s system 53% cpu 0.009 total
```

brentleyjones (approved):

PR #1618, opened 2022-08-11
Add doc for provisioning_profile for ios tests

### LINKED ISSUE #1617: Add documentation for `provisioning_profile` attr of `ios_unit_test`
For an `ios_unit_test` to run on a real device, the `provisioning_profile` needs to be set. However, the documentation of `ios_unit_test` doesn't mention this. Some users may not know about this. https://github.com/bazelbuild/tulsi/issues/266 . 

If it's ok, I'd like to create a PR. But I don't know if any other attrs need to be mentioned. For our project, the `provisioning_profile` is the only one required.

kkpattern:
close #1617 

Rendered result [here](https://github.com/kkpattern/rules_apple/blob/doc_ios_unit_test_provisioning_profile/doc/rules-ios.md#ios_unit_test).

I also added documentation for `ios_ui_test`. I assume it also requires `provisioning_profile` to be set to run on a real device. However, we don't have UI tests, so I can't verify. If this is not the case, I will remove the `provisioning_profile` doc for `ios_ui_test`.

### DIFF
--- a/apple/internal/testing/ios_rules.bzl
+++ b/apple/internal/testing/ios_rules.bzl
@@ -79,6 +79,8 @@ provided test runner when invoked with `bazel test`. When using Tulsi to run
 tests built with this target, `runner` will not be used since Xcode is the test
 runner in that case.
 
+The `provisioning_profile` attribute needs to be set to run the test on a real device.
+
 To run the same test on multiple simulators/devices see
 [ios_ui_test_suite](#ios_ui_test_suite).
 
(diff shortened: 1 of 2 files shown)


keith:
ah so CI is failing here because this file is generated, the problem is that the actual test function called is this macro https://github.com/bazelbuild/rules_apple/blob/c655f5249ea161c1ae6c3892666c33bfcc080c65/apple/ios.bzl#L58, but we generated docs for the underlying test rule https://github.com/bazelbuild/rules_apple/blob/c655f5249ea161c1ae6c3892666c33bfcc080c65/apple/ios.doc.bzl#L55 so that it gets _some_ valid attributes, but it's not getting all for them. I'm not sure what the right move is here, ideally [stardoc](https://github.com/bazelbuild/stardoc) would allow something for this kind of case 🤔

brentleyjones:
Having the macro and and the rule in different bzl files, but have them named the same (with a name override on load into the macro file) works around this.

kkpattern:
Maybe I can mention the provising profile in the rule [doc](https://github.com/bazelbuild/rules_apple/blob/c655f5249ea161c1ae6c3892666c33bfcc080c65/apple/internal/testing/ios_rules.bzl#L105) instead?

keith:
Yea I think that's a fine workaround

kkpattern:
I've updated the doc.

thii (approved):

BalestraPatrick (approved):

keith:
Thanks!

Issue #1625, opened 2022-08-18
watchOS extensions are linked with incorrect `-e _WKExtensionMain` flag


BalestraPatrick:
When building an app with a Watch app which bundles an extensions in its watch extension (such as a watchOS intents or notification service extension), the linking invocation passes `_WKExtensionMain` for everything that is compiled for watchOS. This seems to be incorrect because App Store Connect rejects such bundles with the following error:
```
ITMS-90898: Invalid Mach-O entry point - Please make sure the build system passes “-e _NSExtensionMain” to the linker for the “App.app/Watch/WatchApp.app/PlugIns/WatchAppExtension.appex/PlugIns/WatchIntentsExtension.appex” extension bundle, or the extension will not function. Please refer to the App Extension Programming Guide at https://developer.apple.com/library/content/documentation/General/Conceptual/ExtensibilityPG/Action.html#//apple_ref/doc/uid/TP40014214-CH13-SW1
```

When I compare the linking invocation with Xcode, the same watchOS intents extension is linked with `-e _NSExtensionMain`. The watchOS extension (which bundles in its `PlugIns` directory the watchOS intents extension) should still get `-e _WKExtensionMain`. We should mirror this behavior in our rules as well.

I dove into the code a bit and here are my findings:
- We could add `-e _NSExtensionMain` for `watchos_extension` but that would add it to all types of extensions, which is not what Xcode does: https://github.com/bazelbuild/rules_apple/blob/d7b8f7d45c9bb0533aee2ed7410d111a0d060f71/apple/internal/rule_support.bzl#L750-L752
- The watchOS logic instead of living in `rule_support.bzl` (like other extensions have), lives in `watchos_rules.bzl`. This is likely because this feature used to be available only in Xcode 11+, so we need to disable it for previous Xcode versions. We can probably move this to `rule_support.bzl` if we think it makes sense. https://github.com/bazelbuild/rules_apple/blob/d7b8f7d45c9bb0533aee2ed7410d111a0d060f71/apple/internal/watchos_rules.bzl#L590-L614

So my question is the following: How do we allow linking with `-e _NSExtensionMain` watchOS extensions but maintain the usage of `_WKExtensionMain` for the top-level watchOS extension that ships in the watch application?

BalestraPatrick:
I tried to work around this by passing `-e _NSExtensionMain` to my watchOS intents extensions definition but that doesn't work because the rules always pass `-e _WKExtensionMain` anyway which is what ends up in the binary.

PR #1631, opened 2022-08-22
Add support for watchOS App Extensions

BalestraPatrick:
Fixes https://github.com/bazelbuild/rules_apple/issues/1625.

Previously, `watchos_extension` was supposed to support both WatchKit Extensions (the actual watch app extension) and  watchOS application extensions (such as intents and notification extensions). They were both treated the same, but they actually aren't. Xcode uses the `com.apple.product-type.app-extension` product type for application extensions even if they're targeting watchOS. On top of that, watchOS application extensions should get their entry point set to `_NSExtensionMain` instead of `_WKExtensionMain` otherwise they'll be rejected by App Store Connect during bundle validation.


apple/internal/watchos_rules.bzl
```diff
@@ -592,7 +592,11 @@ def _watchos_extension_impl(ctx):
     # TODO(min(Xcode) >= 11): Make this unconditional when the minimum supported Xcode is Xcode 11.
     xcode_config = ctx.attr._xcode_config[apple_common.XcodeVersionConfig]
     if xcode_support.is_xcode_at_least_version(xcode_config, "11"):
-        extra_linkopts = ["-e", "_WKExtensionMain"]
+        # watchOS application extensions have a different entry point.
+        if ctx.attr.application_extension:
```
brentleyjones, line 596: We also need to update all `product_type` references in here to use `apple_product_type.app_extension` instead of `rule_descriptor.product_type` when this is set.

test/starlark_tests/watchos_extension_tests.bzl
```diff
@@ -196,6 +197,22 @@ def watchos_extension_test_suite(name):
         tags = [name],
     )
 
+    entry_point_test(
```
brentleyjones, line 200: Can you also add a test for the `product_type`?

apple/internal/watchos_rules.bzl
```diff
@@ -644,7 +651,7 @@ def _watchos_extension_impl(ctx):
             label_name = label.name,
             platform_prerequisites = platform_prerequisites,
             predeclared_outputs = predeclared_outputs,
-            product_type = rule_descriptor.product_type,
+            product_type = product_type,
```
brentleyjones, line 654: Also needed above in `process_entitlements`?

BalestraPatrick: Yes indeed!

apple/internal/rule_factory.bzl
```diff
@@ -959,12 +959,23 @@ def _get_watchos_attrs(rule_descriptor):
...
+                providers = [[AppleBundleInfo, WatchosExtensionBundleInfo]],
+                doc = """
 A list of watchOS application extensions to include in the final watch extension bundle.
 """,
-        )})
+            ),
+            "application_extension": attr.bool(
+                default = False,
+                doc = """
+If `true`, this extension is an App Extension instead of a WatchKit Extension.
```
brentleyjones, line 972: I mixed up Swift and Starlark:

```suggestion
If `True`, this extension is an App Extension instead of a WatchKit Extension.
```

brentleyjones (approved):

PR #1643, opened 2022-08-31
Lazy hash compute in bundletool.py

### LINKED ISSUE #1641: Lazy hash computing in bundletool.py
Currently, if `tree artifacts` is not enabled, `bundletool.py` will be used to bundle an iOS app. In `bundletool.py`, before writing each entry into the zip file [the md5 hash of the data will be computed](https://github.com/bazelbuild/rules_apple/blob/master/tools/bundletool/bundletool.py#L196) to check if different files will be written into the same `dest`. If we have large files in the bundle(for us, we have about 2GB of resource files), the hash computing will take some time to finish. However, in most cases, there won't be conflicting files, so time is wasted.

A simple fix is to lazy compute the md5 hash of each entry. Instead of `self._entry_hashes` we can have `self._entry_datas`. Then before we write an entry into the zip file, we check if there are already data for the same `dest`. If it's true, then we calculate the hash value of the two entries to check if they are the same.

We did a simple test with this fix. Before the fix, it cost around 40 seconds to finish the bundling. After it cost around 22 seconds.

I can think about two downsides of this method:

1. memory consumption of bundling will be increased.
2. for cases where the same file will be written multiple times into the zip file, lazy hash computing may actually take more time.

If the lazy hash computing is acceptable, I'd like to create a PR. If there is a better solution, I'd be happy to try it too.

kkpattern:
To prevent multiple files with different content be written into the same dest,
bundletool.py compare files by their hash value before writing. But in most
cases, there are no conflicting files, and the time used to calculate the hash
is wasted. This commit defers the calculation of the hash until there are
actually two files that need to be written to the same dest.

I also added a test case to check if empty files are correctly compared with other files.

close #1641

### DIFF
--- a/tools/bundletool/bundletool.py
+++ b/tools/bundletool/bundletool.py
@@ -92,7 +92,7 @@ class Bundler(object):
 
     # Keep track of hashes of each entry; this will be faster than pulling the
     # data back out of the archive as it's written.
-    self._entry_hashes = {}
+    self._entry_datas = {}
 
   def run(self):
     """Performs the operations requested by the control struct."""
@@ -193,14 +193,13 @@ class Bundler(object):
       BundleToolError: If two files with different content would be placed
           at the same location in the ZIP file.
     """
-    new_hash = hashlib.md5(data).digest()
-    existing_hash = self._entry_hashes.get(dest)
-    if existing_hash:
-      if existing_hash == new_hash:
+    existing_data = self._entry_datas.get(dest, None)
+    if existing_data is not None:
+      if hashlib.md5(existing_data).digest() == hashlib.md5(data).digest():
         return
       raise BundleToolError(BUNDLE_CONFLICT_MSG_TEMPLATE % dest)
 
-    self._entry_hashes[dest] = new_hash
+    self._entry_datas[dest] = data
 
     zipinfo = zipfile.ZipInfo(dest)
     if compress:
--- a/tools/bundletool/bundletool_unittest.py
+++ b/tools/bundletool/bundletool_unittest.py
@@ -287,6 +287,35 @@ class BundlerTest(unittest.TestCase):
           ]
       })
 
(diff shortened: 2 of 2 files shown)


tools/bundletool/bundletool.py
```diff
@@ -193,14 +193,13 @@ def _write_entry(self, dest, data, executable, out_zip, compress):
       BundleToolError: If two files with different content would be placed
           at the same location in the ZIP file.
     """
-    new_hash = hashlib.md5(data).digest()
-    existing_hash = self._entry_hashes.get(dest)
-    if existing_hash:
-      if existing_hash == new_hash:
+    existing_data = self._entry_datas.get(dest, None)
+    if existing_data is not None:
+      if hashlib.md5(existing_data).digest() == hashlib.md5(data).digest():
```
thomasvl, line 198: drive by comments -

The hashes were all about reducing the amount of data in memory; once you are keeping the data just compare the data and never hash. The hash is just wasted work compared to a bytes compare.

kkpattern: Yeah, the memory usage is the main downside I can think of. Another way to do this is to add an additional pass to iterate all the entries that need to be written first, and find if there are multiple files sharing the same dest. I can give it a try.

keith: what if we just stored the paths that have already been written, and lazily do `out_zip.read(path)` to get the data if the path already exists, would that work? also at that point if you were worried about it we could cache only the data from the conflicts in memory, but might not even be worth it

kkpattern: I actually tried the additional pass solution. First, count which files have the same destination path, and then only calculate the hash for these files. I did a quick dirty implementation, and it worked. But I'm busy setting up buildfarm for our project, so I haven't got time to sort out the code and commit it. I will try to upload it within the next week.

PR #1668, opened 2022-09-21
Add test_host_is_bundle_loader configuration

keith:
The default behavior for a test that has a test_host is to use that test
host app as the `-bundle_loader` linker argument. This is required if
you want to test APIs from the host app, but in the case you want to use
a host for other reasons, like testing iOS features that require you to
run in an app, you might not want this. This also works around some
surprising behavior where linkopts from the app are applied to the test
bundle as well.

### DIFF
--- a/apple/internal/rule_attrs.bzl
+++ b/apple/internal/rule_attrs.bzl
@@ -404,6 +404,13 @@ def _test_host_attrs(
             mandatory = is_mandatory,
             providers = providers,
         ),
+        "test_host_is_bundle_loader": attr.bool(
+            default = True,
+            doc = """
+Whether the 'test_host' should be used as the -bundle_loader to allow testing
+the symbols from the test host app
+""",
+        ),
     }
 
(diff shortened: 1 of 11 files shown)


keith:
I only added this to iOS right now, figured we could add as needed to other rules ¯\_(ツ)_/¯

brentleyjones (approved):

brentleyjones:
Feels a little odd to not have it on all the ones that have a test host.

keith:
filed https://github.com/bazelbuild/bazel/issues/16327

keith:
yea i agree, im torn between it's overhead that might not be used and we should add it as needed, vs that's weird. definitely also being lazy about adding more test targets 🙃

brentleyjones:
I think the overhead is next to nothing, so it being weird probably outweighs it.

keith:
now that all the test rule's attributes are unified i think we're good with the tests we have here

PR #1671, opened 2022-09-22
Fixes an issue with plist keys and alt icons

maxwellE:
Fixes a potential issue where certain
plists would cause this logic to crash
due to a `KeyError`


brentleyjones:
Is it valid to have alternate icons like this without the root icons? Are we just holding it wrong over in our rules_xcodeproj example? (I don't have an app with alternate icons, so I'm not sure how this feature normally works.)

maxwellE:
> Is it valid to have alternate icons like this without the root icons? Are we just holding it wrong over in our rules_xcodeproj example? (I don't have an app with alternate icons, so I'm not sure how this feature normally works.)

Screen Shot 2022-09-22 at 9 22 31 AM

Even after adding these keys we get errors. I think that this change makes it easier to use alternate icons out of the box with a default info.plist

PR #1672, opened 2022-09-22
Test not specifying bazel

keith:


keith:
going to hope for https://github.com/bazelbuild/continuous-integration/issues/1455 instead of messing with this

PR #1683, opened 2022-09-29
Add `bundle_only` support everywhere

brentleyjones:


keith (approved):
Probably we only didn't do this before because of adding tests or something for all variants. We should try to make that easier somehow

PR #1685, opened 2022-09-29
Add `apple_symbols_file_partial` to more rules

brentleyjones:


brentleyjones:
@thii Are all of these correct, and do the have the correct `dependency_targets`?

thii:
These are only needed for App Store, so `*_application`s only.

brentleyjones:
But it was already on ios_framework and ios_extension. Don't we need to collect symbols for those as well?

thii:
Ah yes, `*_application`s, and anything that can be bundled with them, but not framework bundling rules for third-party distribution like `ios_dynamic_framework`.

brentleyjones:
But those rules can be used in place of `*_framework` (by putting them in the `frameworks` attribute), so I think they apply as well now.

As an aside, I think we need to either make them only for distribution, or make a single rule that does both. But that's for a different time.

thii (approved):

brentleyjones:
Thinking more about this, I'll revert the ones on the distribution rules. I agree that is what they should be used for, so let's not conflate them even more right now. Later if we decide to sanction then as more than that, we can add to them.

PR #1688, opened 2022-09-29
Fix `watchos_static_framework` for arm64 simulator

brentleyjones:


test/starlark_tests/watchos_static_framework_tests.bzl
```diff
@@ -49,6 +49,34 @@ def watchos_static_framework_test_suite(name):
...
+        name = "{}_swift_arm64_builds".format(name),
+        build_type = "simulator",
+        target_under_test = "//test/starlark_tests/targets_under_test/watchos:swift_static_fmwk",
+        cpus = {
+            "watchos_cpus": ["x86_64", "arm64"],
+        },
+        binary_test_file = "$BUNDLE_ROOT/swift_static_fmwk",
+        binary_test_architecture = "arm64",
+        macho_load_commands_contain = ["cmd LC_BUILD_VERSION", "platform WATCHOSSIMULATOR"],
+        macho_load_commands_not_contain = ["cmd LC_VERSION_MIN_WATCHOS"],
```
keith, line 63: is there a `platform` line that should be here too? I feel like I saw somewhere that all platforms should be using `LC_BUILD_VERSION` moving forward in which case this one wouldn't be here ever, even if this test was incorrect

apple/internal/transition_support.bzl
```diff
@@ -256,10 +256,10 @@ def _xcframework_split_attr_key(*, cpu, environment, platform_type):
         platform_type = platform_type,
     ) + "_" + environment
 
-def _resolved_cpu_for_cpu(*, cpu, environment):
+def _resolved_cpu_for_cpu(*, cpu, environment, platform_type):
     # TODO(b/180572694): Remove cpu redirection after supporting platforms based toolchain
     # resolution.
-    if cpu == "arm64" and environment == "simulator":
+    if cpu == "arm64" and environment == "simulator" and platform_type != "watchos":
```
keith, line 262: want to link https://github.com/bazelbuild/bazel/pull/16181 ?

test/starlark_tests/watchos_static_framework_tests.bzl
brentleyjones: I copied the tvOS and iOS tests. I think they are just confirming the difference between arm64 and non-arm64. If we want to change these, we should probably change all of them?

apple/internal/transition_support.bzl
brentleyjones: Yeah, since that will require this to be reverted.

test/starlark_tests/watchos_static_framework_tests.bzl
brentleyjones: I would just be fumbling, since I don't know what these are. Can you/someone make a PR to change them all together after this one panda?

keith (approved):

PR #1695, opened 2022-10-07
Remove repo name from module import

BalestraPatrick:
This is in preparation for adding support for bzlmod. This Python script loads another file via its runfiles. When bzlmod is enabled, the directory has the name of the actual dependency and its version and not the name of its defined `repo_name` (`rules_apple~1.1.2` instead of `build_bazel_rules_apple`). This causes issues and should be addressed in Bazel (which it is being thought about per [this Slack conversation](https://bazelbuild.slack.com/archives/C014RARENH0/p1664279459175029)), but this is a quick fix that should be backwards compatible.

If this passes CI and there are no objections, I'll remove all instances of this practice from the repo.

### DIFF
--- a/tools/bitcode_strip/bitcode_strip.py
+++ b/tools/bitcode_strip/bitcode_strip.py
@@ -13,7 +13,7 @@
 # limitations under the License.
 #
 
-from build_bazel_rules_apple.tools.wrapper_common import execute
+from tools.wrapper_common import execute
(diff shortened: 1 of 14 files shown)


brentleyjones (approved):

keith (approved):
If we don't need them I don't see any reason to keep them!

PR #1698, opened 2022-10-10
Revert "[rules_apple] Migrate static_framework_file to imported_library (#1697)"

keith:
This doesn't actually work, these libraries end up not being linked.

This reverts commit 8d841342c238457896cd7596cc29b2d06c9a75f0.


brentleyjones (approved):

keith:
actually seems like we just need https://github.com/bazelbuild/rules_swift/pull/917 at the same time

keith:
not going to revert, just folks have to update in lockstep like normal, we'll do a new rules_swift release before a release here anyways

Issue #1702, opened 2022-10-12
[Xcode 14] `ios_sticker_pack_extension` generates stub binary with bitcode


mattrobmattrob:
AppStore Connect has issues with an Xcode 14.0.1 build because of bitcode in our stickers stub binary with `build --apple_bitcode=none` set in `.bazelrc`.
```
ITMS-90482: Invalid Executable - The executable 'FooApp.app/PlugIns/stickers.appex/stickers' contains bitcode.
```

Confirmed locally via :point_down: based on [this post](https://stackoverflow.com/a/33105733):
```
$ otool -l /private/var/tmp/_bazel_matt.robinson/c07525d928605c4e102afeec7f88db9c/execroot/repo/bazel-out/applebin_ios-ios_arm64-opt-ST-1eb592bc8226/bin/FooApp/FooApp_archive-root/Payload/FooApp.app/PlugIns/stickers.appex/stickers | grep __LLVM
  segname __LLVM
   segname __LLVM
  segname __LLVM
   segname __LLVM
  segname __LLVM
   segname __LLVM
```

keith:
https://github.com/bazelbuild/rules_apple/pull/1703

PR #1703, opened 2022-10-12
Strip bitcode from stub binaries

keith:
Xcode 14 doesn't allow uploading any binaries with bitcode, but the
stub binaries it provides for use with message extensions still contain
bitcode. This change strips that based on the bitcode mode.

Fixes https://github.com/bazelbuild/rules_apple/issues/1702


brentleyjones (approved):

mattrobmattrob (approved):
Confirmed this worked in my local reproduction case of https://github.com/bazelbuild/rules_apple/issues/1702. 👍

PR #1706, opened 2022-10-12
Embed frameworks in `ios_message_extension`

brentleyjones:
I noticed that in BwX mode of rules_xcodeproj an iOS Message Extension would correctly embed frameworks. When running in BwB mode it crashed at runtime. This change fixes that.


keith:
shouldn't these be bundled with the top level app instead and the extension have an rpath to its framework directory? to avoid duplication

brentleyjones:
If the top level is the message app stub, no. I guess we need to detect that somehow?

PR #1707, opened 2022-10-13
Fix sdk_framework discovery with other argument format

keith:
This changed with https://github.com/bazelbuild/bazel/commit/977d7fb3837ceeabd17ecd639037f95a99ca4cb1 causing some modulemap tests to fail


keith:
FYI @googlewalt since this logic is the same on the upstream branch, although this case might not matter there

thii (approved):

googlewalt:
Thanks.  I'll patch it internally.

PR #1716, opened 2022-10-13
Raise version number to minimum of 12 or equivalent where possible on iOS/tvOS test targets for Starlark.

keith:
This is done to accomodate when arm64e was formally introduced for iOS and tvOS, and hopefully to address future test issues when building for this arch.

As armv7 will be dropped in Xcode 14 RC as a buildable architecture, we are using arm64e when possible to handle the representation of fat binaries for iOS devices.

PiperOrigin-RevId: 466809744
(cherry picked from commit 02b72e19839a8fa4bc5e526672df959d1f460c54)


keith:
i think i missed some external only tests, but it will be easier to find after https://github.com/bazelbuild/rules_apple//commit/3d622b6395d58124d32cf041a2f73a6a3a84987b

brentleyjones (approved):

keith:
https://github.com/bazelbuild/rules_apple/pull/1718

PR #1717, opened 2022-10-13
Raise the minimum_os_versions of the shell tests to versions supported by Xcode 14 RC and the arm64e architecture for iOS and tvOS.

keith:
Since arm64e requires iOS 12 as a minimum, we use that as a guideline for raising the minimum_os_version of all ios_* tests.

For the remainder, we rely on supported deployment targets from https://developer.apple.com/support/xcode/.

PiperOrigin-RevId: 467315520
(cherry picked from commit 96a5fa31666e3d91118452717f135f615457dfc9)


keith:
https://github.com/bazelbuild/rules_apple/pull/1718

PR #1718, opened 2022-10-13
Cherry pick min OS maintenance changes

keith:


keith:
4 changes combined here since there was a lot of churn on the same kinda stuff

PR #1727, opened 2022-10-13
Handle transforming an entitlements.plist file into a DER representation

keith:
The upstream commits here were a bit of a roller coaster because there was a new tool that was added and then removed. I cherry picked them all anyways to be sure. I will squash this though


keith:
i think we were just blocked here because of an xcode 14 change so we should be good on CI for that now

brentleyjones (approved):

PR #1729, opened 2022-10-13
Fix execute's inputstr argument, which was removed when the timeout argument was introduced.

keith:
PiperOrigin-RevId: 471827739
(cherry picked from commit 2c54e9520b545937791204559a39a80cf7fc352a)


keith:
we had this for a while, but upstream they added a test

brentleyjones (approved):

PR #1746, opened 2022-11-03
Remove subpar dependency

keith:
We haven't used this since we moved to using xctestrunner directly as
python files


BalestraPatrick (approved):

keith:
blah. fails because xctestrunner has this in the BUILD file we reference even though we don't use that target

keith:
oh actually we do still use the par binary.

brentleyjones (approved):

keith:
https://github.com/google/xctestrunner/pull/48

PR #1751, opened 2022-11-10
Allow for deps with only CcInfo on framework imports

dierksen:
I'm trying to add some [`swift_c_module` deps for a framework import](https://github.com/bazel-ios/rules_ios/pull/562), and they don't currently work since that only provides `CcInfo` and `SwiftInfo`.

It appears that all the uses for `apple_common.Objc` check to see if it exists, so that shouldn't be a hard requirement at this point. That said, we also would need to cherry-pick this onto 5.x, and the existence check would need to be added along with this change: https://github.com/bazelbuild/rules_apple/compare/bazel/5.x...dierksen:rules_apple:objc_provider_check?expand=1

### DIFF
--- a/apple/internal/apple_framework_import.bzl
+++ b/apple/internal/apple_framework_import.bzl
@@ -433,6 +433,7 @@ A list of targets that are dependencies of the target being built, which will be
 target.
 """,
                 providers = [
+                    [CcInfo],
                     [apple_common.Objc, CcInfo],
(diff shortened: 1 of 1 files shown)


dierksen:
I'm also not completely sure, style-wise, if all the various combinations of providers make sense here, or if only the lone `[CcInfo]` would make more sense, but this follows the existing pattern.

keith (approved):

PR #1770, opened 2022-11-28
Ignore skip_codesign_simulator_bundles for frameworks when building for device

BalestraPatrick:
We recently enabled `--features=apple.skip_codesign_simulator_bundles` and we just noticed that builds for the device now fail to install with the error: `No code signature found.`

We have a dynamic framework that is included in the app which is shared with extensions, and running `codesign --verify` on the framework gives this:
```
MyFramework.framework: code object is not signed at all
```
If we disable `--features=apple.skip_codesign_simulator_bundles` we get this instead:
```
MyFramework.framework: valid on disk
MyFramework.framework: satisfies its Designated Requirement
```

Going through the code, it seems like there's a missing case for checking if we're building for the device, we should still sign frameworks. As you can see at line 450, we were checking for that already, and now I added the same check at line 430. I tested it in our project and it works, but let me know if you think this should be fixed in some other way.

### DIFF
--- a/apple/internal/codesigning_support.bzl
+++ b/apple/internal/codesigning_support.bzl
@@ -419,6 +419,7 @@ def _codesigning_command(
         rule_descriptor = rule_descriptor,
     )
     paths_to_sign = []
+    is_device = platform_prerequisites.platform.is_device
 
     # The command returned by this function is executed as part of a bundling shell script.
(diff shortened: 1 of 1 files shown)


brentleyjones (approved):

keith (approved):

brentleyjones:
@BalestraPatrick Can you cherry-pick this into the 5.x branch as well?

PR #1782, opened 2022-12-02
Fix asset filename: rules_swift -> rules_apple

thii:


keith (approved):
Oof thanks. I knew I'd miss something like this

thii:
np, I've just found out by trying the new release. I think we should rename the uploaded file or update the code snippet?

thii:
I renamed the file.

keith:
Thanks!

PR #1785, opened 2022-12-09
Add product module name as an optional arg for the test rule

dostrander:
If there are modules under the tests that are named differently then the target itself it will cause issues with the xctestrun file. Specifically using Swift the test_filters do not work unless the module names line up with the test target name. Guessing because Swift's full class names are `MODULE.CLASS` so they do the magic under the hood.


keith:
can you throw a value for this attr into one of the test targets so that it actually gets caught by CI? also note that this has to hit master before this branch, we won't ever merge this branch back into master

dostrander:
> can you throw a value for this attr into one of the test targets so that it actually gets caught by CI? also note that this has to hit master before this branch, we won't ever merge this branch back into master

Will do.

Also open to possibly changing this name to `module_name` or something. I keep going back and forth in my head

apple/internal/testing/apple_test_rule_support.bzl
```diff
@@ -115,6 +115,7 @@ def _get_template_substitutions(test_type, test_bundle, test_environment, test_h
     subs["test_type"] = test_type.upper()
     subs["test_env"] = ",".join([k + "=" + v for (k, v) in test_environment.items()])
     subs["test_filter"] = test_filter or ""
+    subs["product_module_name"] = product_module_name or ""
```
jerrymarino, line 118: Doesn't the `test_filter` address each test by the `PRODUCT_MODULE_NAME`?  Perhaps can you parse this from `test_filter` if it it exists. The thing I liked about this is being able to filter different module names. Say you've got a number of modules in the test:

```
bazel test :MyTest --test_filters MyFirst.swiftmodule:foo
bazel test :MyTest --test_filters MySecond.swiftmodule:bar
```

Then - inside of `xctestrunner` - you'd simply set `PRODUCT_MODULE` in the xctestrun based on the filter - and be sure to get the right one. Probably validate they set that.. 

The downside of adding this at the rule level is you can't filter dynamically like that. 

PS : I wonder if this is a regression - or perhaps `test_filter` only ever workd for objc!

jerrymarino: Also do you imagine that we'd ever want to feed `.xctestrun` ( or partials ) into this rule? That might be another thing to contend with - similar our plist composition stuff in rules_ios you might be able to pass overrides.

I would still question if you can leverage the build graph too: assuming that you pack ( filterable ) tests into the first `.swiftmodule` file - you can set the `PRODUCT_MODULE_NAME` based on the first direct `SwiftInfo`

Obviously these are 2 different solutions. I liked the dep graph way - and might be a nice default - it has some inflexibility but it's probably easier for 99% of tests out there

test/ios_test_runner_unit_test.sh
```diff
@@ -209,6 +209,7 @@ objc_library(
 
 ios_unit_test(
     name = "PassingUnitTest",
+    module_name = "pass_unit_test_lib",
```
jerrymarino, line 212: I imagine we should remove this attribute now

apple/internal/testing/apple_test_rule_support.bzl
```diff
@@ -135,6 +141,11 @@ def _apple_test_rule_impl(ctx, test_type):
     test_bundle_target = ctx.attr.deps[0]
 
     test_bundle = test_bundle_target[AppleTestInfo].test_bundle
+    swift_module_list = test_bundle_target[AppleTestInfo].swift_modules.to_list()
+
+    test_swift_module_file = swift_module_list[len(swift_module_list) - 1]
```
jerrymarino, line 146: Should this do a lenght check - if there is less than 1 swiftmodule it's out of bounds?

apple/internal/testing/apple_test_rule_support.bzl
jerrymarino: Ah cool glad it works - this makes Bazel easier to use in general IMO

jerrymarino:
@dostrander bump - I think there's a few test failures that would prevent us from landing this re: https://github.com/bazel-ios/rules_ios/pull/663

dostrander:
> @dostrander bump - I think there's a few test failures that would prevent us from landing this re: [bazel-ios/rules_ios#663](https://github.com/bazel-ios/rules_ios/pull/663)

This can't realyl be merged until https://github.com/google/xctestrunner/pull/50 is merged unless we do a patch

PR #1786, opened 2022-12-14
Add test that ios_dynamic_framework doesn't include files in archive

keith:


keith:
this currently fails, but that means that if you build with a ios_dynamic_framework in your build graph (as opposed to distributing it for Xcode) you ship your app including the swiftmodule and other files. // @mccorkill1

mccorkill1:
I've got a fix for this locally and will put up a pr in the next day or two to address this. I'll tag you for review @keith

keith:
Thanks!

PR #1790, opened 2022-12-20
Add package_for_distribution flag to ios_dynamic_framework so it won't include files in archive

mccorkill1:
Adding package_for_distribution flag to ios_dynamic_framework so that swiftmodule and other files have the option to not be packaged in. Adding a test from https://github.com/bazelbuild/rules_apple/pull/1786 to confirm the new behavior.

### DIFF
--- a/apple/internal/ios_rules.bzl
+++ b/apple/internal/ios_rules.bzl
@@ -1331,13 +1331,16 @@ def _ios_dynamic_framework_impl(ctx):
             label_name = label.name,
             platform_prerequisites = platform_prerequisites,
         ),
-        partials.swift_dynamic_framework_partial(
-            actions = actions,
-            bundle_name = bundle_name,
-            label_name = label.name,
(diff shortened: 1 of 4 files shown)


keith:
this makes me rethink the value of allowing this to be a dependency in the first place, should we revisit folks just using ios_framework in that case, and this rule in this case without the need for more attrs?

mccorkill1:
The original intent of `ios_dynamic_framework` was to build a framework that could be consumed by Xcode (so really more of a packaging rule). This required that Modules and Headers folder to be generated.

The purpose of allowing it to be a dependency was to reduce the clutter in our build files so we wouldn't have to have a framework target for use as a dependency and an ios_dynamic_framework target for packaging. It was convenient to have it all in one.

With that being said, I didn't realize that it was including those folders in the archive when it is used by an app target, or that this was undesired behavior. I was able to get the tests passing with the additional logic from adding the new attribute, but it kind of took away the purpose of only having one target per framework since I had to build it once for distribution and once without to support this use case.

PR #1796, opened 2022-12-23
Add module extension to include non-bzlmod ready dependencies

BalestraPatrick:
This will remove a patch that we have in the BCR that is required to expose non-bzlmod ready dependencies via a module extension. This is similar to what we do in [rules_swift](https://github.com/bazelbuild/rules_swift/pull/920/) already.

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -8,4 +8,12 @@ module(
 
 bazel_dep(name = "bazel_skylib", version = "1.3.0")
 bazel_dep(name = "apple_support", repo_name = "build_bazel_apple_support", version = "1.3.2")
-bazel_dep(name = "rules_swift", repo_name = "build_bazel_rules_swift", version = "1.2.0")
+bazel_dep(name = "rules_swift", repo_name = "build_bazel_rules_swift", version = "1.5.0")
+
+non_module_deps = use_extension("//apple:extensions.bzl", "non_module_deps")
+
+use_repo(
+    non_module_deps,
+    "subpar",
+    "xctestrunner",
+)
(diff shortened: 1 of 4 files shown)


brentleyjones (approved):

MODULE.bazel
```diff
@@ -8,4 +8,12 @@ module(
...
-bazel_dep(name = "rules_swift", repo_name = "build_bazel_rules_swift", version = "1.2.0")
+bazel_dep(name = "rules_swift", repo_name = "build_bazel_rules_swift", version = "1.5.0")
+
+non_module_deps = use_extension("//apple:extensions.bzl", "non_module_deps")
+
+use_repo(
+    non_module_deps,
+    "subpar",
+    "xctestrunner",
+)
```
keith, line 19: Make sure the generator still works!

keith (approved):

MODULE.bazel
BalestraPatrick: Ah yes, good point. I had to tweak it a bit. It might be better to change it to simply change the line that contains `version = "x"` for `rules_apple` instead of regenerating the file  completely, or was there a speicific reason to do it that way?

keith: that's fine if you can come up with a reasonable grep or w/e for that, alternatively doing what rules_swift is doing works fine too w/o having to dup this https://github.com/bazelbuild/rules_swift/blob/7ea19709611d40776f05bf6d35b950e1d1a3a506/.github/update-module-version.sh#L18

BalestraPatrick: That probably works here too, will make the change in a follow-up then.

PR #1804, opened 2023-01-11
Test new C++ toolchain

keith:


keith:
looks good, head deps negate my change

PR #1810, opened 2023-01-19
Add override for startup_timeout_sec in tests

keith:
This way you can pass `--test_env=STARTUP_TIMEOUT_SEC=N`. The default is
the same as if you don't pass anything to xctestrunner


mattrobmattrob (approved):
Thanks, @keith!

luispadron:
Have had a need for this too, thank you!

thii (approved):

Issue #1756, opened 2022-11-17
Add mixed source library target


keith:
As discussed in the bazelcon office hours, we'd like to add a mixed source library target that we generally discourage, for all the performance downsides, but it helps people migrate

PR #1812, opened 2023-01-22
Add experimental_mixed_language_library macro to support building mixed Objective-C and Swift modules

thii:
This is supposed to help with the migration and in general is discouraged to use. The implementation is a simplified version of a community rule but doesn't support header maps and Clang modules.

Closes https://github.com/bazelbuild/rules_apple/issues/1756

### DIFF
--- a/apple/internal/apple_library.bzl
+++ b/apple/internal/apple_library.bzl
@@ -1,0 +1,309 @@
+"""apple_library macro implementation."""
+
+load(
+    "@build_bazel_rules_swift//swift:swift.bzl",
+    "SwiftInfo",
+    "swift_library",
+)
+load("@bazel_skylib//lib:paths.bzl", "paths")
+
+_CPP_FILE_TYPES = [".cc", ".cpp", ".mm", ".cxx", ".C"]
+
+_NON_CPP_FILE_TYPES = [".m", ".c"]
+
+_ASSEMBLY_FILE_TYPES = [".s", ".S", ".asm"]
+
+_OBJECT_FILE_FILE_TYPES = [".o"]
+
+_HEADERS_FILE_TYPES = [
+    ".h",
+    ".hh",
+    ".hpp",
+    ".ipp",
+    ".hxx",
+    ".h++",
+    ".inc",
+    ".inl",
+    ".tlh",
+    ".tli",
+    ".H",
+    ".hmap",
+]
+
+_OBJC_FILE_TYPES = _CPP_FILE_TYPES + \
+                   _NON_CPP_FILE_TYPES + \
+                   _ASSEMBLY_FILE_TYPES + \
+                   _OBJECT_FILE_FILE_TYPES + \
+                   _HEADERS_FILE_TYPES
+
+_SWIFT_FILE_TYPES = [".swift"]
+
+def _module_map_content(
+        module_name,
+        hdrs,
+        textual_hdrs,
+        swift_generated_header,
+        module_map_path):
+    # Up to the execution root
+    # bazel-out/<platform-config>/bin/<path/to/package>/<target-name>.modulemaps/<module-name>
+    slashes_count = module_map_path.count("/")
+    relative_path = "".join(["../"] * slashes_count)
+
+    content = "module " + module_name + " {\n"
+
+    for hdr in hdrs:
+        if hdr.extension == "h":
+            content += "  header \"%s%s\"\n" % (relative_path, hdr.path)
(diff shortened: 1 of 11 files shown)


apple/internal/apple_library.bzl
```diff
@@ -0,0 +1,309 @@
...
+    },
+    doc = "Generates a module map given a list of header files.",
+    implementation = _module_map_impl,
+    provides = [
+        DefaultInfo,
+    ],
+)
+
+# TODO: Document the apple_library macro
+def apple_library(**kwargs):
```
BalestraPatrick, line 192: You likely put quite some thought into this name, but I'm wondering if this could be confusing for folks thinking this is the default and non-discouraged rule?

brentleyjones: I agree. Can we rename it?

apple/internal/apple_library.bzl
```diff
@@ -0,0 +1,309 @@
...
+    native.objc_library(
+        copts = objc_copts,
+        deps = objc_deps,
+        hdrs = hdrs + [
+            # These aren't headers but here is the only place to declare these
+            # files as the inputs because objc_library doesn't have an
+            # attribute to declare custom inputs.
+            ":" + umbrella_module_map,
+        ],
+        module_map = umbrella_module_map,
```
brentleyjones, line 306: I _think_ you are accounting for the bug I fixed here: https://github.com/buildbuddy-io/rules_xcodeproj/commit/d40173d84764c1256b95a537b2871e010398f18c#diff-9e7ea7f032b7624ef92c4503d20ca20bf16eafd89f6277d71b83dd458972dc3d, but I just wanted to check and make sure.

thii: Yes, `objc_library`'s generated module map shouldn't be used.

apple/internal/apple_library.bzl
thii: I was thinking of mentioning that in the rule's documentation, but we could make it clearer by renaming it. Would `deprecated_apple_libarary` work?

brentleyjones: What about something that includes "mixed_language" in the name?

apple/internal/apple_library.bzl
```diff
@@ -0,0 +1,309 @@
...
+            doc = """\
+The list of C, C++, Objective-C, and Objective-C++ header files used to
+construct the module map. Unlike hdrs, these will be declared as 'textual
+header' in the module map.
+""",
+        ),
+        "deps": attr.label_list(
+            providers = [SwiftInfo],
+            doc = """\
+The list of swift_library targets.  A `${module_name}.Swift` submodule will be
```
brentleyjones, line 179: If we could change this to not physically depend on the swift_library; we could work around https://github.com/buildbuddy-io/rules_xcodeproj/blob/main/docs/faq.md#why-do-some-of-my-swift_librarys-compile-twice-in-bwx-mode. We could pass in any info needed for the generated header path.

Not a blocker, just a shower thought.

apple/internal/apple_library.bzl
thii: So far all the mixed language rules out there are using the `apple_library` name, including the Buck one, but `mixed_language_library` is nice too. Maybe `experimental_mixed_language_library`?

apple/internal/apple_library.bzl
thii: I think I've thought about this when first implemented the rule in the past, but don't remember the details. I can take a look again once this is merged.

apple/internal/apple_library.bzl
brentleyjones: That works.

apple/internal/experimental_mixed_language_library.bzl
```diff
@@ -188,16 +188,38 @@ generated if non-empty.
...
+        *,
+        name,
+        srcs,
+        deps = [],
+        module_name = None,
+        objc_copts = [],
+        swift_copts = [],
+        swiftc_inputs = [],
+        **kwargs):
     """Compiles and links Objective-C and Swift code into a static library.
```
brentleyjones, line 201: Can we add something here saying how's is discouraged and is mainly to aid people migrating? Not sure on the exact language, but something to explain why `experimental` is in the name.

brentleyjones (approved):

BalestraPatrick (approved):

PR #1815, opened 2023-01-27
Add support for watchOS single target applications

BalestraPatrick:
This is an attempt at cherry-picking a couple of commits from the upstream branch that add support for the new watchOS single target applications in Xcode 14. I still need to figure out if it's worth to group multiple commits in one or it'll be better to split them up. Just want to see what CI says first.

Update: looks like we'll need to upgrade CI to Xcode 14 to support this new rule.


keith:
once this is closer to merge plz comment on the cherry picks so we can keep track

BalestraPatrick:
Continuing in https://github.com/bazelbuild/rules_apple/pull/2048

PR #1820, opened 2023-02-06
[CI] Use Xcode 14.2

BalestraPatrick:
Looks like CI ~is~ was broken because the [CI fleet was upgraded to Xcode 14.2](https://github.com/bazelbuild/continuous-integration/issues/1431#issuecomment-1418841590) and some machines were missing the watchOS and tvOS SDKs. Now that the iMacs are properly set up, we should be able to simply use Xcode 14.2.

We can't use Xcode 13.0 anymore because it doesn't run on macOS Ventura (which all the x86_64 machines are on after the upgrade).

Future tasks might include adding support for arm64 in our tooling so that we can switch to run on Mac Studios instead of iMacs.

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -4,6 +4,7 @@ x_defaults:
   # it doesn't know about; so that is used to avoid repeating common subparts.
   common: &common
     platform: macos
+    xcode_version: "14.2"
     build_targets:
     - "tools/..."
     - "test/..."


brentleyjones:
In the end, we don't want this flag, and just want our tests to work on either Xcode 13 or Xcode 14, right? Especially since we might be routed to either set of macs (or will we always be routed to Xcode 14 if we don't specify arm64?).

BalestraPatrick:
If we want to keep on running on x86_64, we need to run on iMacs for the moment. From [this comment](https://github.com/bazelbuild/continuous-integration/issues/1431#issuecomment-1419344499), those machines have Xcode 13.0, 13.4.1 and 14.2 installed, but given they run Ventura now, only 14.2 is usable.

brentleyjones (approved):

PR #1822, opened 2023-02-07
Fix test crash at launch on iOS 14

jszumski:
Fixes an issue where the sanitizer dylib glob pattern is used in its raw form if no sanitizer dylibs are found during the test.  Using a bad path in `DYLD_INSERT_LIBRARIES` breaks iOS 14 simulators although newer versions appear to ignore it.

Without this fix, the added test attempts to load the glob string itself:

> dyld: could not load inserted library '<snip>/test_runner_work_dir.khFsCk/PassingUnitTestiOS14/PassingUnitTestiOS14.xctest/Frameworks/libclang_rt.*.dylib' because image not found

Bash globs will return the raw pattern if no matches are found, so I made sure the file exists before setting `DYLD_INSERT_LIBRARIES`.  I considered setting `shopt -s nullglob` but that could have side effects for other globs in the template.

------

Added `--test_tag_filters=-skipci` to presubmit so that this iOS 14 test can be skipped in CI.

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -12,13 +12,14 @@ x_defaults:
     - "tools/..."
     - "test/..."
     - "examples/..."
+    test_flags:
+    - --test_tag_filters=-skipci
   common_last_green: &common_last_green
     build_flags:
       # https://github.com/bazelbuild/bazel/issues/16939
       - --incompatible_objc_linking_info_migration=false
     test_flags:
       - --incompatible_objc_linking_info_migration=false
-      - --test_tag_filters=-skipci
 
 # NOTE: To avoid listing the same things for build_flags/test_flags for each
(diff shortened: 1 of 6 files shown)


brentleyjones (approved):

luispadron (approved):

jszumski:
> xctestrunner.shared.ios_errors.IllegalArgumentError: The simulator os version 14.5 is not supported. Supported simulator os versions are ['16.2'].

We can either skip this test in CI or get this specific runtime added to the CI nodes.

brentleyjones:
It's pretty hard to get pre-16.0 simulator runtimes installed. I say we skip it on CI for now.

jszumski:
Updated with a new "skipci" tag

brentleyjones (approved):

Issue #1759, opened 2022-11-17
Extend coverage to allow filtering of sources


keith:
Currently `$COVERAGE_MANIFEST` includes everything under `--instrumentation_filter`. In the case you want to be strict about "only a target's direct test should be considered for its coverage" this file includes to much. We should add an attribute to unit test targets to allow filtering more specifically, then users can pass whatever files they want, potentially with a genrule

@maxwellE is going to work on this

maxwellE:
Working on this now

PR #1824, opened 2023-02-08
Extend coverage to allow filtering of sources

maxwellE:
Allows users to provide a custom coverage manifest file to allow for
more precise filtering of sources in llvm coverage invocations.

Related #1759

### DIFF
--- a/apple/testing/default_runner/macos_test_runner.template.sh
+++ b/apple/testing/default_runner/macos_test_runner.template.sh
@@ -162,7 +162,7 @@ llvm_cov_export_status=0
 lcov_args=(
   -instr-profile "$profdata"
   -ignore-filename-regex='.*external/.+'
-  -path-equivalence=.,"$PWD"
+  -path-equivalence="$ROOT,."
 )
 xcrun llvm-cov \
   export \
--- a/apple/testing/default_runner/macos_test_runner.template.sh
+++ b/apple/testing/default_runner/macos_test_runner.template.sh
@@ -148,6 +148,12 @@ if [[ "${COVERAGE:-}" -ne 1 ]]; then
   exit 0
 fi
 
+llvm_coverage_manifest="$COVERAGE_MANIFEST"
+readonly provided_llvm_coverage_manifest="%(test_llvm_coverage_manifest)s"
+if [[ -s "${provided_llvm_coverage_manifest:-}" ]]; then
+  llvm_coverage_manifest="$provided_llvm_coverage_manifest"
+fi
+
 readonly profdata="$TEST_TMP_DIR/coverage.profdata"
 xcrun llvm-profdata merge "$profraw" --output "$profdata"
 
@@ -156,14 +162,14 @@ llvm_cov_export_status=0
 lcov_args=(
   -instr-profile "$profdata"
   -ignore-filename-regex='.*external/.+'
-  -path-equivalence="$ROOT",.
+  -path-equivalence=.,"$PWD"
 )
 xcrun llvm-cov \
   export \
   -format lcov \
   "${lcov_args[@]}" \
   "$test_binary" \
-  @"$COVERAGE_MANIFEST" \
+  @"$llvm_coverage_manifest" \
   > "$COVERAGE_OUTPUT_FILE" \
   2> "$export_error_file" \
   || llvm_cov_export_status=$?
@@ -183,7 +189,7 @@ if [[ -n "${COVERAGE_PRODUCE_JSON:-}" ]]; then
     -format text \
     "${lcov_args[@]}" \
     "$test_binary" \
-    @"$COVERAGE_MANIFEST" \
+    @"$llvm_coverage_manifest" \
     > "$TEST_UNDECLARED_OUTPUTS_DIR/coverage.json"
     2> "$export_error_file" \
     || llvm_cov_json_export_status=$?
--- a/apple/testing/default_runner/macos_test_runner.template.sh
+++ b/apple/testing/default_runner/macos_test_runner.template.sh
@@ -162,7 +162,7 @@ llvm_cov_export_status=0
 lcov_args=(
   -instr-profile "$profdata"
   -ignore-filename-regex='.*external/.+'
-  -path-equivalence="$ROOT",.
+  -path-equivalence=".,$PWD"
 )
 xcrun llvm-cov \
   export \
--- a/apple/testing/default_runner/macos_test_runner.template.sh
+++ b/apple/testing/default_runner/macos_test_runner.template.sh
@@ -162,7 +162,7 @@ llvm_cov_export_status=0
 lcov_args=(
   -instr-profile "$profdata"
   -ignore-filename-regex='.*external/.+'
-  -path-equivalence="$ROOT,."
+  -path-equivalence="$ROOT",.
 )
 xcrun llvm-cov \
   export \
--- a/examples/ios/HelloWorldSwift/BUILD
+++ b/examples/ios/HelloWorldSwift/BUILD
@@ -1,5 +1,5 @@
 load("@bazel_skylib//rules:build_test.bzl", "build_test")
-load("//apple:ios.bzl", "ios_application")
+load("//apple:ios.bzl", "ios_application", "ios_unit_test")
 load("@build_bazel_rules_swift//swift:swift.bzl", "swift_library")
 
 licenses(["notice"])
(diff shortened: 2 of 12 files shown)


apple/testing/default_runner/macos_test_runner.template.sh
```diff
@@ -163,7 +175,7 @@ xcrun llvm-cov \
   -format lcov \
   "${lcov_args[@]}" \
   "$test_binary" \
-  @"$COVERAGE_MANIFEST" \
+  @"$llvm_coverage_manifest" \
```
maxwellE, line 178: I thought that this would be enough to do the filtering, but it does not appear to be the case

apple/testing/default_runner/macos_test_runner.template.sh
```diff
@@ -148,6 +148,12 @@ if [[ "${COVERAGE:-}" -ne 1 ]]; then
   exit 0
 fi
 
+llvm_llvm_coverage_manifest="$llvm_coverage_manifest"
```
keith, line 151:
```suggestion
llvm_llvm_coverage_manifest="$COVERAGE_MANIFEST"
```

apple/testing/default_runner/macos_test_runner.template.sh
```diff
@@ -156,14 +162,14 @@ llvm_cov_export_status=0
 lcov_args=(
   -instr-profile "$profdata"
   -ignore-filename-regex='.*external/.+'
-  -path-equivalence="$ROOT",.
+  -path-equivalence=.,"$PWD"
```
keith, line 165: what's the gist of this change again? even when i did this originally i wasn't sure how this would affect things. are ROOT and PWD actually different?

keith: hrm so in that case wouldn't `ROOT` be the one in the binaries? it looks to me like `PWD` is really specific to the test 🤔

maxwellE: Its not matching up, at least when I run this:

`bazel coverage //examples/ios/HelloWorldSwift:UnitTests --action_env=LCOV_MERGER=/usr/bin/true --experimental_use_llvm_covmap --test_output=all --nocache_test_results --features=swift.coverage_prefix_map`

If I change it back to:

```terminal
  -instr-profile "$profdata"
  -ignore-filename-regex='.*external/.+'
  -path-equivalence="$ROOT",.
)
```

I get misses on all paths

maxwellE: What if we pass a label_list instead of a single file. This label list is the list of files to perform filtering with, we could then try to find out how to get those to have the right path relative to `$ROOT`?

keith: as long as something like this works for all cases i don't think we need to change what you have here, im just surprised.

maxwellE: What else should I test with in this flow? Given that coverage-prefix will be the default and that this works is that sufficient to merge?

mattrobmattrob:
Removing myself similar to the other reviewers that aren't Keith. 👍

test/ios_coverage_test.sh
```diff
@@ -173,6 +183,13 @@ function test_standalone_unit_test_coverage_json() {
       grep '"name":"SharedLogic.m:-\[SharedLogic doSomething\]"'
 }
 
+function test_standalone_unit_test_coverage_coverage_manifest() {
+  create_common_files
+  do_coverage ios --test_output=errors --ios_minimum_os=9.0 --experimental_use_llvm_covmap --action_env=LCOV_MERGER=/usr/bin/true --spawn_strategy=local --sandbox_debug //app:coverage_manifest_test || fail "Should build"
+  assert_contains "SharedLogic.m:-\[SharedLogic doSomething\]" "test-testlogs/app/coverage_manifest_test/coverage.dat"
+  assert_contains "SF:./app/StandaloneTest.m" "test-testlogs/app/coverage_manifest_test/coverage.dat"
```
maxwellE, line 190: @keith I am really surprised when testing this locally, I can never make it work. Even when feeding the same paths and artifacts to `llvm-cov export` I cannot get paths to match up and engage path-equivalence

examples/ios/HelloWorldSwift/BUILD
```diff
@@ -12,6 +12,40 @@ swift_library(
...
+    ],
+    cmd = "echo $(SRCS) > $@",
+)
+
+ios_unit_test(
+    name = "UnitTestsWithCoverageManifest",
+    minimum_os_version = "15.0",
+    test_coverage_manifest = "CoverageManifest.instrumented_files",
+    deps = [":Tests"],
 )
```
maxwellE, line 49: Coverage file from this test: https://gist.github.com/maxwellE/60bc86c1cae5730a0698a2a21a862dfe

examples/ios/HelloWorldSwift/BUILD
```diff
@@ -12,6 +12,40 @@ swift_library(
...
+        "CoverageManifest.instrumented_files",
+    ],
+    cmd = "echo $(SRCS) > $@",
+)
+
+ios_unit_test(
+    name = "UnitTestsWithCoverageManifest",
+    minimum_os_version = "15.0",
+    test_coverage_manifest = "CoverageManifest.instrumented_files",
+    deps = [":Tests"],
```
maxwellE, line 48: Coverage file from this test: https://gist.github.com/maxwellE/60bc86c1cae5730a0698a2a21a862dfe

examples/ios/HelloWorldSwift/BUILD
```diff
@@ -12,6 +12,40 @@ swift_library(
...
+    deps = [
+        ":Sources",
+    ],
+)
+
+ios_unit_test(
+    name = "UnitTests",
+    minimum_os_version = "8.0",
+    deps = [":Tests"],
+)
```
maxwellE, line 33: Coverage from this file without the sources exclusion https://gist.github.com/maxwellE/0180250b83dbc351064a7b68057148c6 . We can see that paths are still made relative even with the $PWD change

test/ios_coverage_test.sh
maxwellE: I am not sure where I am going wrong in these tests since I can get this working [here](https://github.com/bazelbuild/rules_apple/pull/1824/files#r1114530777). Is there some hidden default flags or is this using some older version of rules apple support?

keith:
note that you don't have to pull in master when you see that button if you don't want to, it's just there to make it easier when you do want to

test/ios_coverage_test.sh
```diff
@@ -173,6 +183,14 @@ function test_standalone_unit_test_coverage_json() {
       grep '"name":"SharedLogic.m:-\[SharedLogic doSomething\]"'
 }
 
+# ME: I am not sure why this does not work yet a very similar setup in examples/ios/HelloWorldSwift/BUILD does work
+# function test_standalone_unit_test_coverage_coverage_manifest() {
+#   create_common_files
+#   do_coverage ios --test_output=errors --ios_minimum_os=9.0 --experimental_use_llvm_covmap --action_env=LCOV_MERGER=/usr/bin/true //app:coverage_manifest_test || fail "Should build"
+#   assert_contains "SharedLogic.m:-\[SharedLogic doSomething\]" "test-testlogs/app/coverage_manifest_test/coverage.dat"
+#   assert_contains "SF:./app/StandaloneTest.m" "test-testlogs/app/coverage_manifest_test/coverage.dat"
```
keith, line 191:
```suggestion
  assert_contains "StandaloneTest.m:-\[StandaloneTest testAnything\]" "test-testlogs/app/coverage_manifest_test/coverage.dat"
  assert_contains "SF:./app/StandaloneTest.m" "test-testlogs/app/coverage_manifest_test/coverage.dat"
```

test/ios_coverage_test.sh
```diff
@@ -131,6 +131,10 @@ EOF
   CFBundleShortVersionString = "1.0";
   CFBundleVersion = "1.0";
 }
+EOF
+
+  cat > app/coverage_manifest.txt <<EOF
+app/SharedLogic.m
```
keith, line 137:
```suggestion
app/StandaloneTest.m
```

maxwellE: `app/SharedLogic.m` should appear though in coverage right?

keith:
this depends on https://github.com/bazelbuild/rules_apple/pull/1844

test/ios_coverage_test.sh
keith: hrm yea ok, worth a try, i didn't test after i fixed the core issues locally, so maybe this is fine as is

keith:
you can rebase now

keith (approved):

test/ios_coverage_test.sh
```diff
@@ -173,6 +183,13 @@ function test_standalone_unit_test_coverage_json() {
       grep '"name":"SharedLogic.m:-\[SharedLogic doSomething\]"'
 }
 
+function test_standalone_unit_test_coverage_coverage_manifest() {
+  create_common_files
+  do_coverage ios --test_output=errors --ios_minimum_os=9.0 --experimental_use_llvm_covmap --action_env=LCOV_MERGER=/usr/bin/true //app:coverage_manifest_test || fail "Should build"
+  assert_contains "SharedLogic.m:-\[SharedLogic doSomething\]" "test-testlogs/app/coverage_manifest_test/coverage.dat"
+  assert_contains "SF:./app/SharedLogic.m" "test-testlogs/app/coverage_manifest_test/coverage.dat"
```
mattrobmattrob, line 190: Is it worth adding a `assert_not_contains` to assure that the files outside of `coverage_manifest.txt` aren't present?

keith: yea good spot, i have a follow up change for this

PR #1828, opened 2023-02-13
Exclude test target source files from generated coverage report

mccorkill1:
When the `bazel coverage` command is run on an `ios_unit_test` target, the source files for the test target are included in the coverage report by default. That is not the desired behavior if you only want the coverage for what is being tested. This PR adds a feature`exclude_test_target_coverage` that if passed, the files from the test target itself will be excluded from the generated coverage report. This also is applied to any files present in test_host.


mccorkill1:
Neither feature gave the same behavior. I will add a test for this tomorrow morning. Thanks for taking a look so quickly!

mccorkill1:
I'm closing this pull request because https://github.com/bazelbuild/rules_apple/pull/1824 addresses the issue I was trying to solve with this change. https://github.com/bazelbuild/rules_apple/pull/1824/files#diff-af9a4d0c59504dc7a7afa0286806c4f56504fac1983c9caedd8c33ddaf25c0e2R257 fixes the use of instrumentation_filter with ios_unit_test so the new feature isn't needed.

PR #1831, opened 2023-02-15
Update apple_support

keith:


brentleyjones (approved):
Don't want to merge until 1.4.0 is on the BCR, right?

MODULE.bazel
```diff
@@ -10,7 +10,7 @@ module(
 # --- End generated content
 
 bazel_dep(name = "bazel_skylib", version = "1.3.0")
-bazel_dep(name = "apple_support", version = "1.3.2", repo_name = "build_bazel_apple_support")
+bazel_dep(name = "apple_support", version = "1.4.0", repo_name = "build_bazel_apple_support")
```
brentleyjones, line 13: Ideally we only bump this if we depend on the changes (and by "we" I mean users of the next version of rules_apple). With Bzlmod, rules should generally depend on the minimum version they support within a compatibility_level.

keith:
CI fails if it's not in the BCR so we're good

MODULE.bazel
keith: In this case I do want to start using the new toolchain on the tests here, but also shouldn't it be the other way around and we should always use the highest supported version? Realistically if we don't bump I'm slightly worried about implicit incompatibilities that we miss, but also it feels less risky for users since if they don't pin this one they could end up infinitely behind.

brentleyjones: Root modules should have the highest version, and we should use the highest version on CI (probably by modifying this file before testing), but non-root modules should use the lowest version they require, in order to not force higher versions on users (though they could use single_version_override to force a lowering?): https://bazelbuild.slack.com/archives/C014RARENH0/p1674838476782969

keith: Hrm. In the case of apple_support I'm not sure that's how we want to do things. Specifically since (in the future once crosstool_top is unused) users don't have to directly depend on apple_support, but updates in the toolchain might be useful for them. Also in general with bazel rules not being stuck on super old versions seems like a benefit so we don't break them when we do have to jump so many versions?

PR #1841, opened 2023-02-22
Add data attribute to apple_static_{framework,xcframework}_import rules

andre-alves:
This diff adds the `data` attribute to `apple_static_framework_import` and `apple_static_xcframework_imports` rules. 

Follow up from https://github.com/bazelbuild/rules_apple/pull/869#issuecomment-1438473570.

The new attribute works very similarly to objc_library/swift_library's data: it's a list of resource targets that will be bundled together in the final binary.

### Example:

```python
apple_static_xcframework_import(
    name = "Appboy_iOS_SDK",
    xcframework_imports = glob(["Appboy-iOS-SDK/Frameworks/Appboy_iOS_SDK.xcframework/**"]),
    data = [
        ":AppboyResourceBundle",
        ":AppboyResourceFiles",
    ],
)

apple_bundle_import(
    name = "AppboyResourceBundle",
    bundle_imports = glob(["Appboy-iOS-SDK/AppboyKit/Appboy.bundle/**"]),
)

apple_resource_group(
    name = "AppboyResourceFiles",
    resources = glob([
        "Appboy-iOS-SDK/AppboyUI/**/*.strings",
        "Appboy-iOS-SDK/AppboyUI/**/*.storyboard",
    ]),
)
```

### Why?

The previous implementation of these rules only allowed for `.bundle/` resource files when mixed together with `framework_imports`. This behavior limits use cases because:

1 - Vendors can distribute resources without .bundle packages;
2 - Resources for static libraries don't need to be inside the .framework directory, since they will end up in the root of the binary target anyway.

### Implementation

At first, I had thought that only adding the names of the rules to `resource_aspect.bzl` would be enough, but both rules add `AppleResourceInfo` when `.bundle/` files are present inside `framework_imports`, which blocks the aspect from collecting the resources from `data`. It wouldn't be possible to use both `data` and .bundle/** logic of `framerwork_imports` at the same time.

In order to support both attributes, `AppleResourceInfo` was replaced with `AppleFrameworkImportBundleInfo`, which is a new internal provider that `resource_aspect.bzl` will use to collect .bundle/** files from `framework_imports`.

I'm new to Bazel so let me know if you have any suggestions. Thanks!

### DIFF
--- a/apple/internal/apple_framework_import.bzl
+++ b/apple/internal/apple_framework_import.bzl
@@ -305,11 +305,9 @@ def _apple_static_framework_import_impl(ctx):
     # TODO(b/207475773): Remove grep-includes once it's no longer required for cc_common APIs.
     grep_includes = ctx.file._grep_includes
 
-    providers = []
-
-    # Create DefaultInfo provider
-    default_info_provider = DefaultInfo(runfiles = ctx.runfiles(files = ctx.files.data))
-    providers.append(default_info_provider)
+    providers = [
+        DefaultInfo(runfiles = ctx.runfiles(files = ctx.files.data)),
+    ]
 
     framework_imports_by_category = framework_import_support.classify_framework_imports(
         ctx.var,
--- a/apple/internal/apple_framework_import.bzl
+++ b/apple/internal/apple_framework_import.bzl
@@ -23,10 +23,6 @@ load(
     "dicts",
 )
 load(
-    "@bazel_skylib//lib:partial.bzl",
-    "partial",
-)
-load(
     "@bazel_skylib//lib:paths.bzl",
     "paths",
 )
@@ -39,12 +35,12 @@ load(
     "AppleFrameworkImportInfo",
 )
 load(
-    "@build_bazel_rules_apple//apple/internal:cc_toolchain_info_support.bzl",
-    "cc_toolchain_info_support",
+    "@build_bazel_rules_apple//apple/internal/providers:framework_import_bundle_info.bzl",
+    "AppleFrameworkImportBundleInfo",
 )
 load(
-    "@build_bazel_rules_apple//apple/internal:resources.bzl",
-    "resources",
+    "@build_bazel_rules_apple//apple/internal:cc_toolchain_info_support.bzl",
+    "cc_toolchain_info_support",
 )
 load(
     "@build_bazel_rules_apple//apple/internal/utils:bundle_paths.bzl",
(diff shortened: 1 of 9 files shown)


apple/internal/apple_framework_import.bzl
```diff
@@ -310,6 +306,11 @@ def _apple_static_framework_import_impl(ctx):
     grep_includes = ctx.file._grep_includes
 
     providers = []
+
+    # Create DefaultInfo provider
+    default_info_provider = DefaultInfo(runfiles = ctx.runfiles(files = ctx.files.data))
+    providers.append(default_info_provider)
```
keith, line 312:
```suggestion
    providers = [
        default_info_provider = DefaultInfo(runfiles = ctx.runfiles(files = ctx.files.data))
    ]
```

keith (approved):
looks good, style nit, also testing against the lyft codebase

keith:
looks good

apple/internal/apple_xcframework_import.bzl
```diff
@@ -510,7 +512,13 @@ def _apple_static_xcframework_import_impl(ctx):
...
 
     providers = []
-    providers.append(DefaultInfo(files = depset(xcframework_imports)))
+
+    # Create DefaultInfo provider
+    default_info_provider = DefaultInfo(
+        files = depset(xcframework_imports),
+        runfiles = ctx.runfiles(files = ctx.files.data),
+    )
+    providers.append(default_info_provider)
```
brentleyjones, line 521: Same comment as above, having `providers = [DefaultInfo(...)]` to start.

apple/internal/aspects/resource_aspect.bzl
```diff
@@ -120,6 +125,12 @@ def _apple_resource_aspect_impl(target, ctx):
         collect_args["res_attrs"] = ["data"]
         owner = str(ctx.label)
 
+    elif ctx.rule.kind in ["apple_static_framework_import", "apple_static_xcframework_import"]:
```
brentleyjones, line 128: Does this need a `kind` check, versus just looking for the provider? Also, it seems odd that we need a new provider versus using the existing `AppleResourceInfo` infrastructure.

andre-alves: Yes, because the provider is only present when there are bundle resource files inside the .framework.  The aspect also needs to collect resources from the `data` attribute, similar to `objc_library/swift_library`:

```starlark
collect_args["res_attrs"] = ["data"]
owner = str(ctx.label)
```

In order to use `AppleResourceInfo`, we would need to collect and merge resources from `data` inside both `apple_static_framework_import` and `apple_static_xcframework_import` rules. I chose to delegate this logic to `apple_resource_aspect`, as the other rules do.

brentleyjones: Do we need the kind check though if we are keeping the provider? Couldn't moving this outside the kind checks, and looking for the provider, allow custom rules to add support for this as well?

brentleyjones: Nevermind, I see.

brentleyjones (approved):

keith:
thanks!

PR #1844, opened 2023-02-23
Use new toolchain in tests

keith:
I thought doing this in the root bazelrc was enough but it's not


brentleyjones (approved):

keith:
wonder if this will break things now

PR #1851, opened 2023-02-24
Avoid declaring nested artifacts.

tjgq:
A nested artifact is a declare_file or declare_directory inside another declare_directory. The --incompatible_strict_conflict_checks Bazel flag bans nested artifacts, and will be flipped in Bazel 7. This is the last known blocker for flipping the flag.

### DIFF
--- a/apple/internal/apple_xcframework_import.bzl
+++ b/apple/internal/apple_xcframework_import.bzl
@@ -258,7 +258,6 @@ def _get_xcframework_library_with_xcframework_processor(
     library_suffix = ".framework" if xcframework.bundle_type == _BUNDLE_TYPE.frameworks else ""
     library_path = xcframework.bundle_name + library_suffix
 
-    library_dir = intermediates.directory(dir_name = library_path, **intermediates_common)
     framework_imports_dir = intermediates.directory(
         dir_name = paths.join("framework_imports", library_path),
         **intermediates_common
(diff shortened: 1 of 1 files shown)


brentleyjones:
Are you making a similar change internally (which would then appear in our `upstream` branch)?

tjgq:
@brentleyjones I'm not, thanks for the heads up (I was unsure whether we import to or export from google3). I'll try to make the changes internally in that case.

Since I have your attention, there's a potential issue with this approach. If the `ProcessXCFrameworkFiles` action creates any files in the top-level `--library_dir` (i.e., outside the the declared subdirectories), after this change we will fail to propagate them in the `framework_files` field returned by `_get_xcframework_library_with_xcframework_processor`. Do you happen to know whether this can happen in practice?

brentleyjones:
>  I'll try to make the changes internally in that case.

Since this file has diverged from upstream a bit, please continue to resolve it here as well if possible 🙏. Just wanted to let you know that export from google3, not import.

> Do you happen to know whether this can happen in practice?

I'm not sure. @keith do you know?

tjgq:
It appears that there's already a commit in the `upstream` branch that fixes this issue (https://github.com/bazelbuild/rules_apple/commit/375cfa418f5411cc3f59063b1f4a546f804a87b0), so we could cherry-pick it as an alternative to this PR.

keith:
trying that one here https://github.com/bazelbuild/rules_apple/pull/1857

tjgq:
Much appreciated, thanks!

brentleyjones:
We should probably flip the flag for our CI to prevent regressions?

tjgq:
Sure, sent #1861.

PR #1852, opened 2023-02-24
Add support for UI testing with `ios_xctestrun_runner`

maxwellE:
This PR adds a new default
runner that allows for UI testing.

### DIFF
--- a/apple/testing/default_runner/ios_xctestrun_runner.template.sh
+++ b/apple/testing/default_runner/ios_xctestrun_runner.template.sh
@@ -111,11 +111,12 @@ if [[ -n "$test_host_path" ]]; then
       "$runner_app_destination/Info.plist"
 
     readonly runner_app_frameworks_destination="$runner_app_destination/Frameworks"
-    ditto "$libraries_path/Frameworks/XCTest.framework" "$runner_app_frameworks_destination/XCTest.framework"
-    ditto "$libraries_path/PrivateFrameworks/XCTestCore.framework" "$runner_app_frameworks_destination/XCTestCore.framework"
-    ditto "$libraries_path/PrivateFrameworks/XCUIAutomation.framework" "$runner_app_frameworks_destination/XCUIAutomation.framework"
-    ditto "$libraries_path/PrivateFrameworks/XCTAutomationSupport.framework" "$runner_app_frameworks_destination/XCTAutomationSupport.framework"
-    ditto "$libraries_path/PrivateFrameworks/XCUnit.framework" "$runner_app_frameworks_destination/XCUnit.framework"
+    mkdir -p "$runner_app_frameworks_destination"
+    cp -R "$libraries_path/Frameworks/XCTest.framework" "$runner_app_frameworks_destination/XCTest.framework"
+    cp -R "$libraries_path/PrivateFrameworks/XCTestCore.framework" "$runner_app_frameworks_destination/XCTestCore.framework"
+    cp -R "$libraries_path/PrivateFrameworks/XCUIAutomation.framework" "$runner_app_frameworks_destination/XCUIAutomation.framework"
+    cp -R "$libraries_path/PrivateFrameworks/XCTAutomationSupport.framework" "$runner_app_frameworks_destination/XCTAutomationSupport.framework"
+    cp -R "$libraries_path/PrivateFrameworks/XCUnit.framework" "$runner_app_frameworks_destination/XCUnit.framework"
   fi
 else
   xctestrun_test_host_path="__PLATFORMS__/iPhoneSimulator.platform/Developer/Library/Xcode/Agents/xctest"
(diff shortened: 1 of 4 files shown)


apple/testing/default_runner/ios_xctestrun_runner.template.sh
```diff
@@ -80,11 +80,44 @@ for single_test_env in ${test_env//,/ }; do
...
+xcrun_is_xctrunner_hosted_bundle="false"
+xcrun_is_ui_test_bundle="false"
 if [[ -n "$test_host_path" ]]; then
   xctestrun_test_host_path="__TESTROOT__/$test_host_name.app"
   xctestrun_test_host_based=true
   # If this is set in the case there is no test host, some tests hang indefinitely
   xctestrun_env+="<key>XCInjectBundleInto</key><string>$(escape "__TESTHOST__/$test_host_name.app/$test_host_name")</string>"
+
+  ui_testing_enabled="%(ui_testing_enabled)s"
+  if [[ -n "$ui_testing_enabled" ]]; then
```
keith, line 93: I think instead of doing this and requiring the separate runner we can use the `test_type` expansion? https://github.com/bazelbuild/rules_apple/blob/e6a22d7017813bc1825b3e5fbeb525311f4d9cf8/apple/internal/testing/apple_test_rule_support.bzl#L117 then we can just reuse the ordered runner? (maybe we should just rename that to have a more default name)

examples/ios/HelloWorldSwift/BUILD
```diff
@@ -43,7 +51,7 @@ genrule(
 
 ios_unit_test(
     name = "UnitTestsWithCoverageManifest",
-    minimum_os_version = "15.0",
```
keith, line 46: ?

keith (commented):
looks great!

apple/testing/default_runner/ios_xctestrun_runner.template.sh
tinder-maxwellelliott: Will do

maxwellE: Looks like I always get `XCTEST` from that variable

keith: hrm, definitely should give you something else https://github.com/bazelbuild/rules_apple/blob/e6a22d7017813bc1825b3e5fbeb525311f4d9cf8/apple/internal/testing/ios_rules.bzl#L52

apple/testing/default_runner/ios_xctestrun_runner.template.sh
```diff
@@ -80,11 +80,44 @@ for single_test_env in ${test_env//,/ }; do
...
+    xcrun_is_ui_test_bundle="true"
+    xcrun_target_app_path="$xctestrun_test_host_path"
+    # If ui testing is enabled we need to copy out the XCTRunner app, update its info.plist accordingly and finally
+    # copy over the needed frameworks to enable ui testing
+    readonly runner_app_name="$test_bundle_name-Runner"
+    readonly runner_app="$runner_app_name.app"
+    readonly runner_app_destination="$test_tmp_dir/$runner_app"
+    libraries_path="$(xcode-select -p)/Platforms/iPhoneSimulator.platform/Developer/Library"
+    ditto "$libraries_path/Xcode/Agents/XCTRunner.app" "$runner_app_destination"
+    mv "$runner_app_destination/XCTRunner" "$runner_app_destination/$runner_app_name"
```
keith, line 104: could probably do this in 1 command instead with `cp -R` or something?

keith (commented):
looking good, is there much left here?

apple/testing/default_runner/ios_xctestrun_runner.template.sh
tinder-maxwellelliott: We do need the rename I believe though? cp -R would copy the app as is.

maxwellE:
@keith Seeing some [instability now from simulator creation](https://gist.github.com/maxwellE/279df56a97f012097c60d2b71863c09e). Does that have anything to do with this change?

apple/testing/default_runner/ios_xctestrun_runner.template.sh
keith: not if the target destination doesn't exist, `cp -R foo/bar.app baz/qux.app` should work fine as long as `qux.app` doesn't exist

keith:
i wouldn't think so, that's before this logic

apple/testing/default_runner/ios_xctestrun_runner.template.sh
maxwellE: Should we then just use `XCTRunner.app/XCTRunner` as the bk arh name and update the Info.plist as such. I was trying to match xcode with the `PRODUCTNAME-Runner.app/PRODUCTNAME-Runner` naming

keith: i think that would be fine since our directory is unique. I think in Xcode they'd have to update it like that since derived data could contain multiple runner apps for different UI test targets

keith (approved):

keith:
thanks!

PR #1855, opened 2023-02-24
Remove `apple_common.Objc` provider usage related to linking now that the Apple linking logic in Bazel gets everything from `CcInfo`.

keith:
There are still a couple uses in `swift_clang_module_aspect` unrelated to linking that will be cleaned up separately.

PiperOrigin-RevId: 503167994
(cherry picked from commit 19e48b97a7bc224bc5aff4e04fe2b42aabb5da8d)


keith:
not sure this one will be backwards compat

brentleyjones:
If it's not, we should add code to make it work with both Bazel 6 and above. We shouldn't do a fork like we did last time unless absolutely necessary: https://github.com/bazelbuild/bazel/issues/17372#issuecomment-1438703881.

keith:
we can see how far just ignoring this gets us. seems like there's some reasonable deduping happening anyways

Issue #1873, opened 2023-03-01
[ios_xctestrun_runner] Allow simulator name input to test runner


mattrobmattrob:
Allow simulator name to be passed into the `ios_xctestrun_runner` similar to how `device_id` would work for the old `ios_test_runner` via `--test_arg=--destination=platform=ios_simulator,id=$SIMULATOR_IDENTIFIER`:

https://github.com/bazelbuild/rules_apple/blob/e079b9bdc6a5218574fde349ed4525c6ed4359c7/apple/testing/default_runner/ios_test_runner.template.sh#L28-L34

`simulator_creator.py` already supports it:

https://github.com/bazelbuild/rules_apple/blob/e079b9bdc6a5218574fde349ed4525c6ed4359c7/apple/testing/default_runner/simulator_creator.py#L72-L74

This will allow persistent simulator usage across test bundles (similar to `REUSE_GLOBAL_SIMULATOR`) out of band from Bazel.

keith:
sounds fine, just worth noting that if you change names it will affect test caching

mattrobmattrob:
> sounds fine, just worth noting that if you change names it will affect test caching

Yup, in our case, we do it to mitigate simulator creation instability in the VMs on CI. And restart the VM in the case where it fails. Always using the same name on CI to get cache hits across the runs. Not perfect but it works for us.

PR #1874, opened 2023-03-01
Add support for simulator name input to `ios_xctestrun_runner`

mattrobmattrob:
Pass in a specific simulator name using `bazel coverage --test_arg=--simulator_name=BazelPhone ...` to the test runner to avoid creating it/booting it in-situ in the test runner.

**No Args Provided:**
```
...
Monitoring boot status for BAZEL_TEST_iPhone 8 Plus_16.2 (702D84B5-AE74-422E-8C06-9A4A694D39D3).
Booting device...
...
```

**`--test_arg=--simulator_id=BazelPhone` Provided (Already Booted):**
```
...
Simulator state is: booted
...
```

**`--test_arg=--simulator_id=BazelPhone` Provided (Shutdown):**
```
...
Simulator state is: shutdown
Monitoring boot status for BazelPhone (47558DBF-629E-468E-AC45-CF5CA33EFA70).
Booting device...
...
```

Closes https://github.com/bazelbuild/rules_apple/issues/1873.

### DIFF
--- a/apple/testing/default_runner/ios_xctestrun_runner.template.sh
+++ b/apple/testing/default_runner/ios_xctestrun_runner.template.sh
@@ -155,14 +155,11 @@ readonly xctestrun_file="$test_tmp_dir/tests.xctestrun"
   -e "s@BAZEL_COVERAGE_OUTPUT_DIR@$test_tmp_dir@g" \
   "%(xctestrun_template)s" > "$xctestrun_file"
 
-simulator_creator_args=(
+simulator_id="$("./%(simulator_creator.py)s" \
   "%(os_version)s" \
   "%(device_type)s" \
-)
-if [[ -n "${BAZEL_IOS_SIMULATOR_NAME:-}" ]]; then
-  simulator_creator_args+=(--name "${BAZEL_IOS_SIMULATOR_NAME}")
-fi
-simulator_id="$("./%(simulator_creator.py)s" "${simulator_creator_args[@]}")"
+  --name "${BAZEL_IOS_SIMULATOR_NAME:-}"
+)"
 
 test_exit_code=0
 testlog=$(mktemp)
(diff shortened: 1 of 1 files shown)


mattrobmattrob:
Would we prefer to unify the prefix of these (and update `CREATE_XCRESULT_BUNDLE`) before anyone becomes dependent on them for functionality?

keith:
hrm in this case would it make more sense to use a test_arg instead like your other ID example (which I guess this test runner just doesn't support yet 🙃 )? I'm not really sure env vs arg in general, but maybe arg is nicer because you probably get some validation if the arg you pass is bogus.

yea good question on prefixes. does seem like a good idea

mattrobmattrob:
> hrm in this case would it make more sense to use a test_arg instead like your other ID example (which I guess this test runner just doesn't support yet 🙃 )? I'm not really sure env vs arg in general, but maybe arg is nicer because you probably get some validation if the arg you pass is bogus.
> 
> yea good question on prefixes. does seem like a good idea

I'd probably just prefer to unify the "runtime change" mechanisms in the runner template. So something like parsed input args or env vars for things like `CREATE_XCRESULT_BUNDLE`, this functionailty, and any outcome of https://github.com/bazelbuild/rules_apple/issues/1875.

Seems like input args (being at the top of the file and sorta co-located) might be the best for discoverability along with updating:

https://github.com/bazelbuild/rules_apple/blob/e079b9bdc6a5218574fde349ed4525c6ed4359c7/apple/testing/default_runner/ios_xctestrun_runner.bzl#L156-L157

mattrobmattrob:
Is it nice to avoid environment variables all together to not get any confusing passthrough to the underlying test/test host exection?

mattrobmattrob:
Opted for `--test_arg` since I think it's nice for correctness/validation like you mention. Avoids polluting the test environment just for runner overrides as well.

keith:
there's definitely some confusion there, but realistically we should probably try to solve with docs as well

keith (approved):

PR #1878, opened 2023-03-02
Add `test_filter`/`--test_filter` support to `ios_xctestrun_runner`

mattrobmattrob:
Two approaches handled to mimic this behavior in the older runner:

https://github.com/bazelbuild/rules_apple/blob/e079b9bdc6a5218574fde349ed4525c6ed4359c7/apple/testing/default_runner/ios_test_runner.template.sh#L145-L151

CLI arg:
```bash
# Test class (skip)
$ bazel coverage --test_filter=-FooTests //FooTests
# Test method (skip)
$ bazel coverage --test_filter=-FooTests/testBar //FooTests
```

`test_filter` attribute:
```python
ios_unit_test(
  name = "FooTests",
  ...
  test_filter = ["-FooTests/testBar"],
  ...
)
```

Closes https://github.com/bazelbuild/rules_apple/issues/1877.

### DIFF
--- a/apple/testing/default_runner/ios_xctestrun_runner.template.sh
+++ b/apple/testing/default_runner/ios_xctestrun_runner.template.sh
@@ -154,6 +154,62 @@ if [[ -n "$sanitizer_dyld_env" ]]; then
   xctestrun_libraries="${xctestrun_libraries}:${sanitizer_dyld_env}"
 fi
 
+TEST_FILTER="%(test_filter)s"
+xctestrun_skip_test_section=""
+xctestrun_only_test_section=""
+
+# Use the 'TESTBRIDGE_TEST_ONLY' environment variable set by Bazel's
+# '--test_filter' flag to set the xctestrun's skip/only parameters.
+#
+# Any test prefixed with '-' will be passed to 'SkipTestIdentifiers'. Otherwise
+# the tests is passed to 'OnlyTestIdentifiers',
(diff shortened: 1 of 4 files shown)


mattrobmattrob:
Caveat: Testing the `OnlyTestIdentifiers` functionality in `xcodebuild` didn't work for me, seems like an Xcode problem vs. a test runner problem though.

keith:
can you mirror the unit test we have for the other one for this? i know there were a few PRs trying to make this behavior work correctly that should be covered

keith:
right now this only applies to the xctestrun codepath too, i think it might be able to apply to the simctl one, but otherwise we should make it force the xctestrun one

mattrobmattrob:
> right now this only applies to the xctestrun codepath too, i think it might be able to apply to the simctl one, but otherwise we should make it force the xctestrun one

Yeah, was thinking about this in relation to the other xcodebuild flags too.

keith:
yea i should have caught that yesterday. similar to asking for a result bundle i think we should just force the other path in that case

mattrobmattrob:
> 

I'll add the test. Are you okay if I handle the force-to-`xcodebuild` in another PR though? I'd like to move the `tests.xctestrun` creation within the `xcodebuild`-only branch.

keith:
yea separate PR sounds fine

keith:
do we want to run these on CI or i guess maybe we're skipping for now since we had to disable those other ones?

keith (approved):

PR #1883, opened 2023-03-03
Include clang_rt independent of asan, tsan, ubsan features.

mjburghard:
### What has changed
- Add support for a new feature flag that includes the clang_rt dylib.

### Why this was changed
- This allows specifying sanitizer compiler flags independent of the defaults set by other rules, but uses rules_apple to include the required clang_rt dylibs.

# How to test
Run e.g. `bazelisk run @//examples/ios/HelloWorldSwift:HelloWorldSwift --features=include_clang_rt --linkopt=-fsanitize=address`. The app bundle should include `libclang_rt.asan_iossim_dynamic.dylib` in its frameworks directory.

### DIFF
--- a/test/ios_application_swift_test.sh
+++ b/test/ios_application_swift_test.sh
@@ -150,9 +150,9 @@ EOF
   assert_ipa_contains_swift_dylibs_for_device
 }
 
-# Tests that swift_library build with ASAN enabled and that the ASAN
+# Tests that swift_library builds with ASAN enabled and that the ASAN
 # library is packaged into the IPA when enabled.
-function disabled_test_swift_builds_with_asan() {  # Blocked on b/73547309
+function test_swift_builds_with_asan() {  # Blocked on b/73547309
   create_minimal_ios_application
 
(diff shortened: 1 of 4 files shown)


keith:
Looks good. Can you sign the cla and add a small test case?

test/ios_application_swift_test.sh
```diff
@@ -194,4 +194,82 @@ EOF
   fi
 }
 
+# Tests that swift_library builds with include_clang_rt and asan linker option
```
mjburghard, line 197: Is there any information why the two tests above are disabled? Is `b/73547309` referring to some internal bug report?
The test work fine locally. If this issue is resolved, I can enable them, add a new test covering UBSAN, and add the tests for macOS as well.

keith: hrm yea i would just try to enable them at this point, idk either since it's an internal google one but it's highly likely that it works now

keith (approved):

keith:
thanks!

PR #1889, opened 2023-03-06
Pass TEST_UNDECLARED_OUTPUTS_DIR to test processes

keith:


apple/testing/default_runner/ios_xctestrun_runner.template.sh
```diff
@@ -92,9 +92,9 @@ function escape() {
 # to the test runner
 test_env="%(test_env)s"
```
mattrobmattrob, line 93: It does seem like it should just always come down as part of this substitution but perhaps an improvement for a later time in Bazel.

mattrobmattrob (approved):

apple/testing/default_runner/ios_xctestrun_runner.template.sh
keith: yea i don't think we have it in the process that does the substitutions here since that's a build and not a test run, but idk

luispadron (approved):

PR #1891, opened 2023-03-08
Fix static frameworks with objc provider

keith:
In 3872a85c145710f183065fbc61edbad86d4d1ab0 I got confused about these
arguments being the real objc provider directly vs our wrapper which
actually moves static_framework_file into imported_library. Without this
rules_swift does not correctly remove autolink load commands that are
invalid.


brentleyjones (approved):

keith:
ok i figured out the issue with this and will have to debug more

mattrobmattrob (approved):

keith:
ok so i think this needs to land with https://github.com/bazelbuild/rules_swift/pull/1020

PR #1893, opened 2023-03-10
Fix shellcheck warning

keith:
ALL_TESTS was an array but it only ever had a single element. All we
care about doing is looping through the string separated by comma so we
don't need it to be an array at all.


apple/testing/default_runner/ios_xctestrun_runner.template.sh
```diff
@@ -181,18 +181,16 @@ xctestrun_only_test_section=""
...
-    ALL_TESTS=("$TESTBRIDGE_TEST_ONLY")
+    ALL_TESTS="$TESTBRIDGE_TEST_ONLY"
   else
-    ALL_TESTS=("$TEST_FILTER")
+    ALL_TESTS="$TEST_FILTER"
   fi
 
-  for TEST in $ALL_TESTS; do
+  saved_IFS=$IFS
+  IFS=","; for TEST in $ALL_TESTS; do
```
keith, line 193: this is just stylistic to make it more clear that this is the only place we care about this value being set

mattrobmattrob (approved):

PR #1894, opened 2023-03-10
Support filtering absolute paths in coverage data

keith:
llvm-covs path-equivalence flag is a bit strange. Everything is
absolutized, including the paths in the binary, and the flag isn't
consulted in the case the files exist. Because of this if you're not
using the coverage-prefix-map flags and end up with absolute paths in
your binaries, the previous filtering wasn't working. Using
coverage-prefix-map is the default now, but folks using older rules and
the pre-apple_support toolchain might still be doing this. When the
filtering fails it just fails silently, so enabling this is nice to do
at least for now.


brentleyjones (approved):

mattrobmattrob (approved):
👏

PR #1897, opened 2023-03-15
Add dylib extension to dynamic libraries

codeman9:
Right now the rules append `_lipobin` during linking. This no longer works because: https://github.com/bazelbuild/bazel/blob/1fe702cfbaeb9e5a93f4b122359d416a18c5ee87/src/main/java/com/google/devtools/build/lib/rules/cpp/CcModule.java#L648-L657 is being used (due to toolchain migration?)

This fixes the following error that shows up when linking dynamic libraries by adding `.dylib` if `-dynamiclib` is in the `extra_linkopts`:
```
Error in create_library_to_link: 'MyFramework.dynamic_framework_lipobin' does not have any of the allowed extensions .so, .dylib, .dll or .pyd
```

### DIFF
--- a/test/ios_application_test.sh
+++ b/test/ios_application_test.sh
@@ -586,6 +586,46 @@ function test_prebuilt_static_apple_static_framework_import_resources() {
       "Payload/app.app/fmwk.bundle/Some.plist"
 }
 
+# Tests that a dynamic framework with a "." in the name builds
+function test_ios_framework_name_with_dot() {
+  create_common_files
+  cat >> app/BUILD <<EOF
+load("@build_bazel_rules_apple//apple:ios.bzl",
+    "ios_framework"
+)
+ios_framework(
+    name = "fmwk.dynamic_framework",
+    deps = [":lib"],
+    bundle_id = "my.bundle.id",
+    families = ["iphone"],
+    infoplists = ["Info.plist"],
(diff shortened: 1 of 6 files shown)


test/ios_application_test.sh
```diff
@@ -586,6 +586,46 @@ function test_prebuilt_static_apple_static_framework_import_resources() {
...
 
+# Tests that a dynamic framework with a "." in the name builds
+function test_ios_framework_name_with_dot() {
+  create_common_files
+  cat >> app/BUILD <<EOF
+load("@build_bazel_rules_apple//apple:ios.bzl",
+    "ios_framework"
+)
+ios_framework(
+    name = "fmwk.dynamic_framework",
```
keith, line 597: instead of adding to the shell tests which ideally we would get rid of can you add this under the targets_to_test BUILD files

brentleyjones:
What can we do to get this merged? I can't use `HEAD` right now (with Bazel 6.1, not using the new toolchain), because of this error.

test/ios_application_test.sh
codeman9: Added an `archive_contents_test`

keith (approved):

brentleyjones:
Thanks @codeman9!

PR #1904, opened 2023-03-16
Make emitswiftinterface parameter public

ma-oli:
On repositories where ios_static_framework() are built along side with other ios_*() targets, the emit-swiftinterface transition duplicates all the actions downstream of that transition.
This is by design, but it may not be desirable in all cases. For instance, if the swift code depends on a lot of cc_*() and objc_*() targets, it may be preferrable to not transition and add the swift.emit_swiftinterface feature to swift modules manually.

This changes makes the emitswiftinterface option to ios_static_framework() public so that the caller can opt-in to disable this transition and manage this feature manually.

### DIFF
--- a/apple/internal/rule_factory.bzl
+++ b/apple/internal/rule_factory.bzl
@@ -589,9 +589,9 @@ to manually dlopen the framework at runtime.
         })
     elif rule_descriptor.product_type == apple_product_type.static_framework:
         attrs.append({
-            "_emitswiftinterface": attr.bool(
+            "emitswiftinterface": attr.bool(
                 default = True,
-                doc = "Private attribute to generate Swift interfaces for static frameworks.",
+                doc = "Generate swift interfaces files (this triggers a config transition).",
             ),
             "hdrs": attr.label_list(
                 allow_files = [".h"],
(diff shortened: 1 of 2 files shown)


keith:
the intent of ios_static_framework is to be built for distribution, and therefore the assumption is that you want it to be independent of swift version with this file. in what case are you building ios_static_frameworks and objc_library targets in the same build where you're not distributing the static frameworks?

ma-oli:
The use-case is exactly like you describe, to build for distribution. But doing a config transition means that all your dependencies will be built under the new transition, and that includes not only the swift targets, but also all the cc_*() and objc_*() targets that are depended upon. The swift targets will be affected by this flag, but the others won't; yet, because the output path changes (because of the transition), they will still be built multiple times.
I guess what I'm asking is if it made sense to be able to opt-out from the transition (just to make one's build more efficient). Which means that in order to make the framework distributable one would need to add the swift.emit_swiftinterface feature to all the underlying swift dependencies.

keith:
yea interesting. it's a bummer that those other library targets inherit that configuration unnecessarily (although if they depend on swift_library targets it makes sense that they would have to be rebuilt as well). 

would you mark this as True only when doing distribution? in that case i guess an alternative would be to just not build this target unless you're doing distribution?

ma-oli:
Not really; what I was thinking was to disable the transition by default, and add the swift.emit_swiftinterface feature to individual swift_*() targets, either manually or through macros.
Maybe I'm an isolated case, so I understand if you don't want to merge this. But also; I kind of feel like although the primary use-case of ios_static_library() is to build libraries for distribution, it doesn't really need to make the flag mandatory to produce a valid static library. Hence the proposal to make opting-out of the transition possible.

keith:
i don't think that would entirely work since it wouldn't actually output the swiftinterface file? and i guess also if you missed 1 in the dep tree idk what would happen, might be UB.

But I think we should merge this anyways because i think there's a reasonable case to be made for if you want distribute your own frameworks internally, you might be willing to accept the fact that they have to be tied to a specific swift version, and at least if you mess that up the build will just fail. can you add a test validating that this works when set to false and doesn't generate the file?

PR #1905, opened 2023-03-17
Fix clangrttool if Xcode is installed in a non default location.

mjburghard:
In case multiple Xcode versions are installed of Xcode is not located in `/Applications`, `clang_lib_path` needs to be changed for the zipping to succeed.

### DIFF
--- a/tools/clangrttool/clangrttool.py
+++ b/tools/clangrttool/clangrttool.py
@@ -124,6 +124,10 @@ class ClangRuntimeTool(object):
           "Could not find any clang runtime libraries to package."
           "This is likely a configuration error")
 
+    if "DEVELOPER_DIR" in os.environ:
+      xcode_default_location = "/Applications/Xcode.app/Contents/Developer"
+      clang_lib_path = clang_lib_path.replace(xcode_default_location, os.environ["DEVELOPER_DIR"])
+
     with zipfile.ZipFile(out_path, "w") as out_zip:
       for lib in clang_libraries:
         full_path = os.path.join(clang_lib_path, lib)


mjburghard:
I also tried to use `xcode-select` to get the path for the current Xcode, but it seems it is not installed on all remote build environments, therefore going with `DEVELOPER_DIR`.
Not sure if we can test this with the CI setup here.

tools/clangrttool/clangrttool.py
```diff
@@ -124,6 +124,10 @@ def run(self):
           "Could not find any clang runtime libraries to package."
           "This is likely a configuration error")
 
+    if 'DEVELOPER_DIR' in os.environ:
```
BalestraPatrick, line 127: Minor style, but looks like we use double quotes here in general.

mjburghard: ✅

tools/clangrttool/clangrttool.py
```diff
@@ -124,6 +124,10 @@ def run(self):
           "Could not find any clang runtime libraries to package."
           "This is likely a configuration error")
 
+    if "DEVELOPER_DIR" in os.environ:
+      xcode_default_location = "/Applications/Xcode.app/Contents/Developer"
+      clang_lib_path = clang_lib_path.replace(xcode_default_location, os.environ["DEVELOPER_DIR"])
```
keith, line 129: is this default path always the one here? since it's extracted from the binary's rpaths won't it be just whatever absolute xcode path was used to build the thing?

mjburghard: You run into the issue if Xcode resides in different locations locally and on remote builders and the the script here is called in a different environment than when the binary was linked.
Thinking about it again, not all combinations are covered with this fix. Xcode-beta.app locally and Xcode-15.app on the remote builder would still run into the issue, given the actions run in the different environments.

keith: i think we could do some regex replace here to be more resilient. at the very least doing `/Applications/Xcode.*.app/Contents/Developer` would cover a lot more cases, but we could probably do something smarter too. Maybe even `.*/Contents/Developer`

PR #1911, opened 2023-03-17
Fix lldb when using `experimental_mixed_language_library`

brentleyjones:
This applies a similar fix that https://github.com/buildbuddy-io/rules_xcodeproj/commit/d40173d84764c1256b95a537b2871e010398f18c did, for the same reason.

### DIFF
--- a/apple/internal/experimental_mixed_language_library.bzl
+++ b/apple/internal/experimental_mixed_language_library.bzl
@@ -333,7 +333,7 @@ target only contains Objective-C files.""")
             # attribute to declare custom inputs.
             ":" + umbrella_module_map,
         ],
-        module_map = umbrella_module_map,
+        module_map = objc_module_map_name,
         srcs = objc_srcs,
         **kwargs
     )


brentleyjones:
I probably have to make a couple more tweaks, but the key thing is that only the objc custom modulemap ends up downstream `-Xcc` flags. I'll add a test around this behavior as well.

brentleyjones:
@thii Do you know of an example where this breaks?

thii:
I think `umbrella_module_map` is the correct one, otherwise an `objc_library` won't be able to see the interface of the underlying Swift module. (I forgot to add an example to check it.)

On why lldb breaks, I think it's because the generated Swift header always have this kind of import statement:
```
#import <MyMixedModule/MyMixedModule-Swift.h>
```

and if you don't have codebase structure that makes this work, or proper `-I` flags, or header maps, then lldb won't be able to find `MyMixedModule-Swift.h`.

brentleyjones:
I'll try the single module map method instead then Monday.

thii:
Edit: It was the Obj-C umbrella header import statement in the Swift generated header, not the Swift generated header import statement within itself.

Updated the above comment as well.

thii:
IIRC it was done like that so that we didn't have to change our import statements during the migration to Bazel. For example `@import MixedModule;` would also import the Swift interface of the mixed module. But if you only care about building with Bazel then this change makes sense.

Now thinking about it again, I guess this might have been the reason of many of our lldb issues in the past.

thii (approved):

brentleyjones:
Does this change work for you too @chiragramani?

chiragramani:
> Does this change work for you too @chiragramani?

We have a slightly different implementation, but I think the root cause you highlighted would also fix our case. I will check and let you know. Thanks!

aaronsky:
Closing this as `experimental_mixed_language_library` has been deleted from rules_apple.

PR #1913, opened 2023-03-20
Improve clangrt lib path normalization.

mjburghard:
Improvements as suggested by @keith [here](https://github.com/bazelbuild/rules_apple/pull/1905#discussion_r1140477111).

### DIFF
--- a/tools/clangrttool/clangrttool.py
+++ b/tools/clangrttool/clangrttool.py
@@ -21,9 +21,16 @@ be merged into an IPA product.
 """
 
 import os
+import re
 import subprocess
 import sys
 import zipfile
+
+def normalize_clang_lip_path(path, local_developer_dir):
+  return re.sub(".*\.app/Contents/Developer",
+                local_developer_dir,
+                path,
+                count=1)
 
 
 class ClangRuntimeToolError(RuntimeError):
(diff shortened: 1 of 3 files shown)


tools/clangrttool/clangrttool_unittest.py
```diff
@@ -0,0 +1,70 @@
+import os
+import unittest
+
+from tools.clangrttool.clangrttool import normalize_clang_lip_path, ClangRuntimeToolError
+
+clang_lib_path_common = "Toolchains/XcodeDefault.xctoolchain/usr/lib/clang/14.0.0/lib/darwin"
+xcode_developer_dir_default = "/Applications/Xcode.app/Contents/Developer"
```
brentleyjones, line 7: Can we add one that is named something different, like `ProjectBuilder.app`?

tools/clangrttool/clangrttool.py
```diff
@@ -25,6 +25,21 @@
...
 
+def normalize_clang_lip_path(path, local_developer_dir):
+  normalized_path_components = os.path.normpath(path).split(os.path.sep)
+
+  path_index = None
+  for i in range(len(normalized_path_components) - 1):
+      if (normalized_path_components[i] == "Contents" and
+        normalized_path_components[i+1] == "Developer"):
+        path_index = i
+        break
```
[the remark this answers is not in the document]
mjburghard: Just checked, the using a regex replacement looks simpler and should be easier to maintain. Will switch to it.

tools/clangrttool/clangrttool_unittest.py
mjburghard: See line 12 below.

brentleyjones (approved):

tools/clangrttool/clangrttool.py
```diff
@@ -21,10 +21,17 @@
 """
 
 import os
+import re
 import subprocess
 import sys
 import zipfile
 
+def normalize_clang_lip_path(path, local_developer_dir):
+  return re.sub(".*/Contents/Developer",
```
keith, line 30:
```suggestion
  return re.sub(".*\.app/Contents/Developer",
```

marginally safer, no need to take it if you don't want to

keith (approved):
is this ready to go?

tools/clangrttool/clangrttool.py
mjburghard: I'll take it.
And yes, ready to go.

keith:
so far the CI failure is an infra issue, it's automatically retrying slowly

PR #1918, opened 2023-03-21
Remove support to propagate frameworks via objc_library.runtime_deps

keith:
https://github.com/bazelbuild/rules_apple/commit/f711be032bd62fa13e866607bfb238e1988aca89 added support to propagate Apple frameworks using `data`
rule attribute for both objc_library and swift_library targets. This
change now removes support to propagate frameworks via `runtime_deps`.

PiperOrigin-RevId: 514762665
(cherry picked from commit 7d0ec138654c7009ca3fb00705d130d60ae50ee4)

### DIFF
--- a/apple/internal/aspects/BUILD
+++ b/apple/internal/aspects/BUILD
@@ -15,7 +15,6 @@ bzl_library(
     ],
     deps = [
         "//apple:providers",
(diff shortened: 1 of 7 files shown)


keith:
can't do this until we want to no longer support runtime_deps

PR #1926, opened 2023-03-24
Safer accessing of framework files xcframework

maxwellE:
Some imports can fail this conditional unless we
check for the existence of the attr on the
`xcframework_library`

Error log without this change

```terminal
ERROR: /private/var/tmp/_bazel_maxwellelliott/db37ab0e01ac1d9a8e5ea282dd04a720/external/Google-Mobile-Ads-SDK/BUILD.bazel:6:32: in apple_static_xcframework_import rule @Google-Mobile-Ads-SDK//:GoogleMobileAds: 
        File "/private/var/tmp/_bazel_maxwellelliott/db37ab0e01ac1d9a8e5ea282dd04a720/external/build_bazel_rules_apple/apple/internal/apple_xcframework_import.bzl", line 594, column 51, in _apple_static_xcframework_import_impl
                bundle_files = [x for x in xcframework_library.framework_files if ".bundle/" in x.short_path]
Error: 'struct' value has no field or method 'framework_files'
Available attributes: binary, clang_module_map, framework_imports, framework_includes, headers, includes, swiftmodule
```

### DIFF
--- a/apple/internal/apple_xcframework_import.bzl
+++ b/apple/internal/apple_xcframework_import.bzl
@@ -337,6 +337,7 @@ def _get_xcframework_library_with_xcframework_processor(
         includes = includes,
         clang_module_map = module_map_file,
         swiftmodule = [],
+        framework_files = [],
     )
 
 def _get_library_identifier(


mattrobmattrob (commented):
I also wonder if `_get_xcframework_library_with_xcframework_processor` should just add `framework_files = None` or `framework_files = []` to the response so that return struct is more similar across the `xcframework_library` cases.

maxwellE:
What is really odd is that I can only engage this error by using https://github.com/bazelbuild/rules_apple/commit/a55a5e3b8dd75557f6feded6a5a4ba929bcd8edb. If I dont use it then it works! Super strange

keith (approved):

PR #1940, opened 2023-04-10
Rewrite new test runner in swift

acecilia:
Benefits over existing bash script:
* Type safety
* Allows unit testing the code (no unit tests added in this PR though)
* Allows for more complex improvements in the future. Example: generate junit reports, generate html reports...

I tried to keep the Swift code as similar as possible to the bash code for this PR, to make review easier 🤞 
I believe the code can be improved/reorganised in upcoming PRs.

Already smoked-tested it in a big project: tried logic tests, unit tests and ui tests


apple/testing/default_runner/testrunner/TestRunnerCLI/Commands/SubCommand.swift
```diff
@@ -0,0 +1,121 @@
...
+            try TestAction().execute(input: input)
+        }
+    }
+}
+
+private struct TestArgs: ParsableCommand {
+    @Option(name: .customLong("simulator_name"), help: "The name of the simulator")
+    var simulatorName: String?
+
+    @Option(name: .customLong("xcodebuild_args"), help: "Arguments to be passed through to xcodebuild")
```
acecilia, line 88: * By default SwiftArgumentParser uses `-` instead of `_`
* But currently these flags are using `_`, so using the SwiftArgumentParser default `-` is a breaking change

For now I decided to avoid breaking changes, but maybe in the future make sense to align to avoid confusion. I was looking for a way to configure SwiftArgumentParser to use `_` (which historically I think is the most common), but so far could not find an easy/documented way

acecilia:
Kind of unrelated with this PR but, if I want to use [xchtmlreport](https://github.com/XCTestHTMLReport/XCTestHTMLReport) to produce html reports of the test runs, where do you think it should be integrated?
* My idea was to do the report generation from inside the test runner, after tests are executed. This will increate the test time though, as the test time reported by bazel will become `test time + report generation time`
* Is there any other better place for post-processing test outputs?

sergeykhliustin:
The sharding support could also benefit from swift implementation. So why is there no activity here?

acecilia:
Declining as there seems to be no interest on having the simulator written in swift. If in the future this changes, let me know

Issue #1943, opened 2023-04-12
ExtensionKit App Extensions not supported


maustinstar:
App extensions like [Background Assets](https://developer.apple.com/documentation/backgroundassets) are of type `EXAppExtension`, instead of `NSExtension`, and have some different build settings. An `EXAppExtension` is of product type `com.apple.product-type.extensionkit-extension` and installed into the `Extensions` directory instead of `PlugIns`.

Trying to configure a Background Asset downloader extension with current Bazel rules will run into validation errors, since `PlugIns` expect an `NSExtension` dictionary in Info.plist, but the extension point `com.apple.background-asset-downloader-extension` disallows that property.

For now, we can work around this simply by moving the `.appex` bundle from `PlugIns` to `Extensions`.

qyang-nj:
The new ExtensionKit Extension is very similar to the current App Extension, just with a few differences.
| | App Extension | ExtensionKit Extension |
| --- | --- | --- |
| xcode project type | com.apple.product-type.app-extension | com.apple.product-type.extensionkit-extension| 
| Info.plist | `NSEXtensions` | `EXAppExtensionAttributes` |
| bundle folder | *.app/PlugIns | *.app/Extensions |


There are two potential approaches to implement this:
1. Add a parameter (maybe called "extension_type"?) to the existing `ios_extension` rule, but it looks we have "one product type per rule" convention in `rules_apple`. 
2. Add a new rule called `ios_extensionkit_extension`. This will be extremely similar to `ios_extension`. We probably also need to rename `ios_extension` to `ios_app_extension` for claritrty. 

We're happy to implement this, but want to start the discussion beforehand. cc: @keith

maustinstar:
I noticed the `watchos_extension` has an [attribute to change the attribute type and linking flags](https://github.com/bazelbuild/rules_apple/blob/master/apple/internal/watchos_rules.bzl#L868-L872). It's used by setting `application_extension = True` on `watchos_extension`. We could follow a similar pattern with setting `extensionkit_extension = True`

BalestraPatrick:
Thanks for the extensive and detailed proposal! I think going with an attribute like you did in the linked PR probably makes sense. From what I understand, all extensions support both types except the new "Background Assets" which only supports `EXAppExtension`? We should make sure that we disallow making the mistake of creating a Background Assets extension but not setting `extensionkit_extension = True` (either if some Apple validation tool would fail during compilation, or in our rules).

[...]

PR #1944, opened 2023-04-14
Add support for ExtensionKit App Extensions with `extensionkit_extension = True`

maustinstar:
Related issue: #1943 

#### Plugin App Extensions

Before iOS 16.0, App Extensions were a product type `com.apple.product-type.app-extension`, build artifacts were installed into the app’s `PlugIns` directory, and the bundle’s Info.plist properties are defined in an `NSExtension` dictionary.

#### ExtensionKit App Extensions

iOS 16, macOS 13, tvOS 16, and watchOS 9, introduced [ExtensionKit](https://developer.apple.com/documentation/extensionkit) extensions for newer extensions like the [Background Asset Downloader](https://developer.apple.com/documentation/backgroundassets) and [App Intents](https://developer.apple.com/documentation/appintents/).

Distinct from the plugin-type app extensions, ExtensionKit extensions have a product type `com.apple.product-type.extensionkit-extension`, build artifacts are installed into the app’s `Extensions` directory, and the bundle’s Info.plist properties are defined in an `EXAppExtensionAttributes` dictionary.

#### Architecture Decisions

I noticed the `watchos_extension` has an [attribute to change the product type and linking flags](https://github.com/bazelbuild/rules_apple/blob/master/apple/internal/watchos_rules.bzl#L868-L872). It's used by setting `application_extension = True` on `watchos_extension`. This PR follows a similar pattern by setting `extensionkit_extension = True` to update the product type and bundle install location.

Alternatively, I considered introducing a separate rule for ExtensionKit app extensions, but this seems to make existing extension rules lose generalizability


BalestraPatrick:
Looks like we have this in the upstream branch now: https://github.com/bazelbuild/rules_apple/commit/3eb01c35694deb1fcbbc874885c648786b2f430d

PR #1948, opened 2023-04-20
Allow test bundles to override `bundle_name`

mattrobmattrob:
Tested this with both `ios_test_runner` & `ios_xctestrun_runner` on an unhosted unit test.


mattrobmattrob:
Happy to add a test that explicitly validates this use as well, just let me know.

keith:
can you add a small starlark test verifying this?

mattrobmattrob:
Let me know if that's different than what you were thinking, @keith.

keith (approved):

keith:
looks good!

PR #1952, opened 2023-04-27
update CREATE_XCRESULT_BUNDLE comment

maxwellE:
Users should set `create_xcresult_bundle` on
the runner instead

### DIFF
--- a/apple/testing/default_runner/ios_xctestrun_runner.bzl
+++ b/apple/testing/default_runner/ios_xctestrun_runner.bzl
@@ -182,8 +182,8 @@ ios_unit_test(
 ```
 
 If you would like this test runner to generate xcresult bundles for your tests,
-pass `--test_env=CREATE_XCRESULT_BUNDLE=1`. It is preferable to use the `create_xcresult_bundle`
-on the test runner itself instead of this parameter.
+pass `--test_env=CREATE_XCRESULT_BUNDLE=1`. It is preferable to use the
+`create_xcresult_bundle` on the test runner itself instead of this parameter.
 
 This rule automatically handles running x86_64 tests on arm64 hosts. The only
 exception is that if you want to generate xcresult bundles or run tests in
--- a/apple/testing/default_runner/ios_xctestrun_runner.bzl
+++ b/apple/testing/default_runner/ios_xctestrun_runner.bzl
@@ -181,6 +181,10 @@ ios_unit_test(
 )
 ```
 
+If you would like this test runner to generate xcresult bundles for your tests,
+pass `--test_env=CREATE_XCRESULT_BUNDLE=1`. It is preferable to use the `create_xcresult_bundle`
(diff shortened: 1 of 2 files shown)


keith:
I wonder if there's any case where folks might want it only temporarily, without changing their runner

maxwellE:
> I wonder if there's any case where folks might want it only temporarily, without changing their runner

Ahh that makes sense, I wondered if having two ways to do it would be confusing

keith:
yea i agree that the attribute should be preferred if possible. maybe we should augment the docs to just say that?

mattrobmattrob (commented):
We should remove this, IMO, if we're removing the comment. I'd rather it not be hidden functionality.

https://github.com/bazelbuild/rules_apple/blob/213e3ee3493547943ef7b72bac9eb3fe103ee600/apple/testing/default_runner/ios_xctestrun_runner.template.sh#L16-L19

I think there's value in the multi-input mechanism `--test_env` or runner attribute though.

keith (approved):

apple/testing/default_runner/ios_xctestrun_runner.bzl
```diff
@@ -182,7 +182,8 @@ ios_unit_test(
 ```
 
 If you would like this test runner to generate xcresult bundles for your tests,
-pass `--test_env=CREATE_XCRESULT_BUNDLE=1`
+pass `--test_env=CREATE_XCRESULT_BUNDLE=1`. It is preferable to use the `create_xcresult_bundle`
+on the test runner itself instead of this parameter.
```
mattrobmattrob, line 186: Line length nit:
```suggestion
pass `--test_env=CREATE_XCRESULT_BUNDLE=1`. It is preferable to use the
`create_xcresult_bundle` on the test runner itself instead of this parameter.
```

doc/rules-ios.md
```diff
@@ -660,7 +660,8 @@ ios_unit_test(
 ```
 
 If you would like this test runner to generate xcresult bundles for your tests,
-pass `--test_env=CREATE_XCRESULT_BUNDLE=1`
+pass `--test_env=CREATE_XCRESULT_BUNDLE=1`. It is preferable to use the `create_xcresult_bundle`
+on the test runner itself instead of this parameter.
```
mattrobmattrob, line 664: Line length nit:
```suggestion
pass `--test_env=CREATE_XCRESULT_BUNDLE=1`. It is preferable to use the
`create_xcresult_bundle` on the test runner itself instead of this parameter.
```

mattrobmattrob (approved):

maxwellE:
@mattrobmattrob looks like the CLA check is failing on your account

mattrobmattrob:
> @mattrobmattrob looks like the CLA check is failing on your account

Sigh, different email. Should be good now!

Issue #1954, opened 2023-05-03
Rules Apple Fails on Intel x86 Mac when using CpuInfo (Tensorflow)


alexbeattie42:
# Problem 
If you are attempting to use any rule from `@build_bazel_rules_apple` on an Intel x86_64 Mac with anything that relies on the [cpuinfo](https://github.com/pytorch/cpuinfo) project (like Tensorflow) you will face the build error described by this [mediapipe issue](https://github.com/google/mediapipe/issues/2324) and this [tensorflow issue](https://github.com/tensorflow/tensorflow/issues/50887).

## Root Cause

The root cause of the problem falls with the complex interplay between `cpuinfo`, `tensorflow`, `bazel`, and the `@build_bazel_rules_apple library`. Bazel has reverted the default cpu value on x86 macOS to "darwin" in this [commit](https://github.com/bazelbuild/bazel/commit/e96b8ca0c244e1a6a747b98b8f01e2526c9db862). 

This means that bazel expects Intel x86_64 Macs to report their cpu as `darwin`. `@build_bazel_rules_apple` will report the cpu as `darwin_x86_64`. This cpu does not exist in the `cpuinfo` project as it expects `darwin`. Although there is an unmerged [pull request](https://github.com/pytorch/cpuinfo/pull/139) to fix this behavior, it is unlikely to be resolved soon.

# Solution 

In order to fix the problem, I have created a patch so that the library can read the `host_cpu` value which will report as `darwin` on an Intel x86_64 Mac. If this value is present, it will return `darwin` instead of `darwin_x86_64`. 

## The Fix

It is necessary to add the `--incompatible_enable_apple_toolchain_resolution` flag to your build command and modify`@build_bazel_rules_apple` with the following patch:

```diff
diff --git a/apple/internal/transition_support.bzl b/apple/internal/transition_support.bzl
index 65f51b89..7da7ade3 100644
--- a/apple/internal/transition_support.bzl
+++ b/apple/internal/transition_support.bzl
@@ -102,6 +102,9 @@ def _cpu_string(*, cpu, platform_type, settings = {}):
             return "ios_sim_arm64"
         return "ios_x86_64"
     if platform_type == "macos":
+        host_cpu = settings["//command_line_option:host_cpu"]
+        if host_cpu == "darwin": 
+            return "darwin"
         if cpu:
             return "darwin_{}".format(cpu)
         macos_cpus = settings["//command_line_option:macos_cpus"]
[...]

PR #1955, opened 2023-05-03
Intel x86 Mac Cpu Fix

alexbeattie42:
Fix for this [issue](https://github.com/bazelbuild/rules_apple/issues/1954)


apple/internal/transition_support.bzl
```diff
@@ -102,6 +102,9 @@ def _cpu_string(*, cpu, platform_type, settings = {}):
             return "ios_sim_arm64"
         return "ios_x86_64"
     if platform_type == "macos":
+        host_cpu = settings["//command_line_option:host_cpu"]
+        if host_cpu == "darwin":
+            return "darwin"
```
keith, line 107: this should be below all the other conditions in this block, since --cpu and --macos_cpus should take precedence over this logic

alexbeattie42: I tried moving it below the `cpu` and `macos_cpu` options and both failed

keith: Doesn't this order mean users couldn't cross compile things then for darwin_arm64?

keith: right but if you are on your intel mac, where the host_cpu is `darwin`, and you pass `--cpu=darwin_arm64` to produce an output for an arm64 mac, i think this condition would take precedence

alexbeattie42: I can confirm that this change does break apple silicone mac. I'm also facing another [problem](https://github.com/bazelbuild/rules_foreign_cc/issues/1051) with building on apple silicone. Is there a better approach that could be used to solve this problem without breaking things for the above condition?

keith: you'd probably need to check the times we return darwin_x86_64 and return `darwin` in the case that the host CPU is just `darwin`, and not affect the return  value otherwise

PR #1967, opened 2023-05-18
optimize dtrace, support objc_library compile

kuperxu:
When you want to use dtrace's headers as dependency input for a compilable Target(objc_library...), DefaultInfo does not fit. You need to use CCInfo provider.

### DIFF
--- a/apple/dtrace.bzl
+++ b/apple/dtrace.bzl
@@ -52,7 +52,15 @@ def _dtrace_compile_impl(ctx):
             progress_message = ("Compiling dtrace probes %s" % (src.basename)),
         )
 
-    return [DefaultInfo(files = depset(output_hdrs))]
+    return [
+        apple_common.new_objc_provider(),
+        CcInfo(
+            compilation_context = cc_common.create_compilation_context(
+                headers = depset(output_hdrs),
+            ),
+        ),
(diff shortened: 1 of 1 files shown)


keith:
looks good! can you add a quick test case exercising this so we don't break it in the future?

kuperxu:
OK, I added the test

keith:
this looks good but would it be easier to test the end behavior of this being included in something instead? just incase the requirements there drift over time? I suppose that would mean putting it in the headers

kuperxu:
> this looks good but would it be easier to test the end behavior of this being included in something instead? just incase the requirements there drift over time? I suppose that would mean putting it in the headers

I have added a test case and tested it with internal test rules, please review it

keith (approved):

kuperxu:
> Looks good! Looks like CI is red though, might need a rebase too

sry,i forget submit compiled files

> Looks good! Looks like CI is red though, might need a rebase too

sry. I forgot to submit the compiled file dtrace_test.m for the dtrace test, please check it again

kuperxu:
> Looks good! Looks like CI is red though, might need a rebase too

sry. I forgot to submit the compiled file dtrace_test.m for the dtrace test, please check it again

keith:
Thanks!

PR #1973, opened 2023-05-23
use temporary keychain to resolve errors

yongjincho92:
Running security binary on Jenkins CI can result in permission error like in https://stackoverflow.com/questions/26474949/mac-security-command-needs-write-permissions-when-run-by-daemon.

As a resolution, create a temporary keychain file to run security binary with, and then remove it afterwards.

### DIFF
--- a/tools/codesigningtool/codesigningtool.py
+++ b/tools/codesigningtool/codesigningtool.py
@@ -20,6 +20,7 @@ import plistlib
 import re
 import subprocess
 import sys
+import tempfile
 
 from tools.wrapper_common import execute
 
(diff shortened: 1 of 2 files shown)


keith:
hrm i'm not sure the rules should manage this. in our setup we do this outside of the rules and set it as the default on CI, and then set it back after

yongjincho92:
> hrm i'm not sure the rules should manage this. in our setup we do this outside of the rules and set it as the default on CI, and then set it back after

Fair, I guess I could try to see if we can fix by some additional step in Jenkins

PR #2004, opened 2023-07-03
Make `test_host` required for `ios_ui_test`

louwers:
When I define a `ios_ui_test` without `test_host`, I get errors.

I think this attribute is required (which makes sense).


keith:
yep I think you're right, you'll have to change this in the .bzl file though

louwers:
Yes I created an issue when I realized that https://github.com/bazelbuild/rules_apple/issues/2005

PR #2024, opened 2023-07-17
Fix building examples/...

keith:


luispadron (approved):

brentleyjones:
Why are we excluding so many from CI?

brentleyjones:
Ahh, they are libraries. nvm

brentleyjones (approved):

PR #2029, opened 2023-07-22
Force _main symbol for application extensions

keith:
In the case you're linking with LTO, the `_main` does not get loaded
correctly because it seems that `-Objc` doesn't force static libraries
to be loaded in the case the objects in the archive are bitcode. This
makes sure the main file is loaded and that can cascade to the rest of
the used code. This is only needed because `-e _NSExtensionMain` makes
the linker not need to load the standard `_main`.

https://github.com/apple/swift/issues/67460


keith:
sounds like there are some extension types that won't have a `_main` so this won't work. Folks can use this manually in the meantime with `linkopts = ["-Wl,-u,_main"],` on their `ios_extension` targets if they only have extension types that do require `_main`. The long term solution instead will be flipping the `alwayslink` default to `True` which is in progress

PR #2032, opened 2023-07-27
Remove _xcrunwrapper

keith:
https://github.com/bazelbuild/bazel/commit/35ed67452237c12a6a1dcecaa25f320eb956936d


keith:
kinda hoping this fails because i think old versions of bazel require this still and otherwise that might mean we don't have any coverage on codepaths that use this

keith:
guess it hasn't been used in 6.x

mattrobmattrob (approved):

PR #2044, opened 2023-08-11
Add support for importing Swift XCFrameworks through Swift interface files

keith:
Previously, support for importing Swift based XCFrameworks was limited
to Swift interoperability with Objective-C targets through imported Clang
module map files. This change adopts recently introduced swift_common
API to compile Swift module interface files imported by XCFrameworks.

(cherry picked from commit 3ed9049c0fd7454c4b806ca200962d41c51718a0)


keith:
depends on https://github.com/bazelbuild/rules_swift/pull/1085

brentleyjones (approved):

PR #2050, opened 2023-08-14
Add test validating builds with opt + objc_enable_binary_stripping

keith:
This was previously not covered by our test suite and has some codepaths
in bazel that can break depending on our changes


keith:
this should fail until i revert xcrunwrapper

keith:
expected failure: https://buildkite.com/bazel/rules-apple-darwin/builds/7634#0189f521-23bd-41c1-9011-515a321ce19b

keith:
this reverts the xcrunwrapper removal which is still required but only in a config we weren't testing

PR #2061, opened 2023-08-14
Move platform specific framework import targets into platform/BUILD

keith:
This is to be consistent with other framework import targets across
Apple platforms (i.e. watchOS, tvOS all have their framework import
fixture targets at their own targets_under_test package)

PiperOrigin-RevId: 474672209
(cherry picked from commit b6aad354fe41d55edabbbe92c10d9baefe8e5c7a)


keith:
going to add some of ours to this too

brentleyjones (approved):

PR #2064, opened 2023-08-15
Move XCFramework import test targets into starlark_tests.

keith:
This change moves remaining XCFramework import test targets into
iOS targets under test package, and leaves analysis testing only
targets under Apple package.

PiperOrigin-RevId: 475610495
(cherry picked from commit d4c58bd6d253a113029ab36956dc2d93be9c54ec)


brentleyjones (approved):

thii (approved):

test/testdata/xcframeworks/generate_xcframework.bzl
```diff
@@ -148,6 +148,7 @@ def _create_xcframework(
         ))
 
     args.extend(["-output", xcframework_directory])
+    args.extend([";ls", "-R", xcframework_directory])
```
keith, line 151: need to revert before merge

PR #2068, opened 2023-08-15
Add swiftdocs to imported frameworks with swiftinterface files

keith:
swiftdoc files are produced when building things with
-enable-library-evolution. If downstream IDE integrations want to grab
them they should be propagated correctly so that that's possible.


keith:
depends on https://github.com/bazelbuild/rules_swift/pull/1088

PR #2072, opened 2023-08-16
Fix black screen when using `disable_signing_resource_rules` and launch storyboard

chatwyn:
Hello maintainers,

I'm submitting this pull request to address an issue that occurs when my storyboard is set as the startup screen and I utilize signed acceleration. In this scenario, if I cold launch the application on the simulator a black screen issue arises. To mitigate this problem, I suggest extending the signing process to include the storyboard as well.

## Steps to Reproduce:
Set the storyboard as the startup screen.
Utilize signed acceleration for the startup process.
Launch the application on the simulator.

## Expected Behavior:
The application should launch properly without any black screen issues.

## Actual Behavior
A black screen appears during the startup process on the simulator.

## Proposed Solution:
To address this issue, I recommend extending the code signing process to include the storyboard. By signing both the main application and the storyboard, we can ensure consistent behavior during the startup process, preventing the occurrence of the black screen problem.

Best regards,
caowenbo

### DIFF
--- a/tools/codesigningtool/disable_signing_resource_rules.plist
+++ b/tools/codesigningtool/disable_signing_resource_rules.plist
@@ -6,6 +6,8 @@
 	<dict>
 		<key>.*</key>
 		<false/>
+		<key>.*storyboard</key>
+		<true/>
 	</dict>
 </dict>
 </plist>


keith:
interesting case! I tested this locally on our app where we use this as well and it didn't seem to fix the issue for me. does it reliably fix it for you? could it have something to do with just relaunching the same app instead? removing the resource rules fixes it for me but not this change

keith:
could it be that compiled storyboards actually have different extensions than this? `Lyft.app/AppLaunchUIResources.bundle/Base.lproj/LaunchScreen.storyboardc/LaunchScreen.nib`

keith:
seems like doing `.*nib` might be working in our case

chatwyn:
Sure. when I add it in the resource rules, it can solve my problem.
And In MY `*.app/_CodeSignatureCodeResources` will show the nib name, such as <key>StartLogo.storyboardc/01J-lp-oVM-view-Ze5-6b-2t3.nib</key>
		<data>
		ZX9hKRVSyKfPkYmyQU8Qi+FMXg8=
		</data>

brentleyjones:
Does a `*.storyboardc` based key work? Might need to adjust to to apply to sub-files.

PR #2074, opened 2023-08-16
Inline executable_name reading

keith:
In a future cherry pick we will no longer be able to get the bundle_name
with just the ctx argument as we do for the executable_name fallback.
Instead of passing along all the new required arguments this change
inlines the executable_name attr reading and passes that along, and
makes the consumers of that arg do the behavior of
`executable_name or bundle_name` themselves.

### DIFF
--- a/apple/internal/macos_binary_support.bzl
+++ b/apple/internal/macos_binary_support.bzl
@@ -78,7 +78,6 @@ def _macos_binary_infoplist_impl(ctx):
     actions = ctx.actions
     bundle_name, bundle_extension = bundling_support.bundle_full_name_from_rule_ctx(ctx)
     bundle_id = ctx.attr.bundle_id
-    executable_name = bundling_support.executable_name(ctx)
     rule_descriptor = rule_support.rule_descriptor(ctx)
     rule_label = ctx.label
     platform_prerequisites = platform_support.platform_prerequisites_from_rule_ctx(ctx)
@@ -105,7 +104,6 @@ def _macos_binary_infoplist_impl(ctx):
         bundle_extension = bundle_extension,
         bundle_id = bundle_id,
         bundle_name = bundle_name,
-        executable_name = executable_name,
         environment_plist = ctx.file._environment_plist,
         include_executable_name = False,
         input_plists = infoplists,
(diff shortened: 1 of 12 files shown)


apple/internal/macos_binary_support.bzl
```diff
@@ -78,7 +78,6 @@ def _macos_binary_infoplist_impl(ctx):
     actions = ctx.actions
     bundle_name, bundle_extension = bundling_support.bundle_full_name_from_rule_ctx(ctx)
     bundle_id = ctx.attr.bundle_id
-    executable_name = bundling_support.executable_name(ctx)
```
keith, line 81: this rule didn't have the executable_name attr, but this function was forgiving about that

apple/internal/macos_rules.bzl
```diff
@@ -1714,7 +1712,6 @@ def _macos_dylib_impl(ctx):
     apple_mac_toolchain_info = ctx.attr._mac_toolchain[AppleMacToolsToolchainInfo]
     apple_xplat_toolchain_info = ctx.attr._xplat_toolchain[AppleXPlatToolsToolchainInfo]
     bundle_name, bundle_extension = bundling_support.bundle_full_name_from_rule_ctx(ctx)
-    executable_name = bundling_support.executable_name(ctx)
```
keith, line 1717: same as above

apple/internal/resource_actions/plist.bzl
```diff
@@ -275,7 +275,8 @@ def merge_root_infoplists(
     # Info.plists out of the box coming from Xcode.
     substitutions["DEVELOPMENT_LANGUAGE"] = "en"
 
-    if include_executable_name and executable_name:
+    executable_name = executable_name or bundle_name
+    if include_executable_name:
```
mattrobmattrob, line 279: This is a behavior change, right?

If it's desired, can we do this?
```suggestion
    if include_executable_name:
        executable_name = executable_name or bundle_name
```
If it's not, how about this?
```suggestion
    executable_name = executable_name or bundle_name
    if include_executable_name and executable_name:
```

brentleyjones (approved):

mattrobmattrob (approved):

apple/internal/resource_actions/plist.bzl
keith: i don't think so because previously every declared executable_name always fell back to bundle_name, and i don't think it's valid to not have a bundle_name?

PR #2104, opened 2023-08-18
Add the `apple.no_legacy_swiftinterface` feature to disable the automatic library evolution transition on framework rules.

keith:
We are migrating users to a model where a new attribute, `library_evolution`, must be set on `swift_library` targets that vend the user-facing module in a framework, instead of automatically applying library evolution to the whole subgraph. In order to migrate, users can add this feature to their framework target and at the same time set `library_evolution = True` on their library.

Once everyone is migrated, library evolution should be removed entirely from the Apple transitions and the feature can be removed from BUILD files.

PiperOrigin-RevId: 487022298
(cherry picked from commit cb9a1de11217a616c31a27ee52c6333425b55e46)


keith:
needs this in rules_swift https://github.com/bazelbuild/rules_swift/commit/b9175e66bb1d0e0f8eda6afdb413c685ce7943d7

brentleyjones:
We need to land this in order to land #2401.

PR #2115, opened 2023-08-18
Add/omit the HeadersPath Info.plist key for XCFramework with static libraries.

keith:
An XCFramework with static libraries can include public
Objective-C(++) headers using the `public_hdrs` rule attribute,
or generated headers from the underlying Swift module.

This change updates the XCFramework Info.plist generation logic
to add or omit the `HeadersPath` key if there are any header files.

PiperOrigin-RevId: 499574821
(cherry picked from commit e00404ac6a4d1949674e6ed9e47d456660cd5d26)


keith:
the actual change here wasn't relevant because we produce frameworks not libraries for this rule, so I'm not sure if we'll take these tests or not

PR #2120, opened 2023-08-18
Remove unused targets.

keith:
And remove a stale comment.

PiperOrigin-RevId: 510440204
(cherry picked from commit 044f3f26abc4120a666ac70ef3fadbe6cf254d62)


keith:
i doubt this is valid for us

keith:
nbd, we don't need to take

PR #2130, opened 2023-08-18
Show example of failing label issue

keith:


fmeum:
This strict has more labels that aren't mapped in this way, can they be the cause of this failure?

keith:
updated

fmeum:
The remaining failure is due to https://github.com/bazelbuild/rules_apple/blob/681ab8e588a3e19a575f57bf68493500fd45463b/test/starlark_tests/apple_dynamic_xcframework_import_tests.bzl#L191-L193, which you also need to fix in this way.

brentleyjones:
https://github.com/bazelbuild/rules_apple/pull/2263

PR #2131, opened 2023-08-18
Fix test outputs path for external only change

keith:
Not sure how this passed CI


keith:
(it failed on master so probably just a merge race somehow)

PR #2164, opened 2023-08-22
HACK: fix swiftinterface import support

keith:
This fixes passing `-F` and `-fmodule-map-file` to the compile for the
case that your module is multi-language. It also handles passing the
right module_name in the case the .swiftmodule differs from the
.framework bundle in name (not sure this should work, it doesn't work in
Xcode).


keith:
see also https://github.com/bazelbuild/rules_swift/pull/1092

PR #2165, opened 2023-08-23
Always set apple_split_cpu

keith:
This can easily be derived from the cpu_string and reduces duplicate
transitions that only differ based on this


brentleyjones:
Do we need the `environment_arch if environment_arch else ""` part anymore?

keith:
I don't think so, fixed!

brentleyjones (approved):

PR #2180, opened 2023-08-24
Only use label based archive roots if needed

keith:
This changes the logic that outputs bundles to label-based directories
to only apply if the label name and the bundle name differ, which is the
best approximation for "whether or not this is needed" without
introducing a switch for it. The only downside is the "primary" target
that has the matching label and bundle name uses a different output
directory structure from the other ones. At least in the case of using
separate labels for debug vs alpha etc all the targets probably differ.


apple/internal/outputs.bzl
```diff
@@ -47,11 +47,14 @@ def _archive(
         platform_prerequisites = platform_prerequisites,
     )
     if tree_artifact_enabled:
-        archive_relative_path = rule_descriptor.bundle_locations.archive_relative
-        root_path = label_name + "_archive-root"
-        return actions.declare_directory(
-            paths.join(root_path, archive_relative_path, bundle_name_with_extension),
-        )
+        if bundle_name != label_name:
```
keith, line 50: key change

apple/internal/partials/debug_symbols.bzl
```diff
@@ -196,7 +196,10 @@ def _bundle_dsym_files(
       artifact representation of a .dSYM bundle with the binaries lipoed together as one binary.
     """
     dsym_bundle_name_with_extension = debug_output_filename + bundle_extension
-    dsym_bundle_name = paths.join(label_name + "_dsyms", dsym_bundle_name_with_extension + ".dSYM")
+    if debug_output_filename != label_name.split(".__internal__.")[0]:
```
keith, line 199: key change

test/ios_application_test.sh
```diff
@@ -539,39 +539,4 @@ function test_tree_artifacts_and_disable_simulator_codesigning() {
       --features=apple.skip_codesign_simulator_bundles || fail "Should build"
 }
 
-# Tests tree artifacts of bundle targets(with same bundles names) don't conflict with each other.
```
keith, line 542: moved this to starlark

brentleyjones (approved):

PR #2184, opened 2023-08-24
Move to new provider API initialization pattern

keith:
This makes it clear that constructing these providers is meant to  be private


keith:
going to stack the other changes around this in here too

brentleyjones (approved):

keith:
need to do ours too

PR #2185, opened 2023-08-24
Take upstream runner dsym changes

keith:
Seems that the goal of these changes is to get macOS to discover the dsyms when running directly through bazel

### DIFF
--- a/apple/internal/templates/apple_simulator.template.py
+++ b/apple/internal/templates/apple_simulator.template.py
@@ -67,7 +67,7 @@ AppleSimulatorUDID = collections.abc.Generator[str, None, None]
 logging.basicConfig(
     format="%(asctime)s.%(msecs)03d %(levelname)s %(message)s",
     datefmt="%Y-%m-%d %H:%M:%S",
-    level=logging.DEBUG,
+    level=logging.INFO,
 )
 logger = logging.getLogger(__name__)
 
@@ -484,6 +484,10 @@ def register_dsyms(dsyms_dir: str):
   """
   symbolscache_command = [
       "/usr/bin/symbolscache",
+      "delete",
+      "--tag",
+      "Bazel",
+      "compact",
       "add",
       "--tag",
       "Bazel",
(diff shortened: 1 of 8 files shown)


apple/internal/templates/apple_simulator.template.py
```diff
@@ -444,37 +454,71 @@ def temporary_simulator(
...
 
 
+def register_dsyms(dsyms_dir: str):
+  """Adds all dSYMs in `dsyms_dir` to the symbolscache.
+
+  Args:
+    dsyms_dir: Path to directory potentially containing dSYMs
+  """
+  symbolscache_command = [
+      "/usr/bin/symbolscache",
```
brentleyjones, line 486: Interesting. We might want to do something like this in rules_xcodeproj.

brentleyjones (approved):

apple/internal/templates/apple_simulator.template.py
keith: theoretically shouldn't be needed, but some folks at wwdc also recommended this to me when instruments didn't symbolicate things

apple/internal/templates/apple_simulator.template.py
```diff
@@ -444,37 +454,71 @@ def temporary_simulator(
...
+      "delete",
+      "--tag",
+      "Bazel",
+      "compact",
+      "add",
+      "--tag",
+      "Bazel",
+  ] + [
+      a
+      for a in pathlib.Path(dsyms_dir).glob(
```
mattrobmattrob, line 496: Is this glob ever going to be empty?

mattrobmattrob (approved):

apple/internal/templates/apple_simulator.template.py
keith: yea if your build command doesn't generate dsyms i think it will be

keith: when it's empty symbolscache just deletes in that case:

```
2023-08-25 11:36:31.979 INFO Running command: ['/usr/bin/symbolscache', 'delete', '--tag', 'Bazel', 'compact', 'add', '--tag', 'Bazel']
2023-08-25 11:36:32.119 INFO symbolscache output:   No entries deleted
```

PR #2189, opened 2023-08-25
Add rules_python for dep

keith:


keith:
just an example in case we ever need it

PR #2205, opened 2023-08-28
Take upstream j2objc change

keith:
This isn't used externally but helps reduce conflicts

(cherry picked from commit c97983d3f09086a29f7346cca3f33bf48f870ac4)


brentleyjones (approved):

keith:
seems like this doesn't work w/o more changes, we can just ignore unless it becomes an issue

PR #2220, opened 2023-08-28
Assume target environment is device

keith:
This makes migrating to the 3.x rules easier since not all old platform
definitions had device in the non-simulator case. For example macOS
platforms like `darwin_x86_64` didn't have a simulator or device
constraint.


keith:
this is for discussion https://bazelbuild.slack.com/archives/C04DFUBQGSU/p1693264708205309

brentleyjones (approved):

PR #2222, opened 2023-08-29
Fix script error resulting in xcodebuild always being used

erikkerber:
`build_for_device`, like `create_xcresult_bundle`, is always true or false, but never unset.

This was leading `xcodebuild` to be used for all tests as of 2.5.0


mattrobmattrob:
I was wondering about this just the other day after doing a bit of work in the template. Thank you!

mattrobmattrob (approved):

PR #2224, opened 2023-08-30
Merge 3.0/port migrations to visionOS

erikkerber:
@keith This merges `master`/ eb04fbbc into your visionOS branch. I'm using it to rev on something on our end, and of course if it's useful merge away as well. Most of the work was some pretty basic diffing of the other rules against visionOS.

`bazel test //...` builds all the way to `...does not contain a toolchain for cpu visionos_x86_64`, but I haven't checked the tests e2e quite yet.


keith:
thank you! I took this and squashed it into my PR

PR #2226, opened 2023-08-30
Fix default environment_archs

keith:
Upstream just defaulted to x86_64 but we want to choose arm64 if the
host is arm64 in that case.


keith:
fixes https://bazelbuild.slack.com/archives/CD3QY5C2X/p1693404242922179

jszumski (approved):

mattrobmattrob (approved):

PR #2232, opened 2023-09-04
Fix action_env for ipa post processor

MIP9:
Environment variables that provided by `--action_env` flag are not accessible on ipa_post_processor .sh script.

There is fix for it.


thii:
@MIP9 Can you sign the CLA?

thii:
Can you rebase off master? Looks like buildifier is reporting outdated warnings from an old revision.

Other CI issues are probably related to the current GitHub's incident.

acecilia:
@keith Does this PR still make sense? Or do you think better leave as is and wait for bazel 6.4.0?

keith:
yea that's a good question, can you test with 6.4 and see  if it really does fix it?

acecilia:
Seems [the cherrypick did not make it to 6.4.0 yet](https://github.com/bazelbuild/bazel/pull/19319), and the commit [has not been built with bazel ci](https://buildkite.com/bazel/bazel-bazel/builds?branch=release-6.4.0) (so there is no prebuilt bazel binary), so I am not sure how to test it atm 😅

thii:
If you use bazelisk you can set `last_green` as a version and try that.

thii:
Or build a bazel binary with that patch.

keith:
note that with bazel 7.x HEAD of rules_apple now enables that env for all actions (or tries too, possible I missed some)

PR #2233, opened 2023-09-05
Ignore more variants of codesigning output

keith:
I don't think users ever need to see any "replacing existing signature"
message. The previous regex didn't match this case:

```
bazel-out/...__internal__.__test_bundle_archive-root/Foo.xctest/: replacing existing signature
```

### DIFF
--- a/tools/codesigningtool/codesigningtool.py
+++ b/tools/codesigningtool/codesigningtool.py
@@ -37,7 +37,7 @@ from tools.wrapper_common import execute
 # * Executable=/{path to signed target}
 # * using the deprecated --resource-rules flag
 _BENIGN_CODESIGN_OUTPUT_REGEX = re.compile(
-    r"(signed.*Mach-O (universal|thin)|libswift.*\.dylib: replacing existing signature|signed generic|Executable=/|Warning: --resource-rules has been deprecated)"
+    r"(: replacing existing signature|signed generic|Executable=/|Warning: --resource-rules has been deprecated)"
 )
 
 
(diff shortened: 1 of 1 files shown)


jszumski (approved):
Thanks! This fixes a bunch of noise in our graph

keith:
awesome! are there any other noisy ones like this?

mattrobmattrob (approved):

jszumski:
I'm also seeing a lot of this for both freshly built targets:

```
INFO: From Bundling, processing and signing SomeModule.__internal__.__test_bundle:
bazel-out/ios-sim_arm64-min14.0-applebin_ios-ios_sim_arm64-dbg-ST-a339f48a4533/bin/path/to/SomeModule.__internal__.__test_bundle_archive-root/SomeModule.xctest/Frameworks/libswift_Concurrency.dylib: signed Mach-O thin (arm64) [com.apple.dt.runtime.swiftConcurrency]
```

and imported binary xcframeworks:

```
INFO: From ImportedDynamicFrameworkProcessor path/to/SomeModule-intermediates/_imported_frameworks/BinaryFramework.framework.zip:
bazel-out/ios-sim_arm64-min14.0-applebin_ios-ios_sim_arm64-dbg-ST-a339f48a4533/bin/path/to/SomeModule-intermediates/_imported_frameworks/BinaryFramework.framework: signed bundle with Mach-O thin (arm64) [com.some.binary.framework]
```

keith:
is that with or without this change?

keith:
my first push here broke that, but theoretically before this change those should have been ignored (assuming they go through this codepath). I just fixed it here too

jszumski:
> is that with or without this change?

With.  Giving it another try now

jszumski:
Both "signed bundle with Mach-O thin" cases and the original "replacing existing signature" output are gone with 66ce20d 👍

PR #2242, opened 2023-09-07
Fix other rules table in docs README

luispadron:


keith (approved):
looks good, if you want to improve this more you can sort it, that is probably also a nicer order anyways since versioning.bzl isn't as important as apple.bzl

luispadron:
@keith Sorted all the items in the rows and the `General` rows as well

keith (approved):
thanks!

thii (approved):

PR #2243, opened 2023-09-07
Fix importing of versioned (macOS) dynamic frameworks

brentleyjones:


brentleyjones:
Draft because I think we should fix whatever upstream was trying to do instead? At least for the python stuff?

apple/internal/apple_framework_import.bzl
```diff
@@ -213,21 +213,7 @@ def _apple_dynamic_framework_import_impl(ctx):
...
-    )
-    tree_artifact_enabled = (
-        apple_xplat_toolchain_info.build_settings.use_tree_artifacts_outputs or
-        is_experimental_tree_artifact_enabled(config_vars = ctx.var)
-    )
-    if target_triplet.os == "macos" and has_versioned_framework_files and tree_artifact_enabled:
-        fail("The apple_dynamic_framework_import rule does not yet support versioned " +
-             "frameworks with the experimental tree artifact feature/build setting. " +
-             "Please ensure that the `apple.experimental.tree_artifact_outputs` variable is not " +
-             "set to 1 on the command line or in your active build configuration.")
```
brentleyjones, line 230: Based on the comment, I assume this was only disabled because of RBE support. We could instead add `no-remote-exec` to the execution requirements when `tree_artifact_enabled`?

tools/imported_dynamic_framework_processor/imported_dynamic_framework_processor.py
```diff
@@ -306,75 +306,16 @@ def main() -> None:
...
-    versions_dir = os.path.join(args.temp_path, "Versions")
-    version_dir = os.path.join(versions_dir, version)
-    for entry in os.listdir(version_dir):
-      symlink_path = os.path.join(args.temp_path, entry)
-      symlink_data = os.path.join("Versions", "Current", entry)
-      os.symlink(symlink_data, symlink_path)
-
-    # Modify codesigningtool arg to sign the current framework version
-    # This matches Xcode behavior of re-signing only the effective version.
-    args.target_to_sign = [version_dir]
```
brentleyjones, line 377: Whatever this was trying to do, it created invalid bundles. I tried a couple different edits, but just using the other path works for my test case.

brentleyjones: Tests are failing though. So I wonder if the frameworks the test are creating are the ones that match this, while the format for framework I have is something else (I believe I took it from Xcode itself). Here is my framework I'm importing:

```
ExampleFramework.framework
├── ExampleFramework
├── Headers
│   ├── Bar.h
│   └── ExampleFramework.h
├── Modules
│   └── module.modulemap
├── Resources
│   └── Info.plist
├── Versions
│   └── A
│       └── ExampleFramework -> ../../ExampleFramework
└── _CodeSignature
    └── CodeResources
```

PR #2255, opened 2023-09-19
Use platform_mappings in `bazel_testrunner.sh`

brentleyjones:


brentleyjones:
Ideally we could copy the main one, because we are going to forget to update this.

keith (approved):

PR #2256, opened 2023-09-19
Use platforms inside Apple rules transitions

keith:
This makes C++ toolchain selection work without using platform mapping for Apple rules.

PiperOrigin-RevId: 561560820


keith:
rebased version of https://github.com/bazelbuild/rules_apple/pull/2227/

brentleyjones (approved):

PR #2259, opened 2023-09-20
Fix error swallowing

brentleyjones:
If `exit_status` is set in a subshell, the value is lost.


brentleyjones:
@keith Was this the real issue? Maybe we don't need the text checking now?

keith (approved):
looks good, wasn't the core issue since i ran the thing outside of bazel and observed the same behavior

PR #2265, opened 2023-09-26
Use `build_settings_labels` everywhere

brentleyjones:

PR #2266, opened 2023-09-26
Correctly add `.swiftinterface` files to imported dynamic xcframework `public_hdrs`

brentleyjones:
5926b990842514cea21cb27f2ca70b70179bb4e8 broke importing xcframeworks that have swiftinterface files when using `parse_xcframework_info_plist`. We didn’t notice it before because the tests were using an invalid build setting name (https://github.com/bazelbuild/rules_apple/pull/2265).

This change prevents the `.swiftinterface` files from breaking in `xcframework_processor_tool.py`, while still including them in the `CcInfo`.

I mimicked where we were setting `swiftmodule_imports` before (in particular I didn’t set it for `xcframework_static_import`). It seems we might be over setting it in some places (or under-setting it there).

### DIFF
--- a/apple/internal/apple_framework_import.bzl
+++ b/apple/internal/apple_framework_import.bzl
@@ -281,6 +281,7 @@ def _apple_dynamic_framework_import_impl(ctx):
         kind = "dynamic",
         label = label,
         libraries = [] if ctx.attr.bundle_only else framework.binary_imports,
+        swiftinterface_imports = framework.swift_interface_imports,
         swiftmodule_imports = framework.swift_module_imports,
     )
     providers.append(cc_info)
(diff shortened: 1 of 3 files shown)


apple/internal/apple_framework_import.bzl
```diff
@@ -434,6 +435,7 @@ def _apple_static_framework_import_impl(ctx):
             label = label,
             libraries = framework.binary_imports,
             linkopts = linkopts,
+            swiftinterface_imports = framework.swift_interface_imports,
```
brentleyjones, line 438: This wasn't needed by tests, but I'm not sure if we fully test the path that would need it.

apple/internal/apple_framework_import.bzl
```diff
@@ -281,6 +281,7 @@ def _apple_dynamic_framework_import_impl(ctx):
         kind = "dynamic",
         label = label,
         libraries = [] if ctx.attr.bundle_only else framework.binary_imports,
+        swiftinterface_imports = framework.swift_interface_imports,
```
brentleyjones, line 284: This wasn't needed by tests, but I'm not sure if we fully test the path that would need it.

thii (approved):

Issue #2278, opened 2023-10-03
`xcarchive` rule produces invalid bundleinfo entry for `ApplicationPath`


maxwellE:
When producing xcarchives via the `xcarchive` rule the resulting `*.xcarchive/Info.plist` contains the following entry

```
		<key>ApplicationPath</key>
		<string>Products/Applications/SomeApp.app</string>
```

Xcode expects the Products component naturally, so it can be updated to 

```
		<key>ApplicationPath</key>
		<string>Applications/SomeApp.app</string>
```

This should unblock processing of xcarchives by certain services

luispadron:
To test this can we just open the .xcarchive in Xcode and see if that fails?

PR #2280, opened 2023-10-03
Change xcarchive ApplicationPath

maxwellE:
Removes an unused component `Products` from the info.plist of the built xcarchive. This change allows xcarchives to be processed by certain services

Related #2278


maxwellE:
I was still able to open the built xcarchive in xcode after this change

luispadron (approved):

thii (approved):

Issue #2279, opened 2023-10-03
`xcarchive` rule does not package linkmaps


maxwellE:
The `xcarchive` rule should package in linkmaps if they were built for the bundle in question

luispadron:
Closing this as completed by #2282

PR #2282, opened 2023-10-04
Propagate linkmaps into `xcarchives`

maxwellE:
These files should be propagated into xcarchives if enabled

Related #2279

### DIFF
--- a/tools/xcarchivetool/make_xcarchive.py
+++ b/tools/xcarchivetool/make_xcarchive.py
@@ -35,6 +35,12 @@ def _build_parser() -> argparse.ArgumentParser:
     action="append",
     help="The file path to a dSYM file which will be packaged in the .xcarchive.",
   )
+  parser.add_argument(
+    "--linkmap",
+    required=False,
+    action="append",
+    help="The file path to a linkmap file which will be packaged in the .xcarchive.",
+  )
   return parser
 
 
@@ -84,7 +90,8 @@ def _main(
     infoplist_path: Path,
     bundle_path: Path,
     output_path: Path,
-    dsyms: list):
+    dsyms: list,
+    linkmaps: list):
   """Creates a .xcarchive from the given bundle.
 
   <BundleName>.xcarchive
@@ -95,12 +102,15 @@ def _main(
   PkgInfo
   dSYMs/
     <BundleName>.app.dSYM
+  Linkmaps/
+    <BundleName>.txt
 
   Args:
     infoplist_path: The file path to the Info.plist file for the app.
     bundle_path: The file path to the bundle.
     output_path: The file path to the output .xcarchive.
     dsyms: The file paths to the dSYM files which will be packaged in the .xcarchive.
+    linkmaps: The file paths to the linkmap files which will be packaged in the .xcarchive.
   """
 
   # Collect the required information for the .xcarchive plist
@@ -134,6 +144,13 @@ def _main(
       dsym,
       os.path.join(dsyms_path, os.path.basename(dsym)),
       dirs_exist_ok=True)
+
+  # Copy the linkmap files into the .xcarchive.
+  linkmaps_path = os.path.join(output_path, "Linkmaps")
+  for linkmap in linkmaps or []:
+    shutil.copyfile(
+      linkmap,
+      os.path.join(linkmaps_path, os.path.basename(linkmap)))
 
   # Create the Info.plist for the .xcarchive.
   creation_date = datetime.datetime.now().isoformat()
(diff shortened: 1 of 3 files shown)


luispadron (approved):
Sweet!

mattrobmattrob (approved):
Can you add anything to `test/starlark_tests/xcarchive_tests.bzl` to test this new functionality?

maxwellE:
> Can you add anything to `test/starlark_tests/xcarchive_tests.bzl` to test this new functionality?

Will do

maxwellE:
> Can you add anything to `test/starlark_tests/xcarchive_tests.bzl` to test this new functionality?

How can I force a target to build linkmaps outside of passing a flag? I think that is only controlled via a flag right now, I could be mistaken

luispadron:
> How can I force a target to build linkmaps outside of passing a flag? I think that is only controlled via a flag right now, I could be mistaken

You could add a transition to force that but im not sure how that works with the tests to be fair.

maxwellE:
> > How can I force a target to build linkmaps outside of passing a flag? I think that is only controlled via a flag right now, I could be mistaken
> 
> You could add a transition to force that but im not sure how that works with the tests to be fair.

Do you have an example on hand of this? It appears the dsyms are also not tested

luispadron:
Heres how I did it for the `docc_archive` rule and the `swift.emit_symbol_graph` flag: https://github.com/bazelbuild/rules_apple/blob/ba7c7abefbfa3e868a3f7bfa93a98613e2dd70c5/apple/internal/docc.bzl#L155-L168

We use a transition to always set that feature and then an aspect which collects some providers that the rule can then use: https://github.com/bazelbuild/rules_apple/blob/ba7c7abefbfa3e868a3f7bfa93a98613e2dd70c5/apple/internal/aspects/docc_archive_aspect.bzl

I don't think this is required though if we want to keep it as requiring a flag for now, but it might be nice to transition on these always if thats what Xcode does.

luispadron:
I should've probably added the dsym transition when i originally implemented this, today we just set that flag when building for release which works out.

maxwellE:
@luispadron So that looks good but the flag is not a feature in rules apple or rules swift, its in [bazel core ](https://bazel.build/reference/command-line-reference#flag--objc_generate_linkmap) . How do I set that?

mattrobmattrob:
My guess is that you'd need to update `apple_verification_test` to include an argument that allows you to override `--objc_generate_linkmap`. Here's an example that's similar but for `--apple_generate_dsym`.

The rule attribute for `apple_verification_test`:

https://github.com/bazelbuild/rules_apple/blob/ba7c7abefbfa3e868a3f7bfa93a98613e2dd70c5/test/starlark_tests/rules/apple_verification_test.bzl#L271-L276

Which is then used in a transition in the test implementation to control the `--apple_generate_dsym` value.

https://github.com/bazelbuild/rules_apple/blob/ba7c7abefbfa3e868a3f7bfa93a98613e2dd70c5/test/starlark_tests/rules/apple_verification_test.bzl#L67

And then an example how it's used:

https://github.com/bazelbuild/rules_apple/blob/ba7c7abefbfa3e868a3f7bfa93a98613e2dd70c5/test/starlark_tests/ios_framework_tests.bzl#L440-L444

luispadron:
Nice @mattrobmattrob, must've missed that attr when I added this test originally since it's missing a test for dsym. 

@maxwellE if you just want to add another attr to exercise the feature in the test that's fine and I'll follow up to update the rule so we transition these flags automatically (if that's what Xcode does by default)

maxwellE:
@mattrobmattrob Thank you for that, was great to see how to enable that in the future.

Added a test case to cover the logic

tools/xcarchivetool/make_xcarchive.py
```diff
@@ -135,6 +145,14 @@ def _main(
       os.path.join(dsyms_path, os.path.basename(dsym)),
       dirs_exist_ok=True)
 
+  # Copy the linkmap files into the .xcarchive.
+  linkmaps_path = os.path.join(output_path, "Linkmaps")
+  os.makedirs(linkmaps_path, exist_ok=True)
```
mattrobmattrob, line 150: Can you avoid this by following the dSYM `dirs_exist_ok` param usage in the copy call?

maxwellE: Will do

tinder-maxwellelliott: Linkmaps are single files though, copyfile does not take that parameter

maxwellE: Linkmaps are single files though, that argument does not work for copyfile

mattrobmattrob: 🤦 Makes sense. Can you avoid creating the directory unless linkmap files are present?

PR #2289, opened 2023-10-13
Refactor how `ipa_post_processor` is defined for rules

BalestraPatrick:
PiperOrigin-RevId: 563495066
(cherry picked from commit 510919face3d077ce4bdc3fe6744522782ccea7d)


apple/internal/rule_attrs.bzl
```diff
@@ -571,20 +571,6 @@ of the target will be used instead.
...
-            allow_files = True,
-            executable = True,
-            cfg = "exec",
-            doc = """
-A tool that edits this target's archive after it is assembled but before it is signed. The tool is
-invoked with a single command-line argument that denotes the path to a directory containing the
-unzipped contents of the archive; this target's bundle will be the directory's only contents.
-
-Any changes made by the tool must be made in this directory, and the tool's execution must be
-hermetic given these inputs to ensure that the result can be safely cached.
```
brentleyjones, line 587: Doesn't this change actually remove the attribute for rules that could set it? I think "for rules that aren't using it" means inside of Google, but we probably want to add `rule_attrs.ipa_post_processor_attrs(),` back to all the rules that previously had it?

BalestraPatrick: I see that the attr was kept specifically in only a few rules now (main bundling rules such as `ios_application`, `ios_extension`, `macos_command_line_application`, etc.). I do wonder how many people use this feature in rules like `ios_static_framework` today for example and if it's worth keeping, but I think you're right, the reason behind the change was likely a clean up at Google. If we want to be consistent and keep it everywhere it's supported today, then yes, we shouldn't take this change.

brentleyjones: > then yes, we shouldn't take this change

Well, I think we should take the change (because of the `def _ipa_post_processor_attrs()` refactoring, which really should have been it's own commit IMO), but add `rule_attrs.ipa_post_processor_attrs(),` to all the other bundle rules.

apple/internal/macos_rules.bzl
```diff
@@ -1965,7 +1965,6 @@ def _macos_command_line_application_impl(ctx):
         codesign_inputs = ctx.files.codesign_inputs,
         codesignopts = codesigning_support.codesignopts_from_rule_ctx(ctx),
         features = features,
-        ipa_post_processor = None,
```
BalestraPatrick, line 1968: This is now removed since `None` is the default value

brentleyjones (approved):

PR #2292, opened 2023-10-17
Add missing `bzl_library` dependencies

brentleyjones:


brentleyjones:
Depends on https://github.com/bazelbuild/apple_support/pull/275.

cgrindel:
@brentleyjones Do you want me to put up a PR like [this one](https://github.com/bazelbuild/rules_swift/commit/07cc8520924e1d6d91cb6208632a44614708919c) to catch these missing `bzl_library` deps?

luispadron (approved):

cgrindel:
Also, should we add the bzl deps test to `apple_support` as well?

brentleyjones:
I think the latest stardoc, which has correct runfiles support, notices these just fine now? It's how I found these errors for example. Unless I'm misunderstanding and you believe there are more missing deps?

cgrindel:
The `bzl_test` that I added to `rules_swift` and some other repos ensures that the `bzl_library` targets have the correct deps going forward. Many repos do not generate their docs with every PR. If that is the case, then `bzl_test` ensures that `bzl_library` deps stay up-to-date.

PR #2298, opened 2023-10-17
Enable Bzlmod on last_green CI

brentleyjones:


luispadron (approved):

brentleyjones:
Failures are because we need to use `str(Label())` for some more analysis tests. I'll get those merged in first. I also have shell tests working with bzlmod (after some more PRs), so I'll probably change this PR to allow bzlmod for all tests at that point.

PR #2302, opened 2023-10-17
Allow extended regex in `expect_log`

brentleyjones:
Will be needed in order to support matching on bzlmod or non-bzlmod based label strings.


brentleyjones:
I'll adjust the couple of call sites that break with the new flag soon.

brentleyjones:
Actually, I'll see if I can just craft a different regex in my other change.

brentleyjones:
https://github.com/bazelbuild/rules_apple/pull/2303

PR #2309, opened 2023-10-23
Support extensions within single-target watchOS apps

markvasiv:
### What has changed
- Adds support for app extensions within single target watchOS apps.
- Extension-based watchOS apps can now define their `watchos_extension` either in `extension` or in `extensions`, the former being deprecated.
- ~Removes `extension` all-together from `watchos_application`.~

### Notes
- ~This is a breaking change for projects that use extension-based `watchos_application`, they would need to migrate from `extension = ":WatchKitExtension"` to `extensions = [":WatchKitExtension"]`. Initially I tried to keep both `extension` and `extensions` params to avoid that, but it breaks other things. For instance, [rules_xcodeproj here](https://github.com/MobileNativeFoundation/rules_xcodeproj/blob/8d900a9bbfdcceabe779a9fc269d7226bb9121e7/xcodeproj/internal/top_level_targets.bzl#L241-L244).~
- This is my first PR to this repo, please let me know if I have missed anything, I'd be happy to iterate on this PR if need be!


luispadron:
> From my point of view, I think this makes sense. It's a breaking change and it's not great, but I feel like other options such as adding a new `app_extensions` attribute to specifically support this feature would be confusing given we would need to keep `extension`. Let's hear what other maintainers think!

I agree that would be confusing API, unless we rename them both to be more specific which would probably break more things.

brentleyjones:
> Removes `extension` all-together from watchos_application

Can we instead just error out if they set both? I would prefer to not break existing users. We could mark the old attribute as deprecated (print out a warning if people use it), and remove it in a future major version.

markvasiv:
Sounds good, I can update the PR later this week! We would need to fix `rules_xcodeproj` if we do that though. It breaks [here](https://github.com/MobileNativeFoundation/rules_xcodeproj/blob/8d900a9bbfdcceabe779a9fc269d7226bb9121e7/xcodeproj/internal/top_level_targets.bzl#L241-L244) with "attempting to change a frozen list". Do you think that's all right?

markvasiv:
Updated the PR with the suggested change. Let me know what you think. Here's a [change for rules_xcodeproj](https://github.com/MobileNativeFoundation/rules_xcodeproj/pull/2698).

brentleyjones:
I'll leave it to @BalestraPatrick (or others who actually use the rule) to approve the latest changes.

apple/internal/watchos_rules.bzl
```diff
@@ -703,6 +703,25 @@ later will be met with a rejection.
 
 Please remove the assigned watchOS 2 app `extension` and make sure a valid watchOS application
 delegate is referenced in the single-target `watchos_application`'s `deps`.
+""")
+
+    if len(ctx.attr.extensions) > 1:
+        fail("""
+Extension-based watchOS applications do not support embedded app extensions.
```
BalestraPatrick, line 710:
```suggestion
Extension-based watchOS applications do not support embedded app extensions via the `extensions` attribute.
```

apple/internal/watchos_rules.bzl
```diff
@@ -703,6 +703,25 @@ later will be met with a rejection.
...
+
+Please remove app extensions from this target, and instead add them to the watchos_extension `extensions`.
```
BalestraPatrick, line 712:
```suggestion
Please remove app extensions from this target, and instead add them via `watchos_extension`'s `extensions` attribute.
```

apple/internal/watchos_rules.bzl
```diff
@@ -703,6 +703,25 @@ later will be met with a rejection.
...
+    watch_extension = None
+    if ctx.attr.extension:
+        watch_extension = ctx.attr.extension
+    elif len(ctx.attr.extensions) == 1 and ctx.attr.extensions[0][AppleBundleInfo].product_type == apple_product_type.watch2_extension:
+        watch_extension = ctx.attr.extensions[0]
+    else:
+        fail("""
+Extension-based watchOS applications require a valid watchos_extension.
+
+Please add a watchos_extension to this target `extensions`.
```
BalestraPatrick, line 724:
```suggestion
Please add a `watchos_extension` to this target `extension` attribute.
```

BalestraPatrick: Notice that I'm suggesting the `extension` attribute (instead of the new plural version) since that has always been used to reference the watchOS extension (instead of the app extension), right?

BalestraPatrick: Oh I see we're marking that attribute as deprecated, so ignore the above comment and suggestion. Let's just add the formatting for the rule name.

BalestraPatrick:
We tested submitting a version of our app to App Store Connect and it at least passed validation, so we'll merge this.

BalestraPatrick (approved):

PR #2311, opened 2023-10-26
Avoid failure when shutting down a simulator already in the shutdown state

ileitch:
When performing parallel testing in Xcode 15, some or all simulators may automatically shutdown at the end of the test run. Calling `simctl shutdown` on a simulator already in the shutdown state causes a non-zero exist status, causing the test run to fail even if the tests themselves ran successfully. Instead of calling `shutdown` we can instead rely upon `delete` to shutdown the simulator regardless of its current state.


thii (approved):

mattrobmattrob:
Confirmed `bazel test //test:ios_xctestrun_runner_unit_test` passed locally for me.

PR #2314, opened 2023-10-31
[Docs] Refresh iOS tutorial

BalestraPatrick:
[Recently](https://github.com/bazelbuild/examples/pull/331), the `ios-app` project living in [bazelbuild/examples](https://github.com/bazelbuild/examples) was removed. This PR contains a few improvements:
- I updated our tutorial that shows how to create create a simple iOS app that builds with Bazel not to rely or point to the non-existing project.
- We now provide all the source code (one Swift file and one `Info.plist`) to generate a basic sample iOS app in this tutorial.
- The source code is now Swift instead of Objective-C.
- Simplified many sentences and commands, along with a few bug fixes required due to breaking changes in `rules_xcodeproj` (it now requires a `@bazel_features` explicit dependency).

I think it would be a good idea to move the tutorial to use bzlmod once Bazel 7 ships. For now, we can continue using the `WORKSPACE` for ease of compatibility with the default Bazel 6 behavior.

### DIFF
--- a/MAINTENANCE.md
+++ b/MAINTENANCE.md
@@ -112,6 +112,7 @@ is the recommended process:
 - Update the `apple/repositories.bzl` and `MODULE.bazel` files in `rules_apple` with the new
   `apple_support` and `rules_swift` releases.
 - Repeat the steps above to create a release on `rules_apple`.
+- Update the `url` and `sha256` values in the [ios-app tutorial](https://github.com/bazelbuild/rules_apple/blob/master/doc/tutorials/ios-app.md#update-the-workspace-file) to match the latest release.
 
 ### Notes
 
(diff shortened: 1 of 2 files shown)


brentleyjones (commented):
Meta comment, @mattrobmattrob has a wonderful template for iOS apps: https://github.com/mattrobmattrob/bazel-ios-swiftui-template. I wonder if we can leverage it in any way (maybe just mentioning it)?

luispadron (approved):

BalestraPatrick:
@brentleyjones @mattrobmattrob I wasn't aware of that! Awesome resource, added a note about it.

brentleyjones (approved):

PR #2118, opened 2023-08-18
Include `private_deps` when gathering resources for Apple rules.

keith:
PiperOrigin-RevId: 501409110
(cherry picked from commit 98be7d75a69113ed276214ba3250bd36d784531c)

PR #2317, opened 2023-11-02
Fix/support implementation deps in objc library

xinzhengzhang:
Bazel 6.3 supported  implementation_deps in objc_library https://github.com/bazelbuild/bazel/issues/18298 and it seems that we left out https://github.com/bazelbuild/rules_swift/issues/1081#issuecomment-1636811401 in fix missing resource collect in private_deps https://github.com/bazelbuild/rules_apple/pull/2118

I just added implementation_deps in resource collect and test for these two cases


brentleyjones (approved):

brentleyjones:
Thanks @xinzhengzhang!

PR #2320, opened 2023-11-03
Include dSYMs for any bundled extension targets in the app's xcarchive

jszumski:
Includes dSYMs for transitive targets in a generated xcarchive.  

I added tests to ensure dSYMs are included when `apple_generate_dsym = True` for apps with and without extensions.

### DIFF
--- a/apple/internal/xcarchive.bzl
+++ b/apple/internal/xcarchive.bzl
@@ -33,18 +33,24 @@ def _xcarchive_impl(ctx):
     arguments.add("--bundle", bundle_info.archive.path)
     arguments.add("--output", xcarchive.path)
 
-    for dsym in dsym_info.direct_dsyms:
-        arguments.add("--dsym", dsym.path)
+    arguments.add_all(
+        dsym_info.transitive_dsyms,
+        before_each = "--dsym",
+        expand_directories = False,
+    )
 
     linkmaps = debug_info.linkmaps.to_list()
     for linkmap in linkmaps:
         arguments.add("--linkmap", linkmap.path)
 
     ctx.actions.run(
-        inputs = depset([
-            bundle_info.archive,
-            bundle_info.infoplist,
-        ] + dsym_info.direct_dsyms + linkmaps),
+        inputs = depset(
+            [
+                bundle_info.archive,
+                bundle_info.infoplist,
+            ] + linkmaps,
+            transitive = [dsym_info.transitive_dsyms],
+        ),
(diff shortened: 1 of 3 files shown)


apple/internal/xcarchive.bzl
```diff
@@ -33,7 +33,8 @@ def _xcarchive_impl(ctx):
     arguments.add("--bundle", bundle_info.archive.path)
     arguments.add("--output", xcarchive.path)
 
-    for dsym in dsym_info.direct_dsyms:
+    dsyms = dsym_info.transitive_dsyms.to_list()
```
brentleyjones, line 36: Instead of popping the `depset`, we should use `arguments.add_all()` with `before_each = "--dsym"`.

apple/internal/xcarchive.bzl
```diff
@@ -44,7 +45,7 @@ def _xcarchive_impl(ctx):
         inputs = depset([
             bundle_info.archive,
             bundle_info.infoplist,
-        ] + dsym_info.direct_dsyms + linkmaps),
+        ] + dsyms + linkmaps),
```
brentleyjones, line 48: And here we should use `transitive = [dsym_info.transitive_dsyms]`.

apple/internal/xcarchive.bzl
brentleyjones: How about `expand_directories = False`?

jszumski: TIL, thanks

luispadron (approved):

luispadron:
@brentleyjones can you take another look when you get a chance

apple/internal/xcarchive.bzl
```diff
@@ -33,18 +36,25 @@ def _xcarchive_impl(ctx):
     arguments.add("--bundle", bundle_info.archive.path)
     arguments.add("--output", xcarchive.path)
 
-    for dsym in dsym_info.direct_dsyms:
-        arguments.add("--dsym", dsym.path)
+    arguments.add_all(
+        dsym_info.transitive_dsyms,
+        map_each = _to_path,
```
brentleyjones, line 41: Is this needed? Just a little surprised it is.

brentleyjones (approved):

apple/internal/xcarchive.bzl
jszumski: Turns out it's not, good catch

PR #2324, opened 2023-11-14
Update README.md for platform_mappings requirement

luispadron:


jszumski (approved):

README.md
```diff
@@ -23,9 +23,9 @@ repository.
 
 1. Copy the latest `MODULE.bazel` or `WORKSPACE` snippet from [the releases
 page](https://github.com/bazelbuild/rules_apple/releases).
-2. Copy the
+2. (Optional: only for Bazel 6 & rules_apple 2.x) Copy the
```
brentleyjones, line 26: It was actually only needed for 3.0.0-rc2, we addressed it by 3.0.0, so I think we can remove this bullet entirely?

PR #2328, opened 2023-11-15
Add swiftinterface_imports to apple_static_xcframework_import

keith:
Without this static xcframeworks which have a swiftinterface file will
fail to compile when sandboxing is enabled since their swiftinterface
won't be copied to the xcframework bundle.

This changed in https://github.com/bazelbuild/rules_apple/pull/2266
which was mirroring the swiftmodule_imports behavior that I believe is
also wrong so I fixed that here too.


thii (approved):

brentleyjones:
We will need tests to prevent future regressions.

PR #2332, opened 2023-11-22
Downgrade last_green to avoid failure

BalestraPatrick:
To be reverted once https://github.com/bazelbuild/stardoc/issues/195 is fixed.

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -15,7 +15,9 @@ x_defaults:
     test_flags:
     - --test_tag_filters=-skipci
   common_last_green: &common_last_green
-    bazel: last_green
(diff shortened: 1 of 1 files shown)


.bazelci/presubmit.yml
```diff
@@ -15,7 +15,7 @@ x_defaults:
     test_flags:
     - --test_tag_filters=-skipci
   common_last_green: &common_last_green
-    bazel: last_green
```
brentleyjones, line 18: Can you add a TODO above it?

brentleyjones (approved):

.bazelci/presubmit.yml
BalestraPatrick: Done!

PR #2333, opened 2023-11-27
Update rules_xcodeproj version on ios-app.md

herlianzhang:
Just try to learn Bazel and after following the documentation I have some issue on xcodeproj section, and I realize that cause of version 1.3.3 of rules_xcodeproj don't have `bazel_features` there.

Here is the error
Screenshot 2023-11-27 at 16 37 43


thii (approved):
Thanks @herlianzhang!

PR #2343, opened 2023-12-20
Add example googletest on iOS hack

keith:
This uses a custom test runner to exec a googletest based binary (or any
binary where the test is defined by exiting non-zero) on iOS.


keith:
just wrote as an example, if anyone is particularly motivated to support this feel free to comment here

PR #2350, opened 2023-12-30
Fix minor issues with module_map rule

luispadron:
Fixes some issues I noticed while I was copying `module_map` to our internal project.

- Sort `load`
- `break` in loop after finding swift generated header
- Remove duplicate `umbrella_header` being added to output
- Use `:` prefix for umbrella header label

### DIFF
--- a/apple/internal/experimental_mixed_language_library.bzl
+++ b/apple/internal/experimental_mixed_language_library.bzl
@@ -1,11 +1,11 @@
 """experimental_mixed_language_library macro implementation."""
 
+load("@bazel_skylib//lib:paths.bzl", "paths")
 load(
     "@build_bazel_rules_swift//swift:swift.bzl",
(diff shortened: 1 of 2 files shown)


mattrobmattrob (approved):

thii (approved):

apple/internal/experimental_mixed_language_library.bzl
```diff
@@ -341,7 +341,7 @@ target only contains Objective-C files.""")
             # attribute to declare custom inputs.
             ":" + umbrella_module_map,
         ],
-        module_map = umbrella_module_map,
+        module_map = ":" + umbrella_module_map,
```
brentleyjones, line 344: Since we do this above as well, can it get pulled out so we only do the string creation once?

brentleyjones (approved):

PR #2351, opened 2023-12-31
Add support for clang modules in experimental_mixed_language_library

luispadron:
This adds support for `@import` when using `experimental_mixed_language_library`. There seems to be a "bug" (or at least a usability issue) in Bazel where if `module_map` is set on `objc_library` then the `enable_modules` attr does nothing.

This enables modules via copts instead, similar to how it was done in [rules_apple_line](https://github.com/line/rules_apple_line/blob/63f50b29616e7a439c47a140cadd59c3ccf4066f/apple/mixed_static_framework.bzl#L279).

### DIFF
--- a/apple/internal/experimental_mixed_language_library.bzl
+++ b/apple/internal/experimental_mixed_language_library.bzl
@@ -195,6 +195,7 @@ def experimental_mixed_language_library(
         name,
         srcs,
         deps = [],
+        enable_modules = False,
         module_name = None,
         objc_copts = [],
         swift_copts = [],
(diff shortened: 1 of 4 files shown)


apple/internal/experimental_mixed_language_library.bzl
```diff
@@ -275,6 +278,11 @@ target only contains Objective-C files.""")
         "-import-underlying-module",
     ]
 
+    # Modules is not enabled if a custom module map is present even with
```
mattrobmattrob, line 281: I noticed the same thing when I was looking into this a while ago. Can you file an issue in Bazel core just to track that this oddity exists?

mattrobmattrob (approved):

apple/internal/experimental_mixed_language_library.bzl
luispadron: Yep will do!

luispadron: It seems to be explicitly checked against: https://github.com/bazelbuild/bazel/blob/18d01e7f6d8a3f5b4b4487e9d61a6d4d0f74f33a/src/main/java/com/google/devtools/build/lib/rules/objc/CompilationSupport.java#L1280 but its odd because without `-fmodules` you cant `@import` even if you have a module map

mattrobmattrob: I had figured it intentional given the user is making the call to use a custom module map. But 🤷‍♂️.

luispadron: https://github.com/bazelbuild/bazel/issues/20703

thii (approved):

apple/internal/experimental_mixed_language_library.bzl
mattrobmattrob: Mind adding the new issue in the code comment before merging, please?

PR #2352, opened 2024-01-01
Add `header_map` rule

### LINKED ISSUE #1757: Add headermap include documentation
Currently when migrating to bazel you often hit some issues around include paths that affect include performance. One solution to this is headermaps, but that potentially comes with the downside of a headermap per target, which adds many actions to the build. It seems that the best long term solution to this is to add `-I $(SRCROOT)` as your only include path, but this requires you to change all your includes because this isn't common in the iOS community. The community would benefit from documentation around this, and potentially a headermap rule to ease the migration. One question about a specific headermap rule is what forms of imports do you need to expose

luispadron:
This adds a new `header_map` rule for creating binary `.hmap` files which are required to support `#import <Foo/Foo.h>` semantics without restructuring an entire project.

Header maps are generated by default when using Xcode, so projects built in Xcode tend to end up relying on them. When migrating an existing project to use Bazel this lack of header map support has been a source of confusion in several threads on the Bazel Slack. Users either end up using `rules_ios`, which this implementation comes from, or they end up copying the header map rule locally. 

By adding it to `rules_apple` we can help folks who are migrating existing projects. Along with documentation to make this less confusing.

My goal here is to start a discussion around the addition of the rule and it's implementation and eventually extend `experimental_mixed_language_library` to support generating these header maps (with this rule) when requested.

Closes #1757 

## Open questions

~~- Do we want to use the C implementation from rules_ios? [There is a Swift implementation](https://github.com/milend/hmap/blob/master/Sources/HeaderMapCore/BinaryHeaderMap.swift) we could look at taking from~~
~~- If we do want to use the C implementation, should we spend some time getting the dependencies (lines/uthash) as Bazel modules (or at least consumed via download) vs. copying the sources into the source tree?~~

Decided on using Swift implementation

### DIFF
--- a/apple/header_map.bzl
+++ b/apple/header_map.bzl
@@ -1,0 +1,24 @@
+# Copyright 2024 The Bazel Authors. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+Rules for creating header maps.
+"""
+
+load(
+    "@build_bazel_rules_apple//apple/internal:header_map.bzl",
+    _header_map = "header_map",
+)
+
+header_map = _header_map
--- a/apple/internal/header_map.bzl
+++ b/apple/internal/header_map.bzl
@@ -1,0 +1,171 @@
+# Copyright 2024 The Bazel Authors. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Rule for creating header_maps."""
+
+load("@build_bazel_rules_swift//swift:swift.bzl", "swift_common")
+
+HeaderMapInfo = provider(
+    doc = "Provides information about created `.hmap` (header map) files",
+    fields = {
+        "public_header_maps": "depset of paths of any public header maps",
+    },
+)
+
+def _write_header_map(actions, header_map_tool, output, module_name, hdrs_lists):
+    """Makes a binary hmap file by executing the header_map tool.
+
+    Args:
+        actions: a ctx.actions struct
+        header_map_tool: an executable pointing to @bazel_build_rules_ios//rules/hmap:hmaptool
+        output: the output file that will contain the built hmap
+        module_name: the prefix to be used for header imports
+        hdrs_lists: an array of enumerables containing headers to be added to the hmap
+    """
+
+    args = actions.args()
+    if module_name:
+        args.add("--module_name", module_name)
+
+    args.add("--output", output)
+
+    for hdrs in hdrs_lists:
+        args.add_all(hdrs)
+
+    args.set_param_file_format(format = "multiline")
+    args.use_param_file("@%s")
+
+    actions.run(
+        mnemonic = "HmapWrite",
+        arguments = [args],
+        executable = header_map_tool,
+        outputs = [output],
+    )
+
+def _header_map_impl(ctx):
+    """Implementation of the header_map() rule.
+
+    It creates a text file with mappings and creates an action that calls out to the header_map tool
+    to convert it to a binary header_map file.
+    """
+    hdrs_lists = [ctx.files.hdrs] if ctx.files.hdrs else []
+
+    for dep in ctx.attr.deps:
+        found_headers = []
+        if apple_common.Objc in dep:
+            found_headers.append(getattr(dep[apple_common.Objc], "direct_headers", []))
+        if CcInfo in dep:
+            found_headers.append(dep[CcInfo].compilation_context.direct_headers)
+        if not found_headers:
+            fail("Direct header provider: '%s' listed in 'deps' does not have any direct headers to provide." % dep)
+        hdrs_lists.extend(found_headers)
+
+    hdrs_lists = [[h for h in hdrs if h.basename.endswith(".h")] for hdrs in hdrs_lists]
+
+    _write_header_map(
+        actions = ctx.actions,
+        header_map_tool = ctx.executable._hmaptool,
+        output = ctx.outputs.header_map,
+        module_name = ctx.attr.module_name,
+        hdrs_lists = hdrs_lists,
+    )
+
+    return [
+        apple_common.new_objc_provider(),
+        swift_common.create_swift_info(),
+        CcInfo(
+            compilation_context = cc_common.create_compilation_context(
+                headers = depset([ctx.outputs.header_map]),
+            ),
+        ),
+        HeaderMapInfo(
+            public_header_maps = depset([ctx.outputs.header_map]),
+        ),
+    ]
+
+header_map = rule(
+    implementation = _header_map_impl,
+    output_to_genfiles = True,
+    attrs = {
+        "module_name": attr.string(
+            mandatory = False,
+            doc = "The prefix to be used for header imports",
+        ),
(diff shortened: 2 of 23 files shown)


examples/multi_platform/MixedLib/MixedAnswer.m
```diff
@@ -1,4 +1,5 @@
-#import "examples/multi_platform/MixedLib/MixedAnswer.h"
+#import <MixedAnswer/MixedAnswer.h>
+
 #import "examples/multi_platform/MixedLib/MixedAnswer-Swift.h"
```
luispadron, line 3: Once `experimental_mixed_language_library` is updated to support `header_map` the `#import <MixedAnswer/MixedAnswer-Swift.h` will be possible.

luispadron:
I added a Swift implementation to see the difference between the too. Either works for me but Swift requires a bit less LOC to maintain.

luispadron:
@keith @brentleyjones for when you two have a chance to review. I can add tests and such when we align on the hmap implementation

tools/hmaptool/BUILD
```diff
@@ -0,0 +1,61 @@
...
+swift_library(
+    name = "BinaryHeaderMapTool",
+    srcs = [
+        "BinaryHeaderMap.swift",
+        "BinaryHeaderMapEncoder.swift",
+        "BinaryHeaderMapTool.swift",
+    ],
+)
+
+macos_command_line_application(
```
thii, line 52: Maybe switch to `swift_binary` wrapped by an `apple_universal_binary` instead? I think the minimum version setting should be set by downstream users.

thii:
> Do we want to use the C implementation from rules_ios? [There is a Swift implementation](https://github.com/milend/hmap/blob/master/Sources/HeaderMapCore/BinaryHeaderMap.swift) we could look at taking from

No strong opinion on this—I think this would rarely change, so either is fine. It would be great if you could do a perf test on a large project and/or add a Starlark flag to allow switching.

> If we do want to use the C implementation, should we spend some time getting the dependencies (lines/uthash) as Bazel modules (or at least consumed via download) vs. copying the sources into the source tree?

Yes, I think we should (as a non-module dependency is fine). (Same if we use the Swift implementation)

apple/header_map.bzl
```diff
@@ -0,0 +1,24 @@
...
+"""
+Rules for creating header maps.
+"""
+
+load(
+    "@build_bazel_rules_apple//apple/internal:header_map.bzl",
+    _header_map = "header_map",
+)
+
+header_map = _header_map
```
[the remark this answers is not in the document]
luispadron: I was thinking we'd want this to be public so it could be used by other external macros, etc. We can definitely automatically support it with `experiemental_mixed_language_library` but it might still be wanted by folks using a pure `objc_library` for example.

tools/hmaptool/BUILD
luispadron: Makes sense, I'll update!

luispadron:
Thanks for the review @thii, some thoughts:

> No strong opinion on this—I think this would rarely change, so either is fine. It would be great if you could do a perf test on a large project and/or add a Starlark flag to allow switching.

Lets go with Swift then, it's less code and a bit easier to keep maintained IMO.

> If we do want to use the C implementation, should we spend some time getting the dependencies (lines/uthash) as Bazel modules (or at least consumed via download) vs. copying the sources into the source tree?
> 
> Yes, I think we should (as a non-module dependency is fine). (Same if we use the Swift implementation)

I actually changed the Swift one up a bit and had to make my own CLI around the headermap code so we'd just keep the two files in here for that one.

td-mercari (approved):
LGTM. But please add docs and tests.

apple/header_map.bzl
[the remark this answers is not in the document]
aaronsky: As someone who has recently adapted (copied) this work into their own project to create a functioning mixed-language library, I would very much appreciate it being made part of the public API. Though, that was under the current constraint that headermaps don't propagate correctly without manual configuration, which I recall there has been discussion towards addressing.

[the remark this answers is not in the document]
thii: 👍

apple/internal/header_map.bzl
```diff
@@ -0,0 +1,171 @@
...
+        HeaderMapInfo(
+            public_header_maps = depset([ctx.outputs.header_map]),
+        ),
+    ]
+
+header_map = rule(
+    implementation = _header_map_impl,
+    output_to_genfiles = True,
+    attrs = {
+        "namespace": attr.string(
```
luispadron, line 101: Does `module_name` make more sense here?

thii: I think so.

tools/hmaptool/README.md
```diff
@@ -0,0 +1,4 @@
+# hmaptool
+
+This is a tool which creates binary headermaps to enable angle bracket imports of headers when building with Bazel.
+Required to mimic Xcode's behavior of generating headermaps by default which iOS projects rely on.
```
mattrobmattrob, line 4: This is a general tool for `rules_apple` and not iOS specific, correct?
```suggestion
Required to mimic Xcode's behavior of generating headermaps by default which projects rely on.
```

examples/multi_platform/MixedLib/BUILD
```diff
@@ -3,16 +3,41 @@ load(
...
     ],
-    hdrs = ["MixedAnswer.h"],
+    hdrs = [
+        ":MixedAnswer.public_headers.h",
+    ],
     enable_modules = True,
+    includes = [
+        "MixedAnswerHeaderMap.hmap",
+    ],
+    objc_copts = ["-I."],
```
mattrobmattrob, line 29: Can you add a comment about why this is added here or perhaps reference an issue for `experimental_mixed_language_library` to internally manage this when a header map is created?

examples/multi_platform/MixedLib/BUILD
```diff
@@ -23,6 +48,19 @@ swift_library(
...
 
+objc_library(
+    name = "ObjcLibDependingOnMixedLib",
+    srcs = [
+        "ObjcLibDependingOnMixedLib.m",
+    ],
+    hdrs = [
+        "ObjcLibDependingOnMixedLib.h",
+    ],
+    copts = ["-I."],
```
mattrobmattrob, line 59: Same here WRT an explanation of this existing in comment form.

mattrobmattrob (approved):
It's my preference that we have an explicit mixed language example with `header_map` used and one without, if that's not too difficult. I think not muddling them together might make the path for folks more clear.

I didn't review the `tools/hmaptool` very closely since I'm not that familiar with the format/this functionality. 🙏

examples/multi_platform/MixedLib/BUILD
luispadron: Yep will do, i made: https://github.com/bazelbuild/rules_apple/issues/2371

examples/multi_platform/MixedLib/BUILD
luispadron: Added

tools/hmaptool/README.md
luispadron: Made the change thanks!

luispadron:
> It's my preference that we have an explicit mixed language example with header_map used and one without, if that's not too difficult. I think not muddling them together might make the path for folks more clear.

Do you mean you want me to add a new example specifically for header_map? If so, I can do that!

mattrobmattrob:
> > It's my preference that we have an explicit mixed language example with header_map used and one without, if that's not too difficult. I think not muddling them together might make the path for folks more clear.
> 
> Do you mean you want me to add a new example specifically for header_map? If so, I can do that!

That's my preference. Something like (1) an example without header map and (2) an example with header map. In the future, it'll be easy to use it to demonstrate the two different pathways for the mixed language macro. 🙏

luispadron:
@mattrobmattrob Makes sense, I added a new `examples/multi_platform/MixedLibWithHeaderMap` example and reverted the changes to the existing example.

examples/multi_platform/MixedLibWithHeaderMap/MixedAnswer.m
```diff
@@ -0,0 +1,11 @@
+#import <MixedAnswer/MixedAnswer.h>
+
+#import "examples/multi_platform/MixedLibWithHeaderMap/MixedAnswer-Swift.h"
```
mattrobmattrob, line 3: Want to use header maps in the entire example?

luispadron: We cant yet until we have #2371, it requires knowing the swift_library internal target and passing that to `deps` of `header_map`. It needs to be done by `experimental_mixed_language_library`

mattrobmattrob: Wonderful, sounds good to me then!

mattrobmattrob (approved):

PR #2355, opened 2024-01-08
Test remove objc from new_executable_binary_provider

keith:
This was removed from bazel head and I'm hoping to see how it is used
with older versions


keith:
https://github.com/bazelbuild/rules_apple/pull/2356/

Issue #2353, opened 2024-01-08
Error in new_objc_provider: Key 'linkopt' no longer supported in ObjcProvider



sgowroji:
CI: https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/3575#018ce73b-14e1-4fb3-80d1-dd333ede35b2
Platform: MacOS

 Logs: 
```
(05:01:58) ERROR: /private/var/tmp/_bazel_buildkite/79edcf2dae0ace64ac8fec4fc8c87a0c/external/rules_swift~1.13.0~non_module_deps~build_bazel_rules_swift_local_config/BUILD:9:22: in xcode_swift_toolchain rule @@rules_swift~1.13.0~non_module_deps~build_bazel_rules_swift_local_config//:toolchain:
Traceback (most recent call last):
	File "/private/var/tmp/_bazel_buildkite/79edcf2dae0ace64ac8fec4fc8c87a0c/external/rules_swift~1.13.0/swift/internal/xcode_swift_toolchain.bzl", line 580, column 57, in _xcode_swift_toolchain_impl
		swift_linkopts_providers = _swift_linkopts_providers(
	File "/private/var/tmp/_bazel_buildkite/79edcf2dae0ace64ac8fec4fc8c87a0c/external/rules_swift~1.13.0/swift/internal/xcode_swift_toolchain.bzl", line 256, column 51, in _swift_linkopts_providers
		objc_info = apple_common.new_objc_provider(linkopt = depset(linkopts)),
Error in new_objc_provider: Key 'linkopt' no longer supported in ObjcProvider (use CcInfo instead).
```

 Steps:
```
 git clone https://github.com/bazelbuild/rules_apple.git 
 git reset 5574843206b3adab9fd0ef8ab57064379ff9cf31 --hard
export USE_BAZEL_VERSION=8e8ddaba0e90c280bfd85644d6ccd91df5b6d353
bazel build tools/... test/... 
```

 CC Greenteam @fweikert

sgowroji:
Hi @keith, Latest CI is still shows failure  [https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/3583#018cf658-a8fd-4bc7-8a47-b29dd8d9bd6d](https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/3583#018cf658-a8fd-4bc7-8a47-b29dd8d9bd6d) for rules\_apple.  Can you please take a  look. Thanks!

[...]

PR #2356, opened 2024-01-08
Fix compatibility with --incompatible_objc_provider_remove_linking_info

keith:
Fixes https://github.com/bazelbuild/rules_apple/issues/2353

### DIFF
--- a/apple/internal/framework_import_support.bzl
+++ b/apple/internal/framework_import_support.bzl
@@ -465,24 +465,17 @@ def _objc_provider_with_dependencies(
 
     return apple_common.new_objc_provider(**objc_provider_fields)
 
-def _new_dynamic_framework_provider(*, cc_info, **kwargs):
+def _new_dynamic_framework_provider(**kwargs):
     """A wrapper API for the Bazel API of the same name to better support multiple Bazel versions
 
     Args:
         cc_info: CcInfo provider for the dynamic framework.
         **kwargs: Additional arguments to pass if supported.
     """
+    if not _OBJC_PROVIDER_LINKING:
+        kwargs.pop("objc", None)
 
-    test_provider = apple_common.new_dynamic_framework_provider(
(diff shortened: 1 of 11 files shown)


keith:
depends on https://github.com/bazelbuild/rules_swift/pull/1143

apple/internal/framework_import_support.bzl
```diff
@@ -457,8 +460,30 @@ def _objc_provider_with_dependencies(
...
 
+def _new_dynamic_framework_provider(*, cc_info, **kwargs):
+    """A wrapper API for the Bazel API of the same name to better support multiple Bazel versions
+
+    Args:
+        cc_info: CcInfo provider for the dynamic framework.
+        **kwargs: Additional arguments to pass if supported.
+    """
+
+    test_provider = apple_common.new_dynamic_framework_provider(
```
brentleyjones, line 476: Same as my comment in rules_swift, can we avoid creating a throw-away instance?

keith: done

apple/internal/linking_support.bzl
```diff
@@ -51,6 +54,26 @@ def _debug_outputs_by_architecture(link_outputs):
...
+def _new_executable_binary_provider(*, binary, cc_info, objc):
+    """Wrap the apple_common API of the same name to better support multiple Bazel versions.
+
+    Args:
+        binary: The `File` representing the binary.
+        cc_info: The `CcInfo` provider for the binary.
+        objc: The `apple_common.Objc` provider for the binary.
+    """
+    if _OBJC_PROVIDER_LINKING:
+        return apple_common.new_executable_binary_provider(
```
keith, line 66: it looks to me like these rules are the only consumers of this info, so theoretically we could switch to our own provider for this instead, if we did the only field we currently use is the `binary`

brentleyjones (approved):

PR #2361, opened 2024-01-12
Don't pass the `-iphone` flag to `pngcrush` on macOS

maxbelanger:
We're attempting to migrate to using `macos_application` from Xcode, and noticed that `.png` resources in our macOS app (built via `macos_application`) were corrupt. Specifically, Electron/Chromium (actually `libpng` under the hood) would consistently fail to load images when the app was built using the Bazel rule. 

This appears to be caused by the `-iphone` flag that is passed to `pngcrush`. It's not clear what this does to a `.png`, but it seems to be proprietary somehow. A reasonable workaround for us was to disable this on macOS only, since it's not unlikely to be using something other than `NSImage`/Foundation on this platform. 

I'm not quite sure whether this is the most elegant fix, though, so I'd love to hear suggestions if there's a better way to fix this.


thii:
I looked at how Xcode does it, and looks like it doesn't compress the PNG files if the target platform is macOS, and just copies the files. There are two build settings that control this: `COMPRESS_PNG_FILES` and `STRIP_PNG_TEXT`, and they default to `NO` for macOS.

The logic in copypng:
```
if ( $compress ) {
    @args = ( $PNGCRUSH, "-q", "-iphone", "-f", "0" );
    if ( $stripPNGText ) {
        push ( @args, "-rem", "text" );
    }
    push ( @args, $SRCFILE, $DSTFILE );

} else {
    @args = ( "cp", "$SRCFILE", "$DSTFILE" );
}
```

thii:
So this should be a symlink action if the platform type is macOS.

keith:
fwiw it might have been google was doing this intentionally differently from xcode just for the potential size win

maxbelanger:
@keith happy to make this controllable in some way, maybe a feature to force this to happen on macOS? I'm not sure how to fully replicate the Xcode behavior here though (different defaults on each platform).

maxbelanger:
Alternatively, using a `define` could also work and would allow us to express the default as False on macOS and True on other platforms. The only drawback is that users would have to use the command-line or a `.bazelrc` to override the default. Any preference between the two approaches?

luispadron:
@maxbelanger If you want to run buildifier to format and fix CI we can get this merged

luispadron (approved):
Could you add the new feature flag to: https://github.com/bazelbuild/rules_apple/blob/master/doc/common_info.md

maxbelanger:
> Could you add the new feature flag to: https://github.com/bazelbuild/rules_apple/blob/master/doc/common_info.md

Added a reference to this in `doc/resources.md`, where the functionality to compress PNGs is already discussed.

Issue #2363, opened 2024-01-14
Versioned Mac frameworks with a non-standard install name fail to be imported


maxbelanger:
We're using `apple_dynamic_framework_import` as part of building a macOS application, but certain frameworks don't get processed properly by `imported_dynamic_framework_processor`. The logic used to detect the current version of a versioned macOS framework inspects the `install_name`, which _usually_ points to the right binary under `<framework>/Versions/<version>/<binary>`.

In our case, bundling e.g. Electron would break since its `install_name` is: 
```
otool -D /Applications/Dropbox.app/Contents/Frameworks/Electron\ Framework.framework/Electron\ Framework
/Applications/Dropbox.app/Contents/Frameworks/Electron Framework.framework/Electron Framework:
@rpath/Electron Framework.framework/Electron Framework
```

We've worked around this one by adding an optimization to version detection that lists the contents of `<framework>/Versions/` and figures out the right version if when there's only a single entry (will attach a PR shortly).

I suspect this is otherwise hard to fix given you'd have to inspect symlink and that can be problematic with Bazel, but opening a discussion in case anyone has a better idea.

PR #2364, opened 2024-01-14
Avoid looking at install name of a versioned framework if there's only one version

maxbelanger:
See #2363. Add an optimistic check for when a Mac framework has a single version that lets us avoid inspecting the `install_name`, which can be brittle.


keith (approved):
lgtm

PR #2372, opened 2024-01-19
Add support for headermap to experimental_mixed_language_library

### LINKED ISSUE #2371: Add header map support to `experimental_mixed_language_library`
Once #2352 merges we should update `experimental_mixed_language_library` to use `header_map` and the `header_map_support` rules.

luispadron:
This follows up #2352 by adding header map support to `experimental_mixed_language_library`. It adds a new parameter: `enable_header_map` which if `True` will generate a headermap for the mixed language `objc_library`. This header map will be included by the `swift_library` and `objc_library` to enable `#import <>` semantics.

Closes #2371

### DIFF
--- a/apple/internal/experimental_mixed_language_library.bzl
+++ b/apple/internal/experimental_mixed_language_library.bzl
@@ -6,6 +6,7 @@ load(
     "SwiftInfo",
     "swift_library",
 )
+load("//apple/internal:header_map_support.bzl", "header_map_support")
 
 _CPP_FILE_TYPES = [".cc", ".cpp", ".mm", ".cxx", ".C"]
 
@@ -196,6 +197,7 @@ def experimental_mixed_language_library(
         srcs,
         deps = [],
         enable_modules = False,
+        enable_header_map = False,
         module_name = None,
         objc_copts = [],
         swift_copts = [],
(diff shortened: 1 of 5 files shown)


apple/internal/experimental_mixed_language_library.bzl
```diff
@@ -221,6 +223,7 @@ def experimental_mixed_language_library(
         deps: A list of targets that are dependencies of the target being
             built, which will be linked into that target.
         enable_modules: Enables clang module support for the Objective-C target.
+        enable_header_map: Enables header map support for the Objective-C target.
```
mattrobmattrob, line 226:
```suggestion
        enable_header_map: Enables header map support for the Swift and Objective-C target.
```

apple/internal/experimental_mixed_language_library.bzl
```diff
@@ -288,14 +291,27 @@ target only contains Objective-C files.""")
...
     # Add Obj-C includes to Swift header search paths
     repository_name = native.repository_name()
-    includes = kwargs.get("includes", [])
+    includes = kwargs.pop("includes", [])
     for x in includes:
         include = x if repository_name == "@" else "external/" + repository_name.lstrip("@") + "/" + x
         swift_copts += [
             "-Xcc",
             "-I{}".format(include),
         ]
```
mattrobmattrob, line 300: It's intentional that this doesn't include the `header_map_ctx.includes`, correct?

mattrobmattrob (approved):

apple/internal/experimental_mixed_language_library.bzl
luispadron: Yeah its not needed since it's not needed from what I've seen. The headermap include should be propagated to downstream deps via the `objc_library1`

PR #2373, opened 2024-01-19
Upgrade to Xcode 15.1

BalestraPatrick:


test/apple_shell_testutils.sh
```diff
@@ -523,21 +523,19 @@ function is_device_build() {
...
-  # against.
-  xcrun llvm-objdump --macho --section=__TEXT,__entitlements "$binary" | \
-      sed -e 's/^[0-9a-f][0-9a-f]*[[:space:]][[:space:]]*//' \
-          -e 'tx' -e 'd' -e ':x' | xxd -r -p
+  local output="$2"
+  local archs=($(lipo -archs "$binary"))
+  local thin_binary="tempdir/thin_binary"
+  mkdir -p tempdir
+  lipo -thin ${archs[0]} "$binary" -output "$thin_binary"
+  xcrun segedit "$thin_binary" -extract __TEXT __entitlements "$output"
```
keith, line 537: couldn't we output this to `-` as stdout instead?

BalestraPatrick: Happy to take suggestions though!

keith: I was suggesting to do the lipo, and then just make the segedit part output to stdout so you don't have to make callers know about the temporary files

keith (approved):

test/apple_shell_testutils.sh
BalestraPatrick: I'll merge and investigate

luispadron (approved):

PR #2379, opened 2024-01-22
Update to Xcode 15.2

keith:
This also enables visionOS tests by default


keith:
not sure if CI supports it, we'll see, might need Xcode 15.2

mattrobmattrob (approved):

PR #2385, opened 2024-01-23
Don't set an install path for macOS QuickLook plugins

maxbelanger:
Removes the linker option that sets the `install_name` of QuickLook plugins to `/Library/Frameworks/<bundle_name>.qlgenerator`, and passes `-bundle` rather than `-dynamiclib`. 

QuickLook plugins can be located in multiple locations on disk (e.g. `/Library/QuickLook`, `~/Library/QuickLook`, in an app bundle, etc) and they're usually built as bundles rather than linked to directly (they get loaded manually by `quicklookd`), so an install path isn't necessary. 

Also, as currently implemented this only works if you explicitly set `bundle_name`, and adds a garbage value otherwise (`/Library/Frameworks/.qlgenerator/`).

### DIFF
--- a/apple/internal/macos_rules.bzl
+++ b/apple/internal/macos_rules.bzl
@@ -1018,14 +1018,11 @@ def _macos_quick_look_plugin_impl(ctx):
         validation_mode = ctx.attr.entitlements_validation,
     )
 
-    extra_linkopts = [
-        "-dynamiclib",
-    ]
     link_result = linking_support.register_binary_linking_action(
         ctx,
         entitlements = entitlements.linking,
         exported_symbols_lists = ctx.files.exported_symbols_lists,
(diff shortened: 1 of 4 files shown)


keith (approved):
lgtm, thanks! I wouldn't be surprised if you were one of the few users of this feature so if you spot any other issues feel free to submit more PRs!

maxbelanger:
@keith it occurs to me that the only difference between `macos_quick_look_plugin` and a properly-configured `macos_bundle` (i.e. where you set `bundle_extension` to `qlgenerator`) is that this rule enforces there is no `rpath` (the tests also check this).

We might want to eventually deprecate this rule in favor of some logic such that if `bundle_extension` is set to `qlgenerator` we suppress the `rpath`?

PR #2380, opened 2024-01-22
Add bazel 6 back to CI

keith:

PR #2386, opened 2024-01-23
Fix broken builds with Bazel < 6.4.0

thii:
Which doesn't have `visionos` in the platform type API.

### DIFF
--- a/apple/internal/partials/app_assets_validation.bzl
+++ b/apple/internal/partials/app_assets_validation.bzl
@@ -27,6 +27,8 @@ load(
     "partial",
 )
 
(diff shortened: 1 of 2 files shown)


brentleyjones (approved):

keith:
interestingly bazel 6.x tests pass without this https://github.com/bazelbuild/rules_apple/pull/2380 i guess because it just doesn't hit this path

luispadron (approved):

keith:
ah yea I forgot we cherry picked this. i guess we could test an older 6.x, although also I think we assume that folks can safely and easily update to the newest of a LTS train so it's less of an issue

PR #2393, opened 2024-02-05
Update xcarchive dsyms tests

maxwellE:
This should check to see that the dsym binary does exist

### DIFF
--- a/test/starlark_tests/xcarchive_tests.bzl
+++ b/test/starlark_tests/xcarchive_tests.bzl
@@ -53,6 +53,8 @@ def xcarchive_test_suite(name):
         target_under_test = "//test/starlark_tests/targets_under_test/ios:app_minimal.xcarchive",
         contains = [
             "$BUNDLE_ROOT/dSYMs/app_minimal.app.dSYM",
+            "$BUNDLE_ROOT/dSYMs/app_minimal.app.dSYM/Contents/Resources/DWARF/app_minimal",
+            "$BUNDLE_ROOT/dSYMs/app_minimal.app.dSYM/Contents/Info.plist",
             "$BUNDLE_ROOT/Info.plist",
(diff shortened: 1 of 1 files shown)


luispadron (approved):

tinder-maxwellelliott:
@luispadron Are you seeing these binaries being created?

luispadron:
I believe so yes, I'll double check today. We use it with `--apple_generate_dsym` and output_groups.

maxwellE:
> I believe so yes, I'll double check today. We use it with `--apple_generate_dsym` and output_groups.

I am certain that we are setting these in our invocation. Is there any way you can share a trimmed down invocation for me to see? 

Regardless this test should be passing right if the dSYMs are properly copied into the xcarchive?

luispadron:
We're still validating our releases with Bazel and don't ship with it yet but our xcarchive generated by this rule looks like:

Screenshot 2024-02-07 at 5 46 09 PM

> Regardless this test should be passing right if the dSYMs are properly copied into the xcarchive?

I thought it wouldn't pass here unless you're able to set those flags when the test runs which I don't believe the tests do.

maxwellE:
> We're still validating our releases with Bazel and don't ship with it yet but our xcarchive generated by this rule looks like:
> Screenshot 2024-02-07 at 5 46 09 PM
> 
> > Regardless this test should be passing right if the dSYMs are properly copied into the xcarchive?
> 
> I thought it wouldn't pass here unless you're able to set those flags when the test runs which I don't believe the tests do.

What if you inspect those dSYMs, do they contain valid `Contents/Resources/DWARF/TARGET_NAME` binray paths? They need those binaries to be valid dsyms

luispadron:
Superseded by https://github.com/bazelbuild/rules_apple/pull/2396 that PR adds the test and gets it green. Thanks for raising this @maxwellE

Issue #2392, opened 2024-02-05
dynamic frameworks won't strip archs since version 3.0.0


xiemotongye:
We use bazel to organize & build our iOS application. All worked fine in version 2.5.0.

After I update bazel to version 7.0 & rules_apple to 3.2.1, I was surprised to find out that dynamic frameworks won't strip archs anymore when I upload ipa to AppStoreConnect. The dynamic frameworks in final payload still includes the simulator arch.

```
ERROR ITMS-90087: Unsupported Architectures. The executable for MyApp.app/Frameworks/MyFmk.framework contains unsupported architectures '[x86_64]'.
```

I've confirmed that this issue has been around since version 3.0.0.

xiemotongye:
I've send a PR to fix this. https://github.com/bazelbuild/rules_apple/pull/2395
This issue may block the release plan of some iOS-dev teams that want to use bazel 7.
@keith Can you have a look?

keith:
thanks for debugging this! I submitted a PR similar to yours to fix what you found and cover that with a test https://github.com/bazelbuild/rules_apple/pull/2398

PR #2395, opened 2024-02-08
Fix the issue that the thinned dynamic frameworks would be overwritten by strip_bitcode.

xiemotongye:
Fix https://github.com/bazelbuild/rules_apple/issues/2392


xiemotongye:
The tests for xcframework arm64e could passed previously because the strip_bitcode overwrote the thinned binary to a fat binary.

keith:
looks like the arm64e issue is https://github.com/bazelbuild/apple_support/pull/301

keith:
thanks for submitting this! I took your fix into https://github.com/bazelbuild/rules_apple/pull/2398 which also fixes the tests, please reopen your issue if it didn't fix this!

PR #2398, opened 2024-02-08
Fix dynamic framework lipo issue

keith:
Previously when a dynamic framework was lipo'd and then bitcode
stripped, the bitcode stripping would replace the lipo'd binary and
you'd end up with a binary that contained more architectures than were
requested.

Fixes https://github.com/bazelbuild/rules_apple/issues/2392

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -6,7 +6,7 @@ module(
     repo_name = "build_bazel_rules_apple",
 )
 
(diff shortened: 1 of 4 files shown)


mattrobmattrob:
Want to do the apple_support bump in a separate change?

mattrobmattrob (approved):

keith:
We can't separate that bump because otherwise the test fails for something that was previously covered up by this bug

mattrobmattrob:
Good deal, merge away!

PR #2401, opened 2024-02-16
Allow to opt-out from library evolution

sewerynplazuk:
The pull request adds support for packaging XCFrameworks that were build with library evolution disabled. Additionally it contains changes from [Add the `apple.no_legacy_swiftinterface` feature to disable the automatic library evolution transition on framework rules](https://github.com/bazelbuild/rules_apple/pull/2104) PR that are essential for the packaging to work.

### DIFF
--- a/test/ios_xctestrun_runner_unit_test.sh
+++ b/test/ios_xctestrun_runner_unit_test.sh
@@ -448,6 +448,10 @@ ios_unit_test(
     deps = [":make_var_unit_test_lib"],
     minimum_os_version = "${MIN_OS_IOS}",
     runner = ":ios_x86_64_sim_runner",
+    env = {
+        "MY_MAKE_VAR": "\$(MY_MAKE_VAR)",
+    },
+    toolchains = [":my_make_var"],
 )
 
 ios_unit_test(
@@ -457,6 +461,10 @@ ios_unit_test(
     minimum_os_version = "${MIN_OS_IOS}",
     test_host = ":app",
     runner = ":ios_x86_64_sim_runner",
+    env = {
+        "MY_MAKE_VAR": "\$(MY_MAKE_VAR)",
+    },
+    toolchains = [":my_make_var"],
 )
 EOF
 }


apple/build_settings/build_settings.bzl
```diff
@@ -35,6 +35,12 @@ Declare a code signing identity, to be used in all code signing flows related to
     "use_tree_artifacts_outputs": struct(
         doc = """
 Enables Bazel's tree artifacts for Apple bundle rules (instead of archives).
+""",
+        default = False,
+    ),
+    "no_library_evolution": struct(
+        doc = """
+Disables library evolution and prevents the emission of .swiftinterface files.
```
acecilia, line 43: Could this be rephrased in a positive way, so is more natural to read/understand:

```
        default = True,
    ),
    "enable_library_evolution": struct(
```

sewerynplazuk: Depends on the conventions used in the rules. I didn't finy any flag like this enabled by default. As a alternative, we could go with `allow_internal_distribution` which nicely matches parameter and behavior from `xcodebuild`.

brentleyjones: I like `use_library_evolution`. We shouldn't have `no` in the name of a build setting, because then we get stuff like `--nono_library_evolution`.

apple/build_settings/build_settings.bzl
```diff
@@ -38,6 +38,12 @@ Enables Bazel's tree artifacts for Apple bundle rules (instead of archives).
 """,
         default = False,
     ),
+    "use_library_evolution": struct(
+        doc = """
+Enables library evolution, allowing for the emission of .swiftinterface files.
+""",
+        default = True,
+    ),
```
brentleyjones, line 46: How does this interact with the `apple.no_legacy_swiftinterface`, `swift.emit_swiftinterface`, and `swift.enable_library_evolution` features?

sewerynplazuk: As for `swift.emit_swiftinterface` and `swift.enable_library_evolution`, when `use_library_evolution` has a value of False, they will also be set to `False` for all transitive dependencies.

As for `apple.no_legacy_swiftinterface` (I understand you mean the one declared here https://github.com/bazelbuild/rules_apple/pull/2104) - the behavior is basically the same. From my point of view, it is more convenient to use the flag in the scope of the whole build. However, I am totally fine to use `apple.no_legacy_swiftinterface`.

test/starlark_tests/macos_static_framework_tests.bzl
```diff
@@ -88,6 +92,9 @@ def macos_static_framework_test_suite(name):
         build_type = "simulator",
         compilation_mode = "opt",
         target_under_test = "//test/starlark_tests/targets_under_test/macos:static_fmwk_with_swift_and_avoid_deps",
+        build_settings = {
+            build_settings_labels.use_library_evolution: "True",
+        },
```
sewerynplazuk, line 97: @brentleyjones Do you any idea why the test is failing after setting flag to `true`? Looks like other `build_settings` are ignored as well (i.e. `test/starlark_tests/apple_static_xcframework_import_tests.bzl#L252`) - the test is passing even when I toggle the value.

brentleyjones: I do not.

apple/build_settings/build_settings.bzl
sewerynplazuk: Do you prefer any of the approaches? Would it make sense to cherry pick `apple.no_legacy_swiftinterface` into this PR (or merge it into `main`)?

brentleyjones:
> Additionally it contains changes from Add the apple.no_legacy_swiftinterface feature to disable the automatic library evolution transition on framework rules PR that are essential for the packaging to work

We should land cherry-picks as their own PRs. I commented in that PR as well.

sewerynplazuk:
> > Additionally it contains changes from Add the apple.no_legacy_swiftinterface feature to disable the automatic library evolution transition on framework rules PR that are essential for the packaging to work
> 
> We should land cherry-picks as their own PRs. I commented in that PR as well.

@brentleyjones Makes sense. Is there anything I can do to help with #2104?

brentleyjones:
Is https://github.com/bazelbuild/rules_apple/pull/2104#issuecomment-1683179311 resolved?

PR #2403, opened 2024-02-17
Add docs on optimizing remote cache and build execution

luispadron:


doc/README.md
```diff
@@ -160,11 +161,6 @@ below.
   </tbody>
 </table>
 
-## Available options
-
-Check out [this page](common_info.md) for more information on available options
-that these rules honor during a build.
-
```
luispadron, line 167: I felt that this was hidden behind the other rule docs table and was hard to find

doc/common_info.md
```diff
@@ -521,3 +521,23 @@ exec "$bazel_real" "$@"
...
+```shell
+common --modify_execution_info=^(BitcodeSymbolsCopy|BundleApp|BundleTreeApp|DsymDwarf|DsymLipo|GenerateAppleSymbolsFile|ObjcBinarySymbolStrip|CppLink|ObjcLink|ProcessAndSign|SignBinary|SwiftArchive|SwiftStdlibCopy)$=+no-remote,^(BundleResources|ImportedDynamicFrameworkProcessor)$=+no-remote-exec
+```
+
+The following table provides a rationale for each mnemonic. In general though, the mnemonics that are excluded in `modify_execution_info` are excluded because they are large outputs which change frequently and as such are faster when run locally, or they are not generally configured for remote execution (such as signing).
+
+| Mnemonic | Rationale |
+| --- | --- |
+|ProcessAndSign | A signing action, RBE is not generally configured for signing. |
+|SignBinary | A signing action, RBE is not generally configured for signing. |
```
erikkerber, line 543: Light opinion: for the sake of brevity, I like Brentley and Keith's approach of reversing this and saying:

<Reason>: <List of mnemonic's that map to this reason>

doc/README.md
brentleyjones: We could land that as its own PR.

doc/common_info.md
brentleyjones, line 543: Any plan to list all of the mnemonics?

brentleyjones:
@luispadron Hav an idea on when this will be out of draft and land? I want to link to it from somewhere else.

luispadron:
@brentleyjones i put it up as a starter but I don't really know the rationale for a lot of these, feel free to push up for the mnemonics you know about. 

We can merge whenever too and just add more info as needed

brentleyjones:
Sure, I'll take a stab updating it.

PR #2411, opened 2024-02-22
 Allow underscores in bundle identifiers

aaronsky:
Underscores are valid characters in identifiers, so let's allow them.

### DIFF
--- a/apple/internal/bundling_support.bzl
+++ b/apple/internal/bundling_support.bzl
@@ -377,7 +377,7 @@ def _validate_bundle_id(bundle_id):
             # handled by the split(), so just have to check for '-'.
             for i in range(len(part)):
                 ch = part[i]
-                if ch != "-" and not ch.isalnum():
+                if ch not in ["-", "_"] and not ch.isalnum():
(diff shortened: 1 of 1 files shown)


mattrobmattrob (approved):

luispadron (approved):

keith (approved):
Thanks! At least as of today we tested this on iTC and you can still create bundles with these

keith:
Looks like it's good once you fix the CLA

jiawen:
Did something change at Apple to allow underscores in bundle ids? The documentation for [CFBundleIdentifier](https://developer.apple.com/documentation/bundleresources/information_property_list/cfbundleidentifier) still says A-Z, a-z, - and period.

keith:
No changes that we know of. But previously this wasn't validated and when we started validating it users who had functioning app ids with underscores got hit by this validation even though everything was working on the Apple side. Then we tested by creating a new app id with an underscore and it seems to work fine. So I think the conclusion is that the documentation is incomplete.

jiawen:
Gotcha, thanks! Nice to have a comment "paper trail" 😃.

PR #2417, opened 2024-02-28
Fix undefined symbols warning with intents stub binary

keith:
(cherry picked from commit 570ca6cf01c9827ad7e39c898c0f2f59073774a3)


keith:
The original commit here doesn't support building with `-fatal_warnings`. My second commit works around this in another way to not have that issue.

BalestraPatrick (approved):

PR #2418, opened 2024-02-29
Fix app intents with Xcode 15.3+

keith:
Depends on https://github.com/bazelbuild/rules_swift/pull/1170

### DIFF
--- a/apple/internal/aspects/BUILD
+++ b/apple/internal/aspects/BUILD
@@ -16,7 +16,9 @@ bzl_library(
     deps = [
         "//apple/internal:cc_info_support",
         "//apple/internal/providers:app_intents_info",
+        "@bazel_skylib//lib:collections",
         "@build_bazel_apple_support//lib:apple_support",
+        "@build_bazel_rules_swift//swift:module_name",
         "@build_bazel_rules_swift//swift:providers",
     ],
 )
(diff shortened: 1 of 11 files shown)


pennig:
Has the time come at last?

BalestraPatrick:
Yes, we just need a new `rules_swift` release I believe depending on when we're happy with the status of the cherry-picks there

brentleyjones:
@BalestraPatrick can you update this to point to the **rules_swift** 2.0.0-rc1 release? I'm going to create a release without this change, which is compatible but doesn't require **rules_swift** 2.0, and then once that is out we can start depending on it in order to merge this?

BalestraPatrick:
@brentleyjones Updated now. I also pushed a possible fix for one breaking change (`target_name` related), but there's another breaking change as expected for the symbol graph API which seems to be more involved since we need to switch to using the aspect instead (https://github.com/bazelbuild/rules_apple/issues/2445):
```
Error: 'struct' value has no field or method 'symbol_graph'
```

keith:
@BalestraPatrick looks like this is validated on CI now https://github.com/bazelbuild/rules_apple/issues/2488

BalestraPatrick:
I left a comment here: https://github.com/bazelbuild/rules_apple/issues/2488#issuecomment-2216991527

But it looks on rules_apple side we also need to fix the docc rules (https://github.com/bazelbuild/rules_apple/pull/2487) in order to bump to rules_swift 2.0.0 which is required for this PR to be merged.

luispadron (approved):

luispadron:
The failures are due to DocC rules needing to be updated for rules_swift 2.0, thats being done over in: #2487

PR #2426, opened 2024-03-08
Fix allow space in bundle_name for ipa_post_processor

luispadron:
`bundle_root` can contain spaces if the `bundle_name` of the target has spaces, it should be quoted to prevent issues


brentleyjones:
Are we able to add or adjust tests to cover this?

mattrobmattrob (approved):

brentleyjones (approved):

PR #2432, opened 2024-03-19
Allow .xcprivacy files in infoplists attribute of ios_application

gferon:
When submitting an app for TestFlight review, we got a new warning regarding missing API declarations (`ITMS-91053: Missing API declaration`). The easiest way to fulfill this requirement is to use Xcode to create a so-called "App Privacy" file, which is actually a small `.plist` file.

I'm not sure whether this should also apply to macOS and watchOS bundles, but I'm pretty sure the answer is yes, so it might be a good idea to allow `.xcprivacy` files on any `infoplists` attribute.

### DIFF
--- a/apple/internal/ios_rules.bzl
+++ b/apple/internal/ios_rules.bzl
@@ -2529,7 +2529,7 @@ ios_application = rule_factory.create_apple_rule(
             allowed_families = rule_attrs.defaults.allowed_families.ios,
             is_mandatory = True,
         ),
-        rule_attrs.infoplist_attrs(),
+        rule_attrs.infoplist_attrs(extra_allow_files = [".xcprivacy"]),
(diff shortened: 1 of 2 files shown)


brentleyjones:
Are the values inside of a `.xcprivacy` file supposed to be merged into the root `Info.plist`? If not, this wouldn't work.

andre-alves:
AFAIK `.xcprivacy` files are just regular resources, which can be included like this:

```bazel
apple_resource_group(
    name = "PrivacyManifest",
    resources = ["PrivacyInfo.xcprivacy"],
)
```

gferon:
Oh really 🙈 that's... surprising. Looking at the file and not using Xcode I made the wrong assumption. Thanks!

PR #2435, opened 2024-03-21
Add locales_to_include as target attribute

sebastianv1:
Ports the `apple.locales_to_include` definition to be an attribute that takes precedence. This enables one to define at the target level which locales should be copied into the bundle.


apple/internal/macos_rules.bzl
```diff
@@ -2174,6 +2189,7 @@ def _macos_dylib_impl(ctx):
         codesignopts = codesigning_support.codesignopts_from_rule_ctx(ctx),
         features = features,
         ipa_post_processor = None,
+        locales_to_include = ctx.attr.locales_to_include,
```
brentleyjones, line 2192: dylib doesn't create a bundle, does it? Generally any rule that doesn't produce a bundle shouldn't have it.

sebastianv1:
Note: The set of targets that needs this as an attribute can probably be pared down.

apple/internal/macos_rules.bzl
sebastianv1: It doesn't; I only included it as potentially lproj files _could_ get stripped in `process` if the rule also supports `allows_locale_trimming` in `rule_support.bzl`? Is it worth preventing that?

As for bundles, am I correct in assuming all bundles ultimately call `resources_partial`? So i can make it an attribute only for rules that call into that? It's the only other callsite aside from `process`

brentleyjones: If `bundle_post_process_and_sign = False,` then `_bundle_partial_outputs_files` is never called. But in general, I would only add this to ones where `bundle_post_process_and_sign = True` and `allows_locale_trimming = True`.

apple/internal/partials/resources.bzl
```diff
@@ -109,21 +109,28 @@ def _merge_root_infoplists(
...
 
     Args:
         config_vars: A dictionary (String to String) of config variables. Typically from `ctx.var`.
+        locales_to_include: A string list of locales to bundle.
 
     Returns:
         A set of locales to include or None if all should be included.
     """
-    requested_locales = config_vars.get("apple.locales_to_include")
+    config_locals_to_include = config_vars.get("apple.locales_to_include")
```
brentleyjones, line 125: Could you move this into the `elif` to prevent doing work if not needed?

apple/internal/partials/resources.bzl
```diff
@@ -109,21 +109,28 @@ def _merge_root_infoplists(
...
+    requested_locales = None
+    if len(locales_to_include):
```
brentleyjones, line 127: Do just `if locales_to_include:` work? Should be `False` if `None` or `[]`.

apple/internal/processor.bzl
```diff
@@ -269,7 +270,7 @@ def _bundle_partial_outputs_files(
 
     # Autotrim locales here only if the rule supports it and there weren't requested locales.
     config_vars = platform_prerequisites.config_vars
-    requested_locales_flag = config_vars.get("apple.locales_to_include")
+    requested_locales_flag = len(locales_to_include) or config_vars.get("apple.locales_to_include")
```
brentleyjones, line 274: Can this be `locales_to_include` instead of `len(locales_to_include)`?

apple/internal/rule_attrs.bzl
```diff
@@ -712,6 +712,17 @@ def _simulator_runner_template_attr():
...
     }
 
+def _locales_to_include_attr():
+    """Returns the attribute required to support configuring the explicit set of locales supported for the bundle."""
+    return {
+        "locales_to_include": attr.string_list(
+            mandatory = False,
+            doc = """
+A list of locales to include in the bundle. Only *.lproj directories that are matched will be copied as a part of the build.
+""",
```
brentleyjones, line 722: Can this mention the `--define` and how it interacts with it?

apple/internal/partials/resources.bzl
sebastianv1: Yeah it does. Will update

apple/internal/rule_attrs.bzl
sebastianv1: Yeah will add a note.

apple/internal/partials/resources.bzl
sebastianv1: It'd just duplicate the call to `config_vars.get("apple.locales_to_include")` which seems less clean. The call is just to a dict, right? Should be constant time?

brentleyjones: Instead I would make it an `else`, and in there make this call, then use `if` on it. Even if constant, there is no need to look in the dictionary if `locales_to_include` is set.

sebastianv1: Sounds good

apple/internal/partials/resources.bzl
```diff
@@ -109,21 +109,29 @@ def _merge_root_infoplists(
...
         A set of locales to include or None if all should be included.
     """
-    requested_locales = config_vars.get("apple.locales_to_include")
+    config_locals_to_include = config_vars.get("apple.locales_to_include")
+    requested_locales = None
+    if locales_to_include:
+        requested_locales = locales_to_include
+    else:
+        config_locals_to_include = config_vars.get("apple.locales_to_include")
+        requested_locales = config_locals_to_include.split(",") if config_locals_to_include != None else None
```
brentleyjones, line 131:
```suggestion
        requested_locales = config_locals_to_include.split(",") if config_locals_to_include else None
```

brentleyjones (approved):

luispadron (approved):

PR #2436, opened 2024-03-22
Fix iOS + SwiftUI example project UIWindow issue

Kyle-Ye:
## Related Issue/PR

#2416 cc @keith 

## Modification

Update example's Info.plist file.

> Open the newly created file and paste in it the contents found in this [example Info.plist file](https://github.com/bazelbuild/rules_apple/blob/master/examples/ios/HelloWorldSwift/Info.plist).

The [tutorial](https://github.com/bazelbuild/rules_apple/blob/master/doc/tutorials/ios-app.md) guide us to use the Info.plist file here.

But it lacks some information and will make the ipa binary to launch as compatible mode of legacy iOS app.

## Description

Before the PR:
Old Behavior

After the PR:
New Behavior

## Test

You can verify the behavior by running the [tutorial](https://github.com/bazelbuild/rules_apple/blob/master/doc/tutorials/ios-app.md).

Or

Once we delete the extra `launch_storyboard = "//examples/resources:Launch.storyboard"` in `HelloWorldSwift`. We can also reproduce it by running the following commands

```shell
cd ${PROJECT_ROOT}/examples/ios/HelloWorldSwift
# If you are running a stable bazel, you can `rm ${PROJECT_ROOT}/.bazelversion` to workaround the error
bazel run //examples/ios/HelloWorldSwift:HelloWorldSwift 
```

### DIFF
--- a/examples/ios/HelloWorldSwift/Info.plist
+++ b/examples/ios/HelloWorldSwift/Info.plist
@@ -2,6 +2,11 @@
 <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
 <plist version="1.0">
 <dict>
+	<key>UILaunchScreen</key>
+	<dict>
+		<key>UILaunchScreen</key>
+		<dict/>
+	</dict>
 	<key>CFBundleDevelopmentRegion</key>
 	<string>en</string>
 	<key>CFBundleExecutable</key>
--- a/examples/ios/HelloWorldSwift/BUILD
+++ b/examples/ios/HelloWorldSwift/BUILD
@@ -71,7 +71,6 @@ ios_application(
         "ipad",
     ],
     infoplists = [":Info.plist"],
-    launch_storyboard = "//examples/resources:Launch.storyboard",
     minimum_os_version = "15.0",
     deps = [":Sources"],
 )


Kyle-Ye:
The error message of failed CI are both "Error: 'struct' value has no field or method 'to_json'" which seems to have nothing to do with this PR.

examples/ios/HelloWorldSwift/Info.plist
```diff
@@ -2,6 +2,11 @@
 <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
 <plist version="1.0">
 <dict>
+	<key>UILaunchScreen</key>
+	<dict>
+		<key>UILaunchScreen</key>
+		<dict/>
+	</dict>
```
mattrobmattrob, line 9: This more compact form still works for me, mind updating it?
```suggestion
	<key>UILaunchScreen</key>
	<dict/>
```

mattrobmattrob (approved):
Thanks!

examples/ios/HelloWorldSwift/Info.plist
Kyle-Ye: Haven't reverse-engineered the code detects "UILaunchScreen" here, and this key-value was found and extracted by the generated binary plist file of Xcode's SwiftUI Demo App.

I would recommend that we stick to what Xcode generates in its Info.plist to maximize compatibility.

But if the compact format is working now. We can also update it if you like.

mattrobmattrob: Confirmed! I hadn't realized you'd checked that, thanks for being so thorough.

Xcode 15.2 with SwiftUI + Swift iOS app gives this in the generated `.app/Info.plist`:
```XML
	<key>UILaunchScreen</key>
	<dict>
		<key>UILaunchScreen</key>
		<dict/>
	</dict>
```

mattrobmattrob (approved):

PR #2438, opened 2024-03-26
Remove `ctx.resolve_tools` in favor of `ctx.actions.run(executable = …)` and `ctx.actions.run(tools = ...)`.

tjgq:
By passing the files_to_run for a tool into the `executable` or `tools` argument to `ctx.actions.run` or `ctx.actions.run_shell`, it's no longer necessary to pass the return values of `ctx.resolve_tools` into the `inputs` and `input_manifests` arguments.

Each struct field of the AppleMacToolsToolchainInfo and AppleXPlatToolsToolchainInfo providers previously populated with the return values of `ctx.resolve_tools` is now populated with the files_to_run for the respective tool.

The `resolved_` prefix is removed from field and variable names; in addition to it no longer making sense, this change provides proof that all usages have been audited.

This lets us retire the `ctx.resolve_tools` API in a future Bazel version.


apple/internal/codesigning_support.bzl
```diff
@@ -625,7 +623,7 @@ def _post_process_and_sign_archive_action(
       platform_prerequisites: Struct containing information on the platform being targeted.
       process_and_sign_template: A template for a shell script to process and sign as a file.
       provisioning_profile: The provisioning profile file. May be `None`.
-      resolved_codesigningtool: The `struct` from resolve_tools representing the code signing tool.
+      codesigningtool: The files_to_run for the code signing tool.
```
brentleyjones, line 626: Can you move this, and other renames (like line 603 above and 774 below), to keep the arguments alphabetical?

apple/internal/partials/support/resources_support.bzl
```diff
@@ -211,8 +211,8 @@ def _asset_catalogs(
         output_plist = assets_plist,
         platform_prerequisites = platform_prerequisites,
         product_type = product_type,
-        resolved_alticonstool = apple_mac_toolchain_info.resolved_alticonstool,
-        resolved_xctoolrunner = apple_mac_toolchain_info.resolved_xctoolrunner,
+        alticonstool = apple_mac_toolchain_info.alticonstool,
```
brentleyjones, line 214: Same alphabetical comment. I won't comment this again, just a reminder it's across multiple files.

brentleyjones (commented):
Thank you so much for this @tjgq!

apple/internal/codesigning_support.bzl
tjgq: I think I fixed all of them, PTAL.

apple/internal/codesigning_support.bzl
```diff
@@ -491,17 +491,17 @@ def _generate_codesigning_dossier_action(
...
-      entitlements: Optional file representing the entitlements to sign with.
       label_name: Name of the target being built.
+      dossier_codesigningtool: The files_to_run for the code signing tool.
       output_discriminator: A string to differentiate between different target intermediate files
           or `None`.
       output_dossier: The `File` representing the output dossier file - the zipped dossier will be placed here.
+      embedded_dossiers: An optional List of Structs generated from
+         `embedded_codesigning_dossier` that should also be included in this
+          dossier.
+      entitlements: Optional file representing the entitlements to sign with.
```
brentleyjones, line 502: All three of these should be moved up to maintain alphabetical ordering.

brentleyjones (approved):

apple/internal/codesigning_support.bzl
tjgq: Done. (I also had to make the remaining arguments positional, but that apparently works.)

luispadron:
We should list this as a breaking change in the releases i think, it broke some downstream rules since the `AppleMacToolchainInfo` provide fields have changed (and some of the resource action arguments)

brentleyjones:
That's not public API, right? I think we are free to modify that without any mention.

luispadron:
Yeah you're right. Thought it was in providers.bzl but doesn't look like it

PR #2450, opened 2024-04-18
Update docc rule to support deps and data

luispadron:
Updates the `.docc` aspect to allow both `deps` and `data` attributes as providers for the `.docc` bundle information. Previously, if the rule defined `data` but had no DocC bundle we'd lose out on any docc bundle information from the deps.

### DIFF
--- a/apple/internal/aspects/docc_archive_aspect.bzl
+++ b/apple/internal/aspects/docc_archive_aspect.bzl
@@ -79,7 +79,7 @@ def _docc_bundle_info_aspect_impl(target, ctx):
         first_docc_bundle = _first_docc_bundle(target, ctx)
         if first_docc_bundle:
             return [DocCBundleInfo(bundle = first_docc_bundle)]
-    elif hasattr(ctx.rule.attr, "deps"):
+    if hasattr(ctx.rule.attr, "deps"):
         # If this target has "deps", try to find a DocCBundleInfo provider in its deps.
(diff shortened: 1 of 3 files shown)


mattrobmattrob (approved):

mattrobmattrob:
Possible to add a test case to cover this scenario?

mattrobmattrob (approved):

luispadron:
Some weird failures on last_green, not sure theyre related as the test i wrote is passing in the other jobs

brentleyjones:
Failed on a re-run. We should confirm with a no-op PR that it still fails, and if so, start to `bazelisk bisect` to see when it broke in `bazel`.

luispadron:
CI seems a bit flaky due to simulator issues:

```
   Failure Reason: Failed to create SimDeviceSet at path /Users/buildkite/Library/Developer/Xcode/UserData/IB Support/Simulator Devices. You'll want to check the logs in ~/Library/Logs/CoreSimulator to see why creating the SimDeviceSet failed.
            Description: Unable to determine SimDeviceSet for subscriptions, set_path=/Users/buildkite/Library/Developer/Xcode/UserData/IB Support/Simulator Devices
            Underlying Errors:
    Underlying Errors:
                Failure Reason: Allocation or initialization failed.
                Description: Failed to initialize simulator device set.
```

mattrobmattrob:
I've retried the job.

Issue #1475, opened 2022-05-16
XCFramework for macOS


yuriyhanysh:
Hello!
Is there a way to create XCFramework for both iOS and macOS using `apple_xcframework`?
`minimum_deployment_os_versions` and `minimum_os_versions` attributes don't seem to work. And `ios` attribute allows you to specify only `ios` architectures.

xiemotongye:
Is there a follow-up?
I'm now facing a same issue.

luispadron:
I'd also like this so if anyone is already working on this let me know otherwise I'll likely spend some time to get this working for macOS

luispadron:
Started looking into this a bit and I believe the issue might be in Bazel core, given:

```python
apple_xcframework(
    name = "ios_and_macos_dynamic_xcframework",
    bundle_id = "com.google.example",
    extension_safe = True,
    infoplists = [
        "//test/starlark_tests/resources:Info.plist",
    ],
    ios = {
        "simulator": ["x86_64"],
        "device": ["arm64"],
    },
    minimum_os_versions = {
        "ios": common.min_os_ios.baseline,
        "macos": common.min_os_macos.baseline,
    },
    public_hdrs = [
        "//test/starlark_tests/resources:shared.h",
    ],
    tags = common.fixture_tags,
    deps = [":fmwk_lib"],
)
```

I get the following output when examining the [split transition keys used here](https://github.com/bazelbuild/bazel/blob/ff63e0d0499d6d0ac82dcd3e58935f8718bd46f3/src/main/starlark/builtins_bzl/common/objc/linking_support.bzl#L213C50-L213C85):

```
Error in fail: [("ios_arm64_device", <target //apple:default_cc_toolchain_forwarder>), ("ios_x86_64_simulator", <target //apple:default_cc_toolchain_forwarder>)]
```

luispadron:
I've got something in draft now that generates the expected XCFramework: https://github.com/bazelbuild/rules_apple/pull/2459

Issue #2419, opened 2024-03-03
How to build for multiple architectures


luispadron:
When trying to build a FAT binary for a `macos_dynamic_framework` using `--macos_cpus=x86_64,arm64` I hit the following error:

```
Traceback (most recent call last):
        File "/private/var/tmp/_bazel_lpadron/ad4b52c6b7f1c71771bd38a7ed554766/external/rules_apple~3.2.1/apple/internal/macos_rules.bzl", line 2946, column 13, in _macos_dynamic_framework_impl
                fail(
Error in fail:                     error: Swift dynamic frameworks expect a single swift_library dependency.
    
```

This build works as-is when building for a single architecture. This is with Bazel `6.4.0` and rules_apple `3.2.1`

luispadron:
I see the same with `bazel build //test/starlark_tests/targets_under_test/ios:basic_framework --ios_multi_cpus=x86_64,arm64` in the tests here, is there a different way to generate universal frameworks?

keith:
Hrm did this ever work? Might be worth bisecting (or just testing the first sha that had this rule)

A potential fix is to fetch deps from `ctx.split_attr.deps` instead, but I also wonder if we're misremembering how it should work

luispadron:
This doesn't seem to have worked as far back as 1.0.0 using Bazel 5

luispadron:
My use case right now is we want to release a macOS application and just noticed it was running under Rosetta because it was building using x86_64 github runners. `apple_univaral_binary` seems to have the same error

keith:
Maybe this generally never worked and then it became virtually obsolete because xcframeworks solve multi-archs (even for macOS although it feels unnecessary since there aren't platform arch conflicts)

keith:
What rule is the application target? With a macOS application rule it should work. And if it fails because a macos_dynamic_framework is in the dependency tree you should replace that with a macos_framework. Theoretically the former should only be used for distribution

luispadron:
Ah that was it, makes sense. Thanks!

So in summary for folks:

- If distributing `macos_*_framework` which requires multiple arch support use `xcframework` rules instead.
- If depending on a `macos_*_framework` rule use `macos_framework` instead within other targets.

luispadron:
I guess we cant use `.xcframework` on macOS until #1475

keith:
yea, I guess that was less of a priority since you can still use a fat framework there, but then i guess you hit this even on macOS?

[...]

PR #2459, opened 2024-05-26
Add support for macos in `apple_*_xcframework` rules

luispadron:
This commit adds a `macos` attr to both the `apple_xcframework` and `apple_static_xcframework` rules.
This allows building for multiple architectures and platforms.

The new `macos` attr is a `string_list` (unlike the `ios` attr which is a dict for the simulator/device split).

Example:

```starlark
apple_xcframework(
    name = "ios_and_macos_dynamic_xcframework",
    bundle_id = "com.google.example",
    extension_safe = True,
    infoplists = [
        "//test/starlark_tests/resources:Info.plist",
    ],
    ios = {
        "simulator": ["x86_64"],
        "device": ["arm64"],
    },
    macos = [
        "arm64",
        "x86_64",
    ],
    minimum_os_versions = {
        "ios": common.min_os_ios.baseline,
        "macos": common.min_os_macos.baseline,
    },
    public_hdrs = [
        "//test/starlark_tests/resources:shared.h",
    ],
    tags = common.fixture_tags,
    deps = [":fmwk_lib"],
)
```

Fixes https://github.com/bazelbuild/rules_apple/issues/1475


apple/internal/xcframework_rules.bzl
```diff
@@ -864,6 +864,11 @@ built for those platform variants (for example, `x86_64`, `arm64`) as their valu
 A dictionary of strings indicating which platform variants should be built for the tvOS platform (
 `device` or `simulator`) as keys, and arrays of strings listing which architectures should be
 built for those platform variants (for example, `x86_64`, `arm64`) as their values.
+""",
+            ),
+            "macos": attr.string_list(
+                doc = """
+A list of strings indicating which architecture should be built for the macOS platform (for example, `x86_64`, `arm64`).
```
luispadron, line 871: Keeping this a `string_list_dict` keeps the logic simpler, and maybe we just fail if folks pass `simulator`?

apple/internal/cc_info_support.bzl
```diff
@@ -32,7 +32,8 @@ def _get_all_deps(*, deps, split_deps_keys = []):
 
     all_deps = []
     for split_deps_key in split_deps_keys:
-        all_deps += deps[split_deps_key]
+        if split_deps_key in deps:
+            all_deps += deps[split_deps_key]
```
luispadron, line 36: @keith @brentleyjones this fixes an issue introduced by adding the `macos` attr to `apple_static_xcframework`:

```
ERROR: /Users/lpadron/Development/rules_apple/test/starlark_tests/targets_under_test/apple/BUILD:942:25: in apple_static_xcframework rule //test/starlark_tests/targets_under_test/apple:ios_static_xcframework_with_resources: 
        File "/Users/lpadron/Development/rules_apple/apple/internal/xcframework_rules.bzl", line 943, column 83, in _apple_static_xcframework_impl
                link_outputs_by_library_identifier = _group_link_outputs_by_library_identifier(
        File "/Users/lpadron/Development/rules_apple/apple/internal/xcframework_rules.bzl", line 205, column 45, in _group_link_outputs_by_library_identifier
                if swift_support.uses_swift(deps[split_attr_key]):
Error: key "ios_arm64e_device" not found in dictionary
```

How should we properly fix this? Is the `apple_rule_arm64_as_arm64e_transition` being applied here for some reason?

luispadron: This still leaves us with:

```
ERROR: /Users/lpadron/Development/rules_apple/test/starlark_tests/targets_under_test/apple/BUILD:508:8: declared output 'test/starlark_tests/targets_under_test/apple/ios_xcframework_bundling_static_fmwks.xcframework/ios-arm64_x86_64-simulator/ios_xcframework_bundling_static_fmwks.framework/Headers/ios_xcframework_bundling_static_fmwks.h' was not created by genrule. This is probably because the genrule actually didn't create this output, or because the output was a directory and the genrule was run remotely (note that only the contents of declared file outputs are copied from genrules run remotely)
```

luispadron: AFAICT adding `macos` changes the transitions to add `arm64e`, maybe this is related to #2419 as well since that's a split deps issue as well.

apple/internal/resource_actions/plist.bzl
```diff
@@ -341,19 +341,21 @@ def merge_root_infoplists(
     else:
         plist_key = "MinimumOSVersion"
 
-    input_files.append(environment_plist)
+    if environment_plist:
+        input_files.append(environment_plist)
+        forced_plists.append(environment_plist.path)
```
luispadron, line 346: These aren't created for the `macos` platform (e.g. `environment_plist` is `None` here)

apple/internal/transition_support.bzl
```diff
@@ -658,15 +659,22 @@ def _xcframework_transition_impl(settings, attr):
 
     # TODO(b/288582842): Update for visionOS when we're ready to support it in XCFramework rules.
     for platform_type in ["ios", "tvos", "visionos", "watchos", "macos"]:
-        if not hasattr(attr, platform_type):
+        platform_attr = getattr(attr, platform_type, None)
```
luispadron, line 662: Use `getattr` with a `None` default instead of `hasattr` which will still return true even when `ios` for example is the default value of `{}`, this was causing a `apple_xcframework` with only a `macos` attr to also go through this transition for the `ios` platform. 

With `getattr` we correctly get the value, for example the empty dict `{}`, which then matches on the `if not` check (it's an empty dict which is falsy), where `hasattr` would've returned true since the `ios` attr is technically available.

luispadron:
@brentleyjones @keith mind reviewing this when you have a chance?

brentleyjones (approved):

PR #2461, opened 2024-05-29
Add resource collection for `cc_*` rules

luispadron:
Adds support for `data` attributes of `cc_library` and `cc_import` rules. These attributes were previously being ignored during bundling.

### DIFF
--- a/apple/internal/aspects/resource_aspect.bzl
+++ b/apple/internal/aspects/resource_aspect.bzl
@@ -139,7 +139,7 @@ def _apple_resource_aspect_impl(target, ctx):
         if ctx.rule.attr.srcs or ctx.rule.attr.non_arc_srcs or ctx.rule.attr.deps:
             owner = str(ctx.label)
 
-    elif ctx.rule.kind == "objc_import":
+    elif ctx.rule.kind in ["cc_import", "objc_import"]:
         collect_args["res_attrs"] = ["data"]
 
     elif ctx.rule.kind == "swift_library":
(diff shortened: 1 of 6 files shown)


gkoreman:
Ok, I've tried it out and it is definitely pulling in (some) files, but there are a few issues:
1. The runfiles folder structure is not preserved. All files are collapsed to the root of Resources.
2. Dylibs are not included in the package. When creating a cc_binary the dylibs are placed in folders inside the runfiles folder.

apple/internal/aspects/resource_aspect.bzl
```diff
@@ -131,7 +131,18 @@ def _apple_resource_aspect_impl(target, ctx):
     # The name of the bundle directory to place resources within, if required.
     bundle_name = None
 
-    if ctx.rule.kind == "objc_library":
+    if ctx.rule.kind == "cc_library":
```
brentleyjones, line 134: nit: can we list the `cc_*` rules last, since they will be less common, and thus can be in the slow path?

apple/internal/aspects/resource_aspect.bzl
```diff
@@ -131,7 +131,18 @@ def _apple_resource_aspect_impl(target, ctx):
...
+        collect_args["res_attrs"] = ["data"]
+
+        # Only set cc_library targets as owners if they have srcs or deps. This
+        # treats cc_library targets without sources as resource aggregators.
+        if ctx.rule.attr.srcs or ctx.rule.attr.deps:
+            owner = str(ctx.label)
+
+    elif ctx.rule.kind == "cc_import":
```
brentleyjones, line 142: Could consolidate this with `"objc_import"` with an `in` check?

luispadron: Updated

apple/internal/aspects/resource_aspect.bzl
luispadron: updated

brentleyjones (approved):

PR #2479, opened 2024-06-26
Add use_runfiles aspect_hint to include runfiles for specific cc_libr…

### LINKED ISSUE #2477: Runfiles not bundled with macos_application
Building a macos_application with a cc_library does not include runfiles.

```
cc_library(
    name = "libapp",
    srcs = ["main.cpp",],
    data = [":data.txt"],
)

macos_application(
    name = "app_macos",
    deps = [":libapp"],
)
```

`data.txt` is not packaged in the macos_application.

gkoreman:
Fixes #2477

Include all data from cc_libraries as runfiles or resources, depending on aspect_hints.
Provide aspect_hints for resource collection for all resource targets (eg, swift_library, cc_library, etc)

By default all data is now included from cc_libraries as runfiles and follows the expected runfiles folder structure, with files retaining their nested folders and being placed in /external/pkg_name/some/file.txt when included from a http_archive or local_repository.

The default behavior for other resource collecting targets like swift_library has not changed, and will continue to collect and process data as resources.

To allow modification of this default behavior, users may add aspect_hints to the target cc_library/swift_library/etc. There are three supported aspect_hints:
@build_bazel_rules_apple//apple:use_runfiles
@build_bazel_rules_apple//apple:use_resources
@build_bazel_rules_apple//apple:suppress_resources

##### Example
Here is an example of modifying the default behavior to bundle data.txt as a resource instead of a runfile.
```
cc_library(
    name = "libapp",
    srcs = ["main.cpp",],
    data = [":data.txt"],
    aspect_hints = ["@build_bazel_rules_apple//apple:use_runfiles"],
)

macos_application(
    name = "app_macos",
    deps = [":libapp"],
)
```
data.txt is bundled in Contents/Resources/data.txt

#### Note
Hints apply only to the target and do not affect transitive deps, however if a target includes runfiles then all runfiles are bundled (including transitive runfiles) regardless of the hints applied to transitive targets.


brentleyjones:
I would like the flip the default, such that runfiles are collected by default for cc_library (but only it and the rules that currently don’t collect resources), and you need to use the aspect hint to make new rules collect resources instead. This will make it so the least number of people targets to use aspect hints. We will then need another aspect hint that tells the currently support resource collection rules to use runfiles instead. And we can have it be a build error when both are provided.

luispadron (approved):
Sweet! Left some comments I think we should address.

You'll also need to re-run buildifier as we updated the version and it adds more lints/formatting fixes

luispadron (approved):

luispadron:
@brentleyjones Would you have some time to review this as well? Thank you!

apple/internal/aspects/resource_aspect.bzl
```diff
@@ -125,48 +129,85 @@ def _apple_resource_aspect_impl(target, ctx):
...
         collect_args["res_attrs"] = ["data"]
         owner = str(ctx.label)
 
     elif ctx.rule.kind == "apple_resource_group":
+        default_action = "RESOURCES"
         collect_args["res_attrs"] = ["resources"]
         collect_structured_args["res_attrs"] = ["structured_resources"]
 
     elif ctx.rule.kind == "apple_resource_bundle":
+        default_action = "RESOURCES"
```
brentleyjones, line 196: Instead of a string can we make this a constant somewhere?

apple/internal/aspects/resource_aspect.bzl
```diff
@@ -125,48 +129,85 @@ def _apple_resource_aspect_impl(target, ctx):
...
         collect_infoplists_args["res_attrs"] = ["infoplists"]
         collect_args["res_attrs"] = ["resources"]
         collect_structured_args["res_attrs"] = ["structured_resources"]
         process_args["bundle_id"] = ctx.rule.attr.bundle_id or None
         bundle_name = "{}.bundle".format(ctx.rule.attr.bundle_name or ctx.label.name)
 
+    if hint_action != None:
```
brentleyjones, line 203: Does this need to check against `None` versus a `not hint_action`?

apple/internal/aspects/resource_aspect_hint.bzl
```diff
@@ -0,0 +1,121 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Implementation of the `apple_resource_hint` rule."""
+
+_resource_actions = ["RESOURCES", "RUNFILES", "SUPPRESS"]
```
brentleyjones, line 17: Let's be sure to use the constants here as well.

apple/internal/aspects/resource_aspect_hint.bzl
```diff
@@ -0,0 +1,137 @@
...
+```build
+# //my/project/BUILD
+cc_library(
+    name = "somelib",
+    data = ["mydata.txt"],
+    aspect_hints = ["@build_bazel_rules_apple//apple:use_resources"],
+)
+```
+
+#### Collect runfiles of a objc_library
```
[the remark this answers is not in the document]
gkoreman: I am hesitant to add a root folder (eg "_BazelRunfiles") as the runfiles system already follows a prescriptive folder structure based on project layout. Users will already expect this layout and adding a root folder may be confusing. This is similar to pkg_tar that will package runfiles without any extra subfolders.
Thoughts?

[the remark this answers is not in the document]
brentleyjones: Yeah, maybe. I'm just concerned with these files overiding/shadowing `data` files. Especially on iOS where there is no `Resources` folder, it's just in the root bundle. Maybe it's not an issue though?

gkoreman:
Runfiles are only placed in the root folder if that is where they exist in
the workspace so it is unlikely that there will be conflicts.

In simple projects I would expect this not to be an issue because the user
will likely choose either to use runfiles or resources but not both.

Larger projects with more complex transitive runfiles will have those files
end up in subdirectories anyway so there shouldn’t be an issue there either.

brentleyjones (approved):

jiawen:
> Hints apply only to the target and do not affect transitive deps, however if a target includes runfiles then all runfiles are bundled (including transitive runfiles) regardless of the hints applied to transitive targets.

I had a hard time parsing the second part of this sentence. After some experimentation, I discovered that if `use_resources` is set for a leaf `cc_library`, then for it to be propagated correctly, intermediate `cc_library` targets also need to have `use_resources` set. That is, all targets up the tree from a leaf `cc_library` up to a resource collecting target (an `objc_library` (and probably `swift_library` I imagine) need to have `use_resources` set for it to be processed and bundled correctly.

@gkoreman I'm sure you meant this to be the case in your note, but my brain had a hard time parsing this. Of course it makes sense in hindsight.

I didn't have a chance to experiment with the other three common cases:
1. Every target up the tree sets `use_runfiles` instead. I expect the behavior to be the same.
2. The direct dep has `use_resources` but transitive deps set `use_runfiles`.
3. The opposite: direct dep has `use_runfiles` but transitive deps set `use_resources`.

PR #2491, opened 2024-07-10
Re-add fixed releaser

luispadron:
Turns out because were on bazelbuild org a fork of the registry isnt possible and so the fixed releaser must be set to tell the bot where the fork lives.


mattrobmattrob (approved):
🦅

PR #2499, opened 2024-07-12
Sort load statements per buildifier 6.4.0

erikkerber:
Followup request from #2485 

rules_apple is on buildifier 6.3.2, one version before load sorting improvements. Can update that too.

PR #2500, opened 2024-07-12
Use buildifier 6.4.0 for improved load sorting

erikkerber:
Followup to/depends on #2499 

Bigger PR than 2499 due to me not including `.bzl` files. I was using a command from my history that just operated on BUILD/MODULE files. Big enough it might make upstream merges a little painful? I'll leave that thought to the maintainers.

**6.3.3:** `out_of_order_load` warning is enabled by default
**6.4.0:** More correct sorting order of load statements

Could bump to 7.1.2 which is latest, but not necessary for auto fixing of load sorting

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -44,4 +44,4 @@ tasks:
     - "doc/..."
     <<: *common_last_green
 
-buildifier: 6.3.2
+buildifier: 6.4.0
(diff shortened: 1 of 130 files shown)


luispadron (approved):

brentleyjones (approved):
We already have merge conflicts on all load statements, so this is fine.

PR #2522, opened 2024-09-05
Use `openssl` outside of Linux in `local_provisioning_profile_finder.py`

brentleyjones:


tools/local_provisioning_profile_finder/local_provisioning_profile_finder.py
```diff
@@ -31,7 +32,23 @@ def _build_parser() -> argparse.ArgumentParser:
 
 
 def _profile_contents(profile: str) -> Tuple[str, datetime.datetime, str]:
-    output = subprocess.check_output(["security", "cms", "-D", "-i", profile])
+    if _USE_SECURITY:
+        output = subprocess.check_output(
+            ["security", "cms", "-D", "-i", profile],
+        )
+    else:
+        # We call it this way to silence the "Verification successful" message
```
mattrobmattrob, line 40: What do you mean when you say "this way" here? Just the verify no verify command structure?

mattrobmattrob (approved):

tools/local_provisioning_profile_finder/local_provisioning_profile_finder.py
brentleyjones: I mean the `try:except` with `subprocess.run` instead of `subprocess.check_output`.

luispadron (approved):

Issue #2524, opened 2024-09-09
apple_xcframework: Invalid Info.plist generated for xcframework supporting both device and simulator


zavsby:
I'm using apple_xcframework to generate a xcframework for ios (arm64 + arm64 simulator). But Info.plist for frameworks is always generated same for simulator and device with same value in CFBundleSupportedPlatforms, DTPlatformName and DTSDKName.
It depends what is passed to build command in `--ios_multi_cpus` or not passed at all, but anyway it is always either `iphonesimulator` or `iphoneos` for both ios and simulator frameworks.

```
apple_xcframework(
    name = "SwiftProtobuf",
    deps = [":SwiftProtobufLib"],
    bundle_id = "org.swift.protobuf.SwiftProtobuf",
    extension_safe = True,
    infoplists = [":Info.plist"],
    ios = {"device": ["arm64"], "simulator": ["arm64"]},
    minimum_os_versions = {"ios": "15.0"},
    minimum_deployment_os_versions = {"ios": "15.0"},
)

swift_library(
    name = "SwiftProtobufLib",
    srcs = glob(["*.swift"]),
    module_name = "SwiftProtobuf",
    features = [
        SWIFT_FEATURE_ENABLE_LIBRARY_EVOLUTION,
    ]
)
```

`bazel build //Sources/SwiftProtobuf:SwiftProtobuf --apple_platform_type=ios --ios_multi_cpus=sim_arm64,arm64`

I've found that issue is in platform_support.bzl in `_platform_prerequisites` method which returns same `platform` for any environment.
Maybe I'm doing something wrong in configuration and some fix already exists?

PR #2526, opened 2024-09-09
Fix apple_xcframework framework plist generated for invalid platform

zavsby:
Fixing issue https://github.com/bazelbuild/rules_apple/issues/2524: using apple_xcframework to generate a xcframework for ios (arm64 + arm64 simulator), Info.plist for frameworks is always generated same for simulator and device with same values in CFBundleSupportedPlatforms, DTPlatformName and DTSDKName.


luispadron (commented):
Is there a test we could add for this behavior

PR #2527, opened 2024-09-10
Use devicectl to install and launch *_application targets on a physical device

CognitiveDisson:
In Xcode 15, Apple introduced a new command-line tool that allows management of physical devices. This makes it possible to install and launch *_application targets on a physical device without third-party solutions.

With these changes, you can now use `bazel run //:iOSApp --ios_multi_cpus=arm64`.

Previously, this command would attempt to launch the application on a simulator and fail due to an invalid architecture.

Bazel currently lacks a specific flag for specifying device identifiers suitable for this purpose. I created a [PR to Bazel](https://github.com/bazelbuild/bazel/pull/23599) introducing an `--ios_device` flag. However, it's unlikely to be accepted since it's a platform-specific flag.  If it is accepted, we can utilize it in the future.

For the current implementation, I added astring_flagthat allows passing a device identifier. It can be used like this:--@build_bazel_rules_apple//apple/build_settings:ios_device=<uuid|ecid|serial_number|udid|name|dns_name&gt; .

Note: `//doc:check_apple` failing on master

### DIFF
--- a/apple/internal/run_support.bzl
+++ b/apple/internal/run_support.bzl
@@ -158,7 +158,7 @@ def _register_macos_executable(
 
 # Define the loadable module that lists the exported symbols in this file.
 run_support = struct(
-    register_simulator_executable = _register_simulator_executable,
     register_device_executable = _register_device_executable,
     register_macos_executable = _register_macos_executable,
+    register_simulator_executable = _register_simulator_executable,
 )
--- a/apple/internal/run_support.bzl
+++ b/apple/internal/run_support.bzl
@@ -76,6 +76,59 @@ def _register_simulator_executable(
         },
     )
 
+def _register_device_executable(
+        *,
+        actions,
+        bundle_extension,
+        bundle_name,
+        label_name,
+        output,
+        platform_prerequisites,
+        predeclared_outputs,
+        rule_descriptor,
+        runner_template,
+        device = None):
+    """Registers an action that runs the bundled app on a physical device.
+
+    Args:
+      actions: The actions provider from ctx.actions.
+      bundle_extension: Extension for the Apple bundle inside the archive.
+      bundle_name: The name of the output bundle.
+      label_name: The name of the target.
+      output: The `File` representing where the executable should be generated.
+      platform_prerequisites: Struct containing information on the platform being targeted.
+      predeclared_outputs: Outputs declared by the owning context. Typically from `ctx.outputs`
+      rule_descriptor: The rule descriptor for the given rule.
+      runner_template: The simulator runner template as a `File`.
+      device: The identifier of the device ( <uuid|ecid|serial_number|udid|name|dns_name> ).
+    """
+
+    device = str(device or "")
+    minimum_os = str(platform_prerequisites.minimum_os)
+    platform_type = str(platform_prerequisites.platform_type)
+    archive = outputs.archive(
+        actions = actions,
+        bundle_name = bundle_name,
+        bundle_extension = bundle_extension,
+        label_name = label_name,
+        platform_prerequisites = platform_prerequisites,
+        predeclared_outputs = predeclared_outputs,
+        rule_descriptor = rule_descriptor,
+    )
+
+    actions.expand_template(
+        output = output,
+        is_executable = True,
+        template = runner_template,
+        substitutions = {
+            "%app_name%": bundle_name,
+            "%ipa_file%": archive.short_path,
+            "%minimum_os%": minimum_os,
+            "%platform_type%": platform_type,
+            "%device%": device,
+        },
+    )
+
 def _register_macos_executable(
         *,
         actions,
@@ -106,5 +159,6 @@ def _register_macos_executable(
 # Define the loadable module that lists the exported symbols in this file.
 run_support = struct(
     register_simulator_executable = _register_simulator_executable,
+    register_device_executable = _register_device_executable,
     register_macos_executable = _register_macos_executable,
 )
--- a/apple/internal/templates/apple_device.template.py
+++ b/apple/internal/templates/apple_device.template.py
@@ -173,13 +173,12 @@ class Device(collections.abc.Mapping):
       return self.is_apple_vision
     else:
       raise ValueError(
-          f"Apple platform type not supported for simulator: {platform_type}."
+          f"Apple platform type not supported for running on device: {platform_type}."
       )
 
 
 def os_version_number_to_int(version: str) -> int:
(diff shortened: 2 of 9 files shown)


brentleyjones:
> Unfortunately, Bazel currently lacks a specific flag for specifying device identifiers suitable for this purpose.

Can't we make a new startlark flag?

CognitiveDisson:
> ios_simulator_device

I was planning to do it, but I think it shouldn't block this PR. Even if it's accepted, the flag will be available only in the next Bazel version, if I understand that correctly.

apple/internal/run_support.bzl
```diff
@@ -106,5 +159,6 @@ def _register_macos_executable(
 # Define the loadable module that lists the exported symbols in this file.
 run_support = struct(
     register_simulator_executable = _register_simulator_executable,
+    register_device_executable = _register_device_executable,
     register_macos_executable = _register_macos_executable,
```
brentleyjones, line 163:
```suggestion
    register_device_executable = _register_device_executable,
    register_macos_executable = _register_macos_executable,
    register_simulator_executable = _register_simulator_executable,
```

brentleyjones (approved):

doc/tutorials/ios-app.md
```diff
@@ -391,8 +391,18 @@ rules.
 
 ### Install the app on a device
 
-The easiest way to install the app on the device is to launch Xcode and use the
-`Windows > Devices` command. Select your plugged-in device from the list on the
+You can install the app on a physical device using the bazel run command, just as you would for a simulator:
+
+```bash
+bazel run //:iOSApp --ios_multi_cpus=arm64
+```
```
luispadron, line 398: We should also recommend users use a `common --flag_alias=ios_device=@rules_apple//apple/build_settings:ios_device` in their `.bazelrc` so they can just do `bazel run //FooApp --ios_device="id"`

doc/tutorials/ios-app.md
```diff
@@ -391,8 +391,18 @@ rules.
...
+
+The runner will find any available physical device with an OS version higher than `minimum_os_version`.
+If you want to specify a particular device, you can specify it like this `--@build_bazel_rules_apple//apple/build_settings:ios_device=<uuid|ecid|serial_number|udid|name|dns_name>`.
+To see a list of available devices, use `xcrun devicectl list devices`.
+
+Another way to install the app on the device is to launch Xcode and use the
+`Window > Devices and Simulators` command. Select your plugged-in device from the list on the
 left, then add the app by clicking the **Add** (plus sign) button under
 "Installed Apps" and selecting the `.ipa` file that you built.
```
luispadron, line 407: Is this applicable to `rules_apple`?

apple/internal/templates/apple_device.template.py
```diff
@@ -0,0 +1,493 @@
...
+    return plist["CFBundleIdentifier"]
+
+
+def devicectl_launch_environ() -> Dict[str, str]:
+  """Calculates an environment dictionary for running `devicectl device process launch`."""
+  # Pass environment variables prefixed with "IOS_" to the device, replace
+  # the prefix with "DEVICECTL_CHILD_". bazel adds "IOS_" to the env vars which
+  # will be passed to the app as prefix to differentiate from other env vars. We
+  # replace the prefix "IOS_" with "DEVICECTL_CHILD_" here, because "devicectl" only
+  # pass the env vars prefixed with "DEVICECTL_CHILD_" to the app.
```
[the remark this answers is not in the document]
CognitiveDisson: It is expected by `devicectl`, similar to `simctl` ( similar to `apple_simulator.template.py` ). Unfortunately, I couldn't find any official documentation about it. I found this information in `xcrun devicectl device process launch --help`.

apple/internal/templates/apple_device.template.py
```diff
@@ -0,0 +1,493 @@
...
+  return result
+
+
+def discover_best_compatible_device(
+    *,
+    platform_type: str,
+    devicectl_path: str,
+    minimum_os: str,
+) -> Optional[Device]:
+  """Discovers the best compatible device.
```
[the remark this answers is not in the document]
CognitiveDisson: Done

doc/tutorials/ios-app.md
CognitiveDisson: Done

doc/tutorials/ios-app.md
CognitiveDisson: This universal method for installing an `.ipa` file on a device isn't specific to rules_apple. I've retained it as an alternative to `bazel run` since it was originally included in this documentation.

CognitiveDisson:
> This is great so far i have some very minor nits about docs and some questions about whats expected now when doing `bazel run` for an app.
> 
> Is it now the default that doing `bazel run //FooApp` where `FooApp` is an `ios_application` will run on a device if it's connected to the host machine? I think this make sense if so, just checking! If it is the new default I think we should make it a bit more obvious in the docs

Yes, it's now the default behavior. The choice between running on a simulator or a device depends on the target architecture. However, it's worth noting that users typically build this target rather than running it directly.

To me, it looks like this is emphasized enough in the documentation, but I'm open to suggestions.

luispadron:
> > This is great so far i have some very minor nits about docs and some questions about whats expected now when doing `bazel run` for an app.
> 
> > 
> 
> > Is it now the default that doing `bazel run //FooApp` where `FooApp` is an `ios_application` will run on a device if it's connected to the host machine? I think this make sense if so, just checking! If it is the new default I think we should make it a bit more obvious in the docs
> 
> 
> 
> Yes, it's now the default behavior. The choice between running on a simulator or a device depends on the target architecture. However, it's worth noting that users typically build this target rather than running it directly.
> 
> 
> 
> To me, it looks like this is emphasized enough in the documentation, but I'm open to suggestions.

Ah makes sense it'll run on a device based on the arch, thank you!

doc/tutorials/ios-app.md
```diff
@@ -391,8 +391,19 @@ rules.
 
 ### Install the app on a device
 
-The easiest way to install the app on the device is to launch Xcode and use the
-`Windows > Devices` command. Select your plugged-in device from the list on the
+You can install the app on a physical device using the bazel run command, just as you would for a simulator:
```
luispadron, line 394:
```suggestion
You can install the app on a physical device using a bazel run command when targeting a specific device architecture:
```

luispadron (approved):

Issue #2540, opened 2024-09-23
apple_static_xcframework - resulting XCFramework contains dependencies bundles passed as avoid_deps 


sewerynplazuk:
As stated in the title, the result of `apple_static_xcframework` rule contains all the bundles of all dependencies of the packaged framework, even if the dependencies are passed as `avoid_deps`.

This seems unexpected from the user perspective and leads to:
- excessive size of resulting XCFramework
- duplicated bundles when dependencies are packed into XCFramework independently
    - `ios_application` will throw `Multiple files would be placed at "..." in the bundle, which is not allowed. check input file` error

PR #2541, opened 2024-09-23
Fix apple_static_xcframework rule to exclude bundles of dependencies marked as avoid_deps

sewerynplazuk:
Fixes https://github.com/bazelbuild/rules_apple/issues/2540.

~Alternatively, I thought about not including resources from the dependencies passed in `avoid_deps`, but I'm not sure that this would not be a breaking change.~


brentleyjones:
> Alternatively, I thought about not including resources from the dependencies passed in `avoid_deps`, but I'm not sure that this would not be a breaking change.

That seems like the right way to go? I'll wait for other opinions though.

luispadron:
> > Alternatively, I thought about not including resources from the dependencies passed in `avoid_deps`, but I'm not sure that this would not be a breaking change.
> 
> That seems like the right way to go? I'll wait for other opinions though.

When would we expect to use `avoid_bundles` but not `avoid_deps`? If its never, then we should just use `avoid_deps` for this imo

sewerynplazuk:
@brentleyjones @luispadron Thanks guys for your input. As `avoid_bundles` would rather be used always together with `avoid_deps` I have updated the PR to rely on the latter instead.

luispadron (approved):

brentleyjones (approved):

PR #2549, opened 2024-10-04
Fix wrong os in xcframework frameworks infoplist

themz:
This PR aims to fix an issue where the Info.plist contains an incorrect value in the xcframework multiplatform bundle.

```
apple_xcframework(
    ...
    ios = {
        "simulator": ["x86_64"],
        "device": ["arm64"],
    },
    ..
)
```
In this configuration, the Info.plist for x86_64 will contain the 'iphoneos' value for DTPlatformName, but it is supposed to be iphonesimulator.

Related to #2524

### DIFF
--- a/apple/internal/platform_support.bzl
+++ b/apple/internal/platform_support.bzl
@@ -29,6 +29,13 @@ _DEVICE_FAMILY_VALUES = {
     "mac": None,
 }
 
+# Align with Migrate apple_common.platform to Starlark implimentaion
+TARGET_ENVIROMENT = struct(
+    device = "device",
+    catalyst = "macabi",
+    simulator = "simulator",
+)
+
 def _ui_device_family_plist_value(*, platform_prerequisites):
     """Returns the value to use for `UIDeviceFamily` in an info.plist.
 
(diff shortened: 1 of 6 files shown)


apple/internal/platform_support.bzl
```diff
@@ -88,13 +96,26 @@ def _platform_prerequisites(
       platform_type_string: The platform type for the current target as a string.
       uses_swift: Boolean value to indicate if this target uses Swift.
       xcode_version_config: The `apple_common.XcodeVersionConfig` provider from the current context.
+      environment: "device" or "simulator" environment of the current target. Optional.
 
     Returns:
       A struct representing the collected platform information.
     """
     platform_type_attr = getattr(apple_common.platform_type, platform_type_string)
     platform = apple_fragment.multi_arch_platform(platform_type_attr)
```
themz, line 105: [review-note]
Here we don't have an option to get correct platform for simulator as [platform_type](https://bazel.build/rules/lib/toplevel/apple_common#platform_type) is can be only ios, tvos but [platform](https://bazel.build/rules/lib/toplevel/apple_common#platform) can be  ios_simulator as well as ios_device. Same for tvos_device and tvos_simulator.

Issue #2555, opened 2024-10-10
[Bazel CI]Error in glob: glob pattern '*.jar' didn't match anything



sgowroji:
CI: https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/4189#019274a0-a407-4de1-9306-84ef2daf9eb2

 Platform: Ubuntu

 Logs: 
```
ERROR: Traceback (most recent call last):
	File "/var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/d4ff19de6e746f1679fe4bc1387beb8f/external/stardoc+/stardoc/BUILD", line 72, column 16, in <toplevel>
		srcs = glob(["*.jar"]),
Error in glob: glob pattern '*.jar' didn't match anything, but allow_empty is set to False (the default value of allow_empty can be set with --incompatible_disallow_empty_glob).
(04:21:52) ERROR: /var/lib/buildkite-agent/builds/bk-docker-8rpq/bazel-org-repo-root/rules_apple/doc/BUILD.bazel:27:12: Target '@@stardoc+//stardoc:templates/markdown_tables/repository_rule.vm' contains an error and its package is in error and referenced by '//doc:visionos.doc_doc'
(04:21:52) ERROR: Analysis of target '//doc:check_visionos.doc' failed; build aborted: Analysis failed
(04:21:52) INFO: Elapsed time: 2.420s, Critical Path: 0.03s
(04:21:52) INFO: 1 process: 1 internal.
(04:21:52) ERROR: Build did NOT complete successfully
(04:21:52) ERROR: No test targets were found, yet testing was requested
``` 
 Culprit: https://github.com/bazelbuild/bazel/issues/8195

 Steps:
  ``` 
 git clone https://github.com/bazelbuild/rules_apple
git reset 2e45cefc23c1d7744d11be8fa8ba54d48b47465a  --hard
export USE_BAZEL_VERSION=788b6080f54c6ca5093526023dfd9b12b90403f8
bazel test doc/... 
 ``` 
 CC Greenteam @fweikert

sgowroji:
We are currently closing this issue due to this PR https://github.com/bazelbuild/continuous-integration/pull/2101 and now they are tested in bcr-bazel-compatibility-test CI.

PR #2556, opened 2024-10-10
Update bazelbuild/stardoc to 0.7.1

mattrobmattrob:
Fixes https://github.com/bazelbuild/rules_apple/issues/2555 by pulling in https://github.com/bazelbuild/stardoc/pull/238.

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -17,7 +17,7 @@ bazel_dep(
 
 bazel_dep(
     name = "stardoc",
-    version = "0.6.2",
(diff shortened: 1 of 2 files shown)


luispadron (approved):

mattrobmattrob:
Hmm, stardoc 0.7.0 doesn't support Bazel 6. Will need to come up with a different solution.

brentleyjones:
Docs should only be run on HEAD, so that shouldn't block us.

Issue #2558, opened 2024-10-18
I need to apply a transition on a target defined by ios_extension rule


vakhidbetrakhmadov:
Hello,

I need to apply a transition on a target defined by `ios_extension` rule. 

I have the following setup: 
```
App -> Extension 
App -> Module_A -> Module_B -> Module_C -> ... -> Module_Z
Extension -> Module_B -> Module_C 
```
- Modules are statically linked into top level targets: `App`, `Extension`.
- `Module_C` is the largest module.

In order to reduce `App` size, i would like to flavour `Module_C` for `Extension`.

In order flavour `Module_C` for `Extension` i need to propagate a custom build setting along the `Extension -> ...` dependency tree.

I have done some research and it looks like that at the moment there is no supported way of applying a transition on a target defined by a third party rule.

The most commonly proposed workaround that i found is to create a wrapper rule around the third party rule.
The wrapper rule can apply required transitions on the wrapped rule and propagate its providers.

I have pretty much done this with `ios_extension`, but there is an issue that i am facing.

The issue is that not all `ios_extension`'s providers are accessible from outside of `rules_apple`, and without for example `AppleDebugInfo` no dSYMs for `Extension` are generated, or without `_AppleSymbolsFileInfo` no `.symbols` are generated for `Extension`. 

I didn't find any difference in output with/without `_AppleCodesigningDossierInfo`, `AppleExtensionSafeValidationInfo`, `_AppleSwiftDylibsInfo`.

```
load("//Bazel/flavour:flavour_transition.bzl", "flavour_transition")
load(
    "@build_bazel_rules_apple//apple:providers.bzl",
    "AppleBundleInfo",
    "AppleCodesigningDossierInfo",
    "AppleDsymBundleInfo",
    "AppleResourceInfo",
    "IosExtensionBundleInfo",
)
load(
    "@build_bazel_rules_apple//apple/internal/providers:embeddable_info.bzl",
    "AppleEmbeddableInfo",
)

def _flavoured_ios_extension_impl(ctx):
    ios_extension = ctx.attr.ios_extension[0]

    # The list of providers extracted from the `print(ios_extension)` output (unaccessible providers are commented out).
    ios_extension_providers = [
        DefaultInfo,
        apple_common.AppleExecutableBinary,
        IosExtensionBundleInfo,
        apple_common.AppleDebugOutputs,
        AppleBundleInfo,
        # _AppleCodesigningDossierInfo,
[...]

PR #2559, opened 2024-10-18
Make AppleDebugInfo and AppleSymbolsFileInfo available from outside of rules_apple

vakhidbetrakhmadov:
Resolves: #2558


vakhidbetrakhmadov:
I am not sure why "Doc tests" job is red, i have run `bazel test doc/...` and `bazelisk run //doc:update` locally and they both were successful.

PR #2563, opened 2024-10-21
Update docs and bzl_library usage for Bazel 8

luispadron:
This should fix the `last_green` Bazel doc tests

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -9,6 +9,7 @@ module(
 bazel_dep(name = "apple_support", version = "1.15.1", repo_name = "build_bazel_apple_support")
 bazel_dep(name = "bazel_skylib", version = "1.3.0")
 bazel_dep(name = "platforms", version = "0.0.9")
+bazel_dep(name = "rules_cc", version = "0.0.2")
 bazel_dep(
     name = "rules_swift",
     version = "2.1.1",
@@ -20,6 +21,12 @@ bazel_dep(
     version = "0.6.2",
     dev_dependency = True,
     repo_name = "io_bazel_stardoc",
+)
+bazel_dep(
+    name = "protobuf",
+    version = "21.7",
+    dev_dependency = True,
+    repo_name = "com_google_protobuf",
 )
 
 non_module_deps = use_extension("//apple:extensions.bzl", "non_module_deps")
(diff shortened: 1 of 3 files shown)


MODULE.bazel
```diff
@@ -9,6 +9,7 @@ module(
 bazel_dep(name = "apple_support", version = "1.15.1", repo_name = "build_bazel_apple_support")
 bazel_dep(name = "bazel_skylib", version = "1.3.0")
 bazel_dep(name = "platforms", version = "0.0.9")
+bazel_dep(name = "rules_cc", version = "0.0.10")
```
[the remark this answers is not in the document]
mattrobmattrob: Nevermind, saw the actual usages of it. Assumed it was for `stardoc`.

MODULE.bazel
```diff
@@ -17,10 +18,16 @@ bazel_dep(
 
 bazel_dep(
     name = "stardoc",
-    version = "0.6.2",
+    version = "0.7.1",
```
brentleyjones, line 21: Can we get away without upgrading this? I want to save the change to the docs where it uses `@rules_apple` for when we drop Bazel. I was thinking we could add the flag to ignore the empty globs to the Doc job instead for now.

luispadron: Unfortunately not:

```
ERROR: Traceback (most recent call last):
        File "/private/var/tmp/_bazel_lpadron/d8d9324fa7fd2d11d5178d8cf2834c55/external/stardoc+/stardoc/BUILD", line 72, column 16, in <toplevel>
                srcs = glob(["*.jar"]),
```

Which was fixed in 0.7.1 it seems

luispadron: Oh i see:

>  I was thinking we could add the flag to ignore the empty globs to the Doc job instead for now.

Yeah lets do that for stardoc

luispadron: > I want to save the change to the docs where it uses @rules_apple for when we drop Bazel.

Even with stardoc 0.6.2 using last_green Bazel the `load("@rules_apple")` stuff is added to all docs.

brentleyjones: Hmm. I only saw that when using the newer stardoc.

brentleyjones: `protobuf` has a non-dev dependency on `stardoc` 🙃

luispadron: Fixed the stardoc upgrade issue, it was being brought in by protobuf:

```starlark
# TODO: Remove override when a protobuf release that marks `stardoc` as a
# dev_dependency is available, until then it's upgrading our stardoc version
# so override it here.
single_version_override(
    module_name = "stardoc",
    version = "0.6.2",
)
```

luispadron:
Theres other incompatibilities here though with Bazel 6, protobuf override (to fix stardoc) breaks Bazel 6:

```
ERROR: Bazel version 6.5.0 is not compatible with module "rules_java@7.6.1" (bazel_compatibility: [>=7.0.0])
ERROR: Error computing the main repository mapping: Bazel compatibility check failed
```

brentleyjones (approved):

luispadron:
Issue was rules_cc 0.0.10 is not compatible with Bazel 6, downgraded to 0.0.2 to match rules_swift

PR #2564, opened 2024-10-23
Remove unnecessary `if` in `apple_precompiled_resource_bundle`

brentleyjones:


luispadron (approved):
Just double checking, the code is different one sets `parent_dir`, do we just want it to always be set?

brentleyjones:
`bundle_name` is always set, so we always went down the one path.

PR #2570, opened 2024-10-25
Add the `strip_structured_resources_prefixes` attribute where `structured_resources` exists

brentleyjones:
This allows removing intermediate directories from structured resource paths, without having to do wasteful file copies.


aaronsky (approved):
worked for our use-case

PR #2583, opened 2024-11-05
Fix `docc_archive` preview without docc bundle

### LINKED ISSUE #2582: `bazel run` fails for docc_archive targets with a swift_library dep which does not have a docc bundle
This can be reproduced with the HelloWorldSwift example in the repository by removing the .docc from data.

luispadron:
Fixes #2582

`docc preview` supports previewing with or without a `.docc` bundle. There were some issues in the argument handling of the template script that stopped this from working when no bundle was provided. This fixes it and additionally fixes the following warning:

```
warning: '--index' is deprecated. Use '--emit-lmdb-index' indead.
```

`docc preview` will emit a json representation by default and we do not need the `lmdb` index afaict.

### DIFF
--- a/apple/internal/aspects/docc_archive_aspect.bzl
+++ b/apple/internal/aspects/docc_archive_aspect.bzl
@@ -76,7 +76,7 @@ def _docc_symbol_graphs_aspect_impl(target, ctx):
     if not symbol_graphs:
         return []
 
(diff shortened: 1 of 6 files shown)


AttilaTheFun:
I tested this and it's working as expected with your branch. Thanks!

PR #2590, opened 2024-11-08
Fix `ios_xctestrun_runner` not updating binary on delta updates

brentleyjones:


mattrobmattrob (approved):
Oh my.

brentleyjones:
It might only apply to UI tests, but I did it for all because why not.

luispadron (approved):

jpsim:
Nice but slightly sneaky fix. I had run into this as well. Thanks!

PR #2595, opened 2024-11-13
Fix docc archive bugs

AttilaTheFun:
This PR builds upon and fixes @luispadron 's draft PR:
https://github.com/bazelbuild/rules_apple/pull/2583

Which fixes https://github.com/bazelbuild/rules_apple/issues/2582

docc preview supports previewing with or without a .docc bundle. There were some issues in the argument handling of the template script that stopped this from working when no bundle was provided. This fixes it and additionally fixes the following warning:

warning: '--index' is deprecated. Use '--emit-lmdb-index' indead.
 docc preview will emit a json representation by default and we do not need the lmdb index afaict.

Additionally, previously the files within a .docc directory were not included as inputs to the action, so it would not be properly invalidated if you changed the directory contents. 

This PR includes all of the files within the data attribute under the .docc directory.

### DIFF
--- a/apple/internal/docc.bzl
+++ b/apple/internal/docc.bzl
@@ -138,7 +138,7 @@ def _docc_archive_impl(ctx):
             "{fallback_display_name}": fallback_display_name,
             "{platform}": platform.name_in_plist,
             "{sdk_version}": str(xcode_config.sdk_version_for_platform(platform)),
-            "{symbol_graph_dirs}": " ".join([f.short_path for f in symbol_graphs]) if symbol_graphs else "",
+            "{symbol_graph_dirs}": " ".join([f.path for f in symbol_graphs]) if symbol_graphs else "",
             "{target_name}": ctx.attr.name,
             "{xcode_version}": str(xcode_config.xcode_version()),
         },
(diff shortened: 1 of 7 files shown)


test/docc_archive_test.sh
luispadron, line 1: If you remove this test CI should pass. I can follow up on seeing what it takes to fix, some issue with this test runner change the run path.

luispadron (approved):
Thank you!

apple/internal/docc.bzl
```diff
@@ -258,6 +262,10 @@ This value must be either `fileprivate`, `internal`, `private`, or `public`. The
                 allow_single_file = True,
                 default = "//apple/internal/templates:docc_preview_template",
             ),
+            "_bash_runfiles": attr.label(
+                allow_files = True,
+                default = "@bazel_tools//tools/bash/runfiles",
+            ),
```
luispadron, line 268:
```suggestion
```

Leftover from my PR when testing but I don't think it's needed

test/docc_archive_test.sh
AttilaTheFun: Ok! Yeah it was timing out after five minutes for me.

PR #2598, opened 2024-11-13
Recover from simulator already booted errors

### LINKED ISSUE #2597: Simulator reuse can run into concurrent booting issues
As fortold in https://github.com/bazelbuild/rules_apple/pull/1322#issuecomment-1017117150.

```
Existing simulator 'BAZEL_TEST_iPhone 14 Pro Max_18.0' (3F1B7631-5845-4D13-8F5D-20FD49F0A6E6) state is: shutdown
Device boot failed
An error was encountered processing the command (domain=com.apple.CoreSimulator.SimError, code=405):
Unable to boot device in current state: Booted
```

I think this should be a simple fix, where we detect this and try again (which should work).

brentleyjones:
Fixes #2597.

### DIFF
--- a/apple/testing/default_runner/simulator_creator.py
+++ b/apple/testing/default_runner/simulator_creator.py
@@ -34,17 +34,48 @@ def _boot_simulator(simulator_id: str) -> None:
         output = _simctl(["bootstatus", simulator_id, "-b"])
         print(output, file=sys.stderr)
     except subprocess.CalledProcessError as e:
+        exit_code = e.returncode
+
+        # When reusing simulators we may encounter the error:
+        # 'Unable to boot device in current state: Booted'.
+        #
+        # This is because the simulator is already booted, and we can ignore it
+        # if we check and the simulator is in fact booted.
+        if exit_code == 149:
+            devices = json.loads(
+                _simctl(["list", "devices", "-j", simulator_id]),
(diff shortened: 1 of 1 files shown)


apple/testing/default_runner/simulator_creator.py
```diff
@@ -34,17 +34,48 @@ def _boot_simulator(simulator_id: str) -> None:
...
+        exit_code = e.returncode
+
+        # When reusing simulators we may encounter the error:
+        # 'Unable to boot device in current state: Booted'.
+        #
+        # This is because the simulator is already booted, and we can ignore it
+        # if we check and the simulator is in fact booted.
+        if exit_code == 149:
+            devices = json.loads(
+                _simctl(["list", "devices", "-j", simulator_id]),
```
mattrobmattrob, line 46: Should this call to `simctl` be wrapped in a `try/except` block as well?

mattrobmattrob (approved):

apple/testing/default_runner/simulator_creator.py
brentleyjones: If this one fails I don't want to catch it. The `except` is to try to recover from specific error codes during booting, but this call is just getting the list of devices.

luispadron (approved):

PR #2603, opened 2024-11-22
Make `AppleExtensionSafeValidationInfo` a public provider

brentleyjones:
This allows for custom `*_framework` rules to work with rules_apple.


brentleyjones:
Ehh, not possible because other providers will need to be exposed.

brentleyjones:
We can probably make those public as well, but I'm not doing that right now.

PR #2605, opened 2024-11-26
Add `apple.codesign_frameworks_without_provisioning_profile` feature

brentleyjones:
This feature causes `*_{dynamic_,}framework` targets to codesign themselves even when they don’t have a provisioning profile set. This allows moving and caching some of the codesigning time into the framework targets, which can be a big win if your incremental builds don’t touch a lot of the frameworks.

### DIFF
--- a/apple/internal/ios_rules.bzl
+++ b/apple/internal/ios_rules.bzl
@@ -908,10 +908,12 @@ def _ios_framework_impl(ctx):
     provisioning_profile = ctx.file.provisioning_profile
     resource_deps = ctx.attr.deps + ctx.attr.resources
     signed_frameworks = []
-    if provisioning_profile:
-        signed_frameworks = [
-            bundle_name + rule_descriptor.bundle_extension,
-        ]
+    if codesigning_support.should_sign_bundles(
+        provisioning_profile = provisioning_profile,
+        rule_descriptor = rule_descriptor,
+        features = features,
+    ):
+        signed_frameworks = [bundle_name + bundle_extension]
     top_level_infoplists = resources.collect(
         attr = ctx.attr,
         res_attrs = ["infoplists"],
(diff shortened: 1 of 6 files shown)


apple/internal/ios_rules.bzl
```diff
@@ -1489,10 +1491,12 @@ def _ios_dynamic_framework_impl(ctx):
     )
 
     signed_frameworks = []
-    if getattr(ctx.file, "provisioning_profile", None):
-        signed_frameworks = [
-            bundle_name + rule_descriptor.bundle_extension,
-        ]
+    if codesigning_support.should_sign_bundles(
+        provisioning_profile = ctx.file.provisioning_profile,
```
mattrobmattrob, line 1495: The move away from `getattr(ctx.file, "provisioning_profile", None)` is acceptable here, correct?
```suggestion
        provisioning_profile = getattr(ctx.file, "provisioning_profile", None),
```

apple/internal/macos_rules.bzl
```diff
@@ -3083,10 +3085,12 @@ def _macos_dynamic_framework_impl(ctx):
     )
 
     signed_frameworks = []
-    if getattr(ctx.file, "provisioning_profile", None):
-        signed_frameworks = [
-            bundle_name + rule_descriptor.bundle_extension,
-        ]
+    if codesigning_support.should_sign_bundles(
+        provisioning_profile = ctx.file.provisioning_profile,
```
mattrobmattrob, line 3089: And same here and a few other places that I won't comment on.

mattrobmattrob (approved):

apple/internal/ios_rules.bzl
brentleyjones: I'm not sure why it wasn't a direct reference to begin with. Was it ever optional/not there?

mattrobmattrob: No clue. At least it's unified and not a smattering of getattr and direct access. I like it!

PR #2618, opened 2024-12-09
Add Label() to archive_contents_test marco

erikkerber:
Since this macro is not specific to any test or target under test in rule_apple, Label() helps resolve the script form externally defined targets and tests.

There are many more uses of the `apple_verification_test` rule, but they are specific internally and I chose not to add across the board.


mattrobmattrob (approved):

keith:
Thanks!

PR #2623, opened 2024-12-13
WIP: make llvm-cov export err output non-blocking, if the exit code was still 0

aaronsky:
We may want more customization here (since this could obscure real
issues), or better ordering of the error log
and the stderr output from llvm-cov.

Context: https://bazelbuild.slack.com/archives/CD3QY5C2X/p1734100078982459


aaronsky:
note: keith said this was a bad idea, so i'm looking further into root causes

PR #2628, opened 2025-01-10
Remove `emit_swiftinterface` from the Apple rule transitions.

AttilaTheFun:
Instead of relying on the transition to set the flag throughout the dependency graph, users should set `library_evolution = True` on the specific libraries that they want to support library evolution.

PiperOrigin-RevId: 629760009
(cherry picked from commit 24a862d0f9eda59fe9c0809f27b029f077ce9138)

### DIFF
--- a/apple/internal/templates/apple_device.template.py
+++ b/apple/internal/templates/apple_device.template.py
@@ -496,7 +496,9 @@ def run_app(
     ]
     launch_args = ensure_devicectl_terminate_existing(launch_args)
     # Append optional launch arguments.
-    app_args = [app_bundle_id] + sys.argv[1:]
+    # devicectl keeps parsing dash-prefixed values as command options even
+    # after the bundle identifier. Terminate option parsing before app args.
+    app_args = ["--", app_bundle_id] + sys.argv[1:]
     launch_app(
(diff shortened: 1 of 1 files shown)


luispadron (approved):

brentleyjones (approved):

AttilaTheFun:
@brentleyjones You already cherry-picked the removal of the build config from rules_swift, so the head of rules_apple is now incompatible with the head of rules_swift without this change.

We either need to revert the change in rules_swift, or apply this change to rules_apple from upstream.

brentleyjones:
Sure, we need this, but I'm wondering if it means a 4.0 version bump.

AttilaTheFun:
How do you view what counts as a breaking change?

If we cut a new version of rules_apple for this, I'd argue we'd need a new version of rules_swift, too. It was actually that breakage that necessitated this one.

brentleyjones:
> How do you view what counts as a breaking change?

If people upgrade to a version of **rules_apple** that have this change and it breaks their builds.

aaronsky:
@AttilaTheFun looks like there are some test failures that need to be fixed before we can land this

PR #2629, opened 2025-01-12
Escape backslashes in example build label regex

pswaminathan:
I'm not sure if this worked in an earlier version of Bazel, but in Bazel 7, these get interpreted as escape sequences for Starlark strings (and thus a failure for invalid escape sequence), rather than being passed to the underlying tool as regex sequences. Escaping with two backslashes makes it work correctly.


luispadron (approved):

aaronsky (approved):

apple/versioning.bzl
```diff
@@ -254,8 +254,8 @@ apple_bundle_version(
     build_label_pattern = "MyApp_{version}_build_{build}",
     build_version = "{version}.{build}",
     capture_groups = {
-        "version": "\\d+\\.\\d+",
-        "build": "\\d+",
+        "version": "\\\\d+\\\\.\\\\d+",
+        "build": "\\\\d+",
```
pswaminathan, line 258: This change isn't really useful for reading the .bzl file, but because this gets used to generate the doc, *each* backslash needs to be double-escaped.

PR #2630, opened 2025-01-14
Support bundling live activities extensions with ios_app_clip

adincebic:
Per [Apple's documentation](https://developer.apple.com/documentation/appclip/offering-live-activities-with-your-app-clip) we are allowed to bundle extensions with app clips but only live activities.

This pull request adds support for that.

- [ ] Checks for live activity so that no one tries to bundle standard iOS extension. These are not worth persuing as vanilla Xcode does not do it. The verification happens at the level of entitlements, meaning that correct entitlement must be present to be able to submit an app clip with live activity extension to the AppStore.
- [x] Documentation.
- [x] Tests


luispadron (approved):
Thanks!

PR #2634, opened 2025-01-17
Add support for `.provisionprofile` in `local_provisioning_profile`

luispadron:
macOS profiles have the `.provisionprofile` extension so `local_provisioning_profile` would always fail to find any macOS provisioning profiles. I added a new `profile_extension` attr to the rule to support specifying the profile is for macOS as it's the only platform that seems to use this extension I kept the `.mobileprovision` extension as the default.


mattrobmattrob (approved):

aaronsky (commented):
This looks good, but is it possible to write a test for this?

luispadron:
> This looks good, but is it possible to write a test for this?

I looked for existing ones and didn't see any, probably because it would be tricky to setup in a public repo / CI where profiles would need to be installed.

aaronsky (approved):

PR #2655, opened 2025-02-19
Add `manual` tag to test bundle rule

brentleyjones:
Fixes an issue related to toolchain selection when doing `//...` multi-platform builds.


keith (approved):
seems fine but maybe the bundle rule should get the macOS specific platform stuff too?

brentleyjones:
Yeah, it can as a separate change.

Issue #2397, opened 2024-02-08
Support On-Demand Resources


qyang-nj:
It seems `--enable-on-demand-resources YES` is not used anywhere in [actool.bzl](https://github.com/bazelbuild/rules_apple/blob/master/apple/internal/resource_actions/actool.bzl), so I believe [on-demand resources](https://developer.apple.com/library/archive/documentation/FileManagement/Conceptual/On_Demand_Resources_Guide/index.html#//apple_ref/doc/uid/TP40015083-CH2-SW1) are not supported currently.

It's probably non trivial to add the support, as it also requires generating a AssetPackOutputSpecifications.plist for `--asset-pack-output-specifications` argument.

PR #2658, opened 2025-02-27
Always pass '--enable-on-demand-resources YES' to actool

adincebic:
This needs to be set to be able to support ODR in rules_apple. Xcode also always passes it. See [Apple docs](https://developer.apple.com/library/archive/documentation/FileManagement/Conceptual/On_Demand_Resources_Guide/).

### DIFF
--- a/apple/internal/resource_actions/actool.bzl
+++ b/apple/internal/resource_actions/actool.bzl
@@ -284,6 +284,8 @@ def compile_asset_catalog(
         "--minimum-deployment-target",
         platform_prerequisites.minimum_os,
         "--compress-pngs",
+        "--enable-on-demand-resources",
+        "YES",
     ]
 
     args.extend(_actool_args_for_special_file_types(


qyang-nj:
I think supporting ODR requires more than just passing this option. I did file an issue a while ago, https://github.com/bazelbuild/rules_apple/issues/2397

adincebic:
> I think supporting ODR requires more than just passing this option. I did file an issue a while ago, #2397

Oh that's absolutely true. I have managed to get ODR working by post processing .ipa and plan to soon start working on actual proper integration into rules_apple. This is one of the many steps required.

luispadron:
Is there any downside to enabling this always?

adincebic:
> Is there any downside to enabling this always? 

Don't think so, Xcode passes it also. Let me compare the build with and without it and I'll confirm than.

adincebic:
I'll enable this once I find time to fully add support for ODR.

PR #2659, opened 2025-02-27
Cache more resource processing by default

brentleyjones:
Should remove the need for `apple_precompiled_resource_bundle`.


adincebic:
@brentleyjones what's stoping us from merging this?

PR #2662, opened 2025-03-11
Allow only keeping `.xcresults` for failed tests

brentleyjones:


mattrobmattrob (commented):
Looks like this changes the existing/default behavior, correct? Is that desired?

mattrobmattrob:
🤦 I scanned that as `-1` default value. Whoops.

mattrobmattrob (approved):

luispadron (approved):

PR #2669, opened 2025-03-26
Update swift_intent_library to check for srcs and data from kwargs

ssarad:
As per the docs 

```
It takes a single `.intentdefinition` file and creates a target that can be added as a dependency from `objc_library` or
`swift_library` targets.
It accepts the regular `swift_library` attributes too.
```

However, if `swift_intent_library` is passed srcs and data params, it errors out with 

```sh
Error in swift_library: rule(...) got multiple values for parameter 'srcs'
```

### DIFF
--- a/apple/resources.bzl
+++ b/apple/resources.bzl
@@ -202,6 +202,7 @@ def swift_intent_library(
         swift_version = None,
         testonly = False,
         **kwargs):
+    # buildifier: disable=function-docstring-args
     """
 This macro supports the integration of Intents `.intentdefinition` files into Apple rules.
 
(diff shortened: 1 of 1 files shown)


apple/resources.bzl
```diff
@@ -202,6 +202,7 @@ def swift_intent_library(
         swift_version = None,
         testonly = False,
         **kwargs):
+    # buildifier: disable=function-docstring-args
```
ssarad, line 205: Buildifier kept complaining about this, also noticed that other functions above have added the same disable command. Let me know if this is fine or you'd prefer we not do this.

ssarad (commented):
Left a comment about buildifier incase we don't want to do that.

Also, looks like some CI checks are failing that I don't think are because of my changes 🤔

PR #2675, opened 2025-03-29
wip: fix sdk->runtime mapping assumptions in ios test runners

aaronsky:
- bumped bazelci to Xcode 16.2
- bumped bazelci to Buildifier 7.3.1
- heavily refactored simulator_creator.py
  - discovered during the OTA rollout of the iOS 18.3.1 simulator runtime, our assumptions around sensible defaults for which runtime for a testrunner to use break down when the os version either doesn't match the SDK build version, or align with the underlying CoreSimulator identifier (such as if it contains a patch bit).
  - the new implementation leverages `xcrun simctl runtime match` to identify a default runtime if one is not specified by `ctx.attr.os_version` or `--ios_simulator_version`. This may result in breaking behavior for downstream users' CI if Xcode is not assigning this automatically.
  - the runtime match implementation is temporary until simctl ships with a built-in way to identify the sdk->runtime mappings, see FB17012461. an alternative to this approach could be to pull down the [index](https://devimages-cdn.apple.com/downloads/xcode/simulators/index2.dvtdownloadableindex) from an Apple CDN during test runs, but this could get expensive and/or be incompatible with airgapped test environments.
  - see also bazelbuild/continuous-integration#2189


apple/testing/default_runner/ios_xctestrun_runner.bzl
```diff
@@ -53,16 +53,14 @@ def _get_execution_environment(ctx):
     return {"XCODE_VERSION_OVERRIDE": xcode_version}
 
 def _ios_xctestrun_runner_impl(ctx):
-    os_version = str(ctx.attr.os_version or ctx.fragments.objc.ios_simulator_version or
-                     ctx.attr._xcode_config[apple_common.XcodeProperties].default_ios_sdk_version)
-
     # TODO: Ideally we would be smarter about picking a device, but we don't know what the current version of Xcode supports
     device_type = ctx.attr.device_type or ctx.fragments.objc.ios_simulator_device or "iPhone 15"
+    os_version = str(ctx.attr.os_version or ctx.fragments.objc.ios_simulator_version or ctx.attr._xcode_config[apple_common.XcodeProperties].default_ios_sdk_version)
```
aaronsky, line 58: this default is still problematic

.bazelci/presubmit.yml
```diff
@@ -4,7 +4,13 @@ x_defaults:
   # it doesn't know about; so that is used to avoid repeating common subparts.
   common: &common
     platform: macos_arm64
-    xcode_version: "15.4"
+    xcode_version: "16.2"
+    # TODO(aaronsky): delete prior to landing
```
aaronsky, line 8: TODO: delete prior to landing

apple/testing/default_runner/simulator_creator.py
```diff
@@ -108,47 +128,76 @@ def _build_parser() -> argparse.ArgumentParser:
...
+    device_type: str,
+    os_version: str,
+    name: Optional[str],
+    reuse_simulator: bool,
+) -> None:
+    simctl_state = json.loads(_simctl("list", "-j"))
+    devices = simctl_state["devices"]
+    runtimes = simctl_state["runtimes"]
+    runtime = next(
+        (runtime for runtime in runtimes if runtime["version"].startswith(os_version)),
```
keith, line 142: should we also be considering `isAvailable` here?

apple/testing/default_runner/ios_test_runner.template.sh
```diff
@@ -221,39 +225,37 @@ if [[ -n "${LAUNCH_OPTIONS_JSON_STR}" ]]; then
 fi
 
 target_flags=()
-if [[ -n "${REUSE_GLOBAL_SIMULATOR:-}" ]]; then
```
keith, line 224: i don't follow the logic change here, but this is intended for users to pass so removing this functionality is probably bad?

keith (commented):
might be too annoying for this one now but in the future it would be easier to review changes like this if we could submit refactors in separate changes

PR #2679, opened 2025-03-31
Fix CI with Xcode 16.2

keith:
- CI VMs seem to be very resource constrained. `system_profiler SPHardwareDataType` shows 6 CPUs and 32gb of memory but I don't think that's true. This reduces the jobs and resources a lot. We can continue to play with these numbers.
- Asset catalogs seem to take forever, I've made these exclusive, we have no large catalogs so these _should_ all just be instant
- xctestrunner picks the wrong simulator device type (iPhone 6s Plus), I forced that to be something newer.
- New Xcode validates some old asset catalog issues, deleted the violations
- Temporarily pin back last_green just to get CI green


mattrobmattrob:
AFAIK, we're still going to have issues with the XRSimulator stuff: https://github.com/bazelbuild/continuous-integration/issues/2189#issuecomment-2764264032

keith:
yea looking at that too

apple/testing/default_runner/ios_xctestrun_runner.template.sh
```diff
@@ -384,6 +384,7 @@ fi
 
 readonly profraw="$test_tmp_dir/coverage.profraw"
 
+set -x
```
aaronsky, line 387: get rid of this prior to land please

aaronsky (approved):

apple/testing/default_runner/ios_xctestrun_runner.template.sh
keith: tyty

PR #2680, opened 2025-04-01
xctrunner: Fix provider of platform type

kapoorlakshya:
Was incorrectly using `AppleBundleInfo` instead of `_TestBundleInfo`.

### DIFF
--- a/apple/internal/xctrunner.bzl
+++ b/apple/internal/xctrunner.bzl
@@ -66,7 +66,7 @@ def _test_bundle_info_aspect_impl(target, ctx):
             ],
         )
         platform_types = [
-            dep[AppleBundleInfo].platform_type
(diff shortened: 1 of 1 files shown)


mattrobmattrob (approved):
In practice, when is this different? Remote execution of some sort?

mattrobmattrob:
Actually, wondering if the if block at the top of the mutated func should be altered too now.

kapoorlakshya:
@mattrobmattrob Hey! We started using `test_suite` which does not have `AppleBundleInfo` as a provider. `_TestBundleInfo` provides the required info in that situation.

mattrobmattrob (approved):
Good deal. Also just realized this was the new `xctrunner` stuff. 👍

brentleyjones (approved):

PR #2681, opened 2025-04-04
Add `exec_data` attribute to tests

brentleyjones:
This differs from `data` in that it uses the `exec` config for the files. This is useful for running host-based tools in UI tests.

We could add `data` to the test runners for the same effect, but then we would need different test runners per test, which isn’t ideal.


keith (approved):

brentleyjones:
I'm not going to merge for now, as a better practice might be to use something like `force_exec` on some targets and place those targets in `resources`, so the files are bundled up with the test.

Issue #2551, opened 2024-10-08
`com.apple.developer.pass-type-identifiers` entitlement is replaced by `<team-id>.<bundle-id>` no matter what


vakhidbetrakhmadov:
Hello,

I am migrating a project from Xcode to Bazel.

When comparing `.entitlements` in the signature of the `.app` bundle built with Xcode and with Bazel, i have noticed the following diff:
```
   "com.apple.developer.pass-type-identifiers" => [
-    0 => "FOOBARBAZ1.*"
+    0 => "FOOBARBAZ1.com.foo.bar"
   ]
```

I have prepared a branch where this can be reproduced.

Please checkout [repro pass-type-identifiers](https://github.com/vakhidbetrakhmadov/rules_apple/commit/fe2989418bb8520a268ccc8d9764b167c8979500), run `bazelisk test //test/starlark_tests:ios_application_entitlements_contents_device_test` and observe `test.log`
```
exec ${PAGER:-/usr/bin/less} "$0" || exit 1
Executing tests from //test/starlark_tests:ios_application_entitlements_contents_device_test
-----------------------------------------------------------------------------
ERROR: Expected 'Array { FOOBARBAZ1.com.google.example }' to match 'FOOBARBAZ1.*' for key com.apple.developer.pass-type-identifiers
Actual contents were:
Dict {
    application-identifier = FOOBARBAZ1.*
    test-an-entitlement = false
    get-task-allow = true
    com.apple.developer.pass-type-identifiers = Array {
        FOOBARBAZ1.com.google.example
    }
}
```

vakhidbetrakhmadov:
This seems to be the reason:
https://github.com/bazelbuild/rules_apple/blob/master/tools/plisttool/plisttool.py#L1121-L1125

vakhidbetrakhmadov:
Given the `com.apple.developer.pass-type-identifiers` entitlement description, current behaviour doesn't seem correct ? 🤔  

![Screenshot 2024-10-08 at 15 22 06](https://github.com/user-attachments/assets/c4e5387c-a5eb-4af3-8ff1-1b437cf951d4)

https://developer.apple.com/documentation/bundleresources/entitlements/com_apple_developer_pass-type-identifiers

mkoohang:
I'm also running into this issue. Can we take a look and see if we can merge the related PR?

PR #2683, opened 2025-04-09
Disable "PREFIX.*" -> "PREFIX.BUNDLE_ID" substitutions in entitlements

vakhidbetrakhmadov:
Fix: #2551


mkoohang:
Hey folks I'm also running into this issue. Can we take a look and see if we can get it merged?

PR #2686, opened 2025-04-14
Use `TEST_TMPDIR` where possible

brentleyjones:
`TEST_TMPDIR` is in the Bazel exec root. This is better than `/tmp` or `$TMPDIR` since it’s managed by Bazel and will be cleaned up with `bazel clean`. Also, when using remote execution, it allows the output files to stay within the defined output tree which the executor can clean up properly.


luispadron (approved):
Nice, we have a bunch of these dirs left over

brentleyjones:
I need CI green before I feel safe merging this. Works for us, but I still want more checks.

Issue #2687, opened 2025-04-15
Linker inputs are not deduplicated in Bazel 8, causing linker warnings around duplicate rpaths


aaronsky:
In Bazel 8, I'm observing a number of strange behavior around linkopts, potentially due to the change from propagating linker context in `apple_common.Objc` to `CcInfo`. In particular, I'm seeing the following:

- `SwiftToolchainInfo` provides [required linkopts](https://github.com/bazelbuild/rules_swift/blob/master/swift/toolchains/xcode_swift_toolchain.bzl#L183-L193) into each linking context that uses Swift
- Framework rules like `ios_framework` consider each item in its `deps` as providing its own linking context via its dep tree, and these are not deduplicated.
- Any binary rule that takes a `frameworks` attr copies the linking context of each framework in the list into its own linking context in-full, and these are not deduplicated.
- Linking contexts from libraries that are "avoided" in [`subtract_linking_contexts`](https://github.com/bazelbuild/rules_apple/blob/master/apple/internal/multi_arch_binary_support.bzl#L60) still have their user-link flags appended into the merged linking context. 

What this means in practice is, given a simple `ios_application` with an `ios_extension` and some shared code in an `ios_framework`:

- If the `ios_framework` has multiple `deps` that either are or have a dependency on a `swift_library`, the `SwiftToolchainInfo` flags will appear multiple times.
- Each app/extension/binary that uses a framework will have its frameworks' linker inputs duplicated verbatim, which may include repetitious `SwiftToolchainInfo` user-link flags.
- These `SwiftToolchainInfo` flags include `-Wl,-rpath,/usr/lib/swift`, which produce a warning under ld for duplicate rpaths when repeated. 

I've attached a repro for this on a branch [here](). You can see that when building for any Bazel 7 release, the params are not duplicated, and under Bazel 8, they will be repeated once each. This issue affects our ability to adopt Bazel 8, as we strictly enforce warnings-as-errors in our codebase and this ld warning fails under `-Wl,-fatal_warnings`.

keith:
i think it would be useful to link the java linker context subtraction logic to understand why it's different. my hunch is they avoided the entire linking context vs now ignoring just the libraries

aaronsky:
[...]

PR #2689, opened 2025-04-15
Deduplicate linker inputs in merged linking contexts

aaronsky:
* adds an example of an app with an extension that both link a shared framework consuming Swift code. repros #2687.
* introduces a fix wherein entire linker inputs are deduplicated upon subtracting and combining into one linking context. This primarily affects duplicate linker inputs from `SwiftToolchainInfo`, but could have other effects.
  * This has the potential to affect linker param order, which is meaningful. Reviewers may want to consider this a breaking change.
  
I would like any reviewers to test the patch out in their own repo before approving, using either Bazel 7 or 8, to help ensure minimal breakage.


apple/internal/multi_arch_binary_support.bzl
```diff
@@ -51,26 +53,39 @@ def subtract_linking_contexts(owner, linking_contexts, avoid_dep_linking_context
     libraries = []
     user_link_flags = []
     additional_inputs = []
+    non_library_inputs = []
     linkstamps = []
     avoid_library_set = _build_avoid_library_set(avoid_dep_linking_contexts)
+
     for linking_context in linking_contexts:
         for linker_input in linking_context.linker_inputs.to_list():
+            if not linker_input.libraries:
```
keith, line 62: im still not sure if this is nicer, or your version which just dedup's linker_inputs all together, which feels like it could lead to a lot less work in this loop / duplicate flags for non-avoided libraries

aaronsky: I can go either way. My version is more aggressive, but it makes intuitive sense to me

aaronsky: Per [this thread](https://bazelbuild.slack.com/archives/C04DFUBQGSU/p1744825677535229), I've put forward a patch using my more aggressive deduplication of whole linker inputs. I would like any reviewers to test the patch out in their own repo before approving, using either Bazel 7 or 8, to help ensure minimal breakage.

aaronsky:
I would feel better if CI was passing before landing this. As mentioned above, I need to do more testing of this version of the patch internally as well, once I'm back from vacation.

apple/internal/multi_arch_binary_support.bzl
```diff
@@ -44,23 +46,34 @@ def subtract_linking_contexts(owner, linking_contexts, avoid_dep_linking_context
     additional_inputs = []
     linkstamps = []
     avoid_library_set = _build_avoid_library_set(avoid_dep_linking_contexts)
+    linker_inputs_seen = dict()
```
luispadron, line 49: we could use `bazel_skylib` [sets](https://github.com/bazelbuild/bazel-skylib/blob/main/docs/new_sets_doc.md), not sure if thats better perf or not though

luispadron (approved):
This fixed our duplicate link warnings for all our top-level targets! It would be nice to be able to test the link flags for our test cases, but we can add that after

apple/internal/multi_arch_binary_support.bzl
aaronsky: an older implementation used them, but i decided to keep to the existing setup for ease of review. Bazel 8 and later versions of Bazel 7 also have a native [set](https://bazel.build/rules/lib/core/set) type now, which I’m looking forward to being able to use.

aaronsky: the underlying implementation of skylib sets are very similar to this, fwiw, so I’d be surprised if there was a performance advantage

luispadron: Sounds good!

keith (approved):

keith (approved):

PR #2692, opened 2025-04-18
remove `--nocheck_visibility` flag from .bazelrc

aaronsky:
this seemed fixed, and was easy to unravel. let me know if this is actually bad.


keith (approved):
huh i thought there was more than this, but cool!

PR #2694, opened 2025-04-19
wip: run presubmit CI in github actions

aaronsky:
experimenting to potentially move off of bazelci, due to various challenges. this is a 1:1 representation of the presubmit.yml as-is, meaning buildifier hasn't been updated from 6.4.0, xcode is still 16.2, and pins for doc_tests are still in place. it is **not** a 1:1 representation in terms of UX or behavior.

updating buildifier and moving the stardoc/doc_tests pins should be done in a separate PR, and potentially xcode 16.3 as well (this would also allow us to more easily test a matrix of xcodes, if desired).

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -28,6 +28,11 @@ bazel_dep(
     dev_dependency = True,
     repo_name = "com_google_protobuf",
 )
+bazel_dep(
+    name = "buildifier_prebuilt",
+    version = "7.1.2",
+    dev_dependency = True,
+)
 
 non_module_deps = use_extension("//apple:extensions.bzl", "non_module_deps")
 use_repo(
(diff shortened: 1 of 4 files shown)


MODULE.bazel
```diff
@@ -28,6 +28,11 @@ bazel_dep(
     dev_dependency = True,
     repo_name = "com_google_protobuf",
 )
+bazel_dep(
+    name = "buildifier_prebuilt",
+    version = "7.1.2",
```
aaronsky, line 33: explicitly chose this version to keep us at buildifier 6.4.0

aaronsky:
build failures are due to insufficient permissions for `install_name_tool` to work in the sandbox. i assume the runner's permissions (`runner:staff`) are insufficient, because this works for me locally (`sky:wheel`)

PR #2696, opened 2025-04-20
Fix simulators missing

keith:
Seems like this is actually destructive on CI so we're getting machines
that don't have them. I guess I was lucky in the original PR since it
ran a few dozen times without hitting this.


mattrobmattrob (commented):
We definitely noticed it being destructive as well when @aaronsky  did a similar thing at the beginning of all this. Thanks for hacking this into a working state!

PR #2697, opened 2025-04-21
Always link XCTestSwiftSupport into Swift test targets

jflan-dd:
Unit tests that only use Swift Testing would not link `XCTestSwiftSupport` which resulted in the tests not actually being run.

Adding `import XCTest` anywhere in the Swift target would work around the issue, but for purely Swift Testing based tests this shouldn't be required.

I copied this set of flags from https://github.com/swiftlang/swift-build/blob/f974e0dc3841d82df08ead210fb900f850b919ac/Sources/SWBUniversalPlatform/Specs/ProductTypes.xcspec#L293

### DIFF
--- a/apple/internal/testing/apple_test_bundle_support.bzl
+++ b/apple/internal/testing/apple_test_bundle_support.bzl
@@ -362,7 +362,7 @@ def _apple_test_bundle_impl(*, ctx, product_type):
         "-bundle",
     ]
 
-    if platform_prerequisites.uses_swift:
+    if any([_is_swift_target(dep) for dep in ctx.attr.deps]):
         extra_linkopts.extend([
             "-Xlinker",
             "-needed-lXCTestSwiftSupport",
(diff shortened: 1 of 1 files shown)


keith (approved):
thanks! please fix the CLA check

apple/internal/testing/apple_test_bundle_support.bzl
```diff
@@ -353,10 +353,22 @@ def _apple_test_bundle_impl(*, ctx, product_type):
...
         "-framework",
         "XCTest",
         "-bundle",
     ]
+
+    if platform_prerequisites.uses_swift:
+        extra_linkopts.extend([
+            "-Xlinker",
+            "-needed-lXCTestSwiftSupport",
+            "-lXCTestSwiftSupport",
```
brentleyjones, line 369: Do we need an Xcode version check? Or is this valid for our minimum supported Xcode versions?

luispadron (approved):

PR #2698, opened 2025-04-22
Adds ios_dylib rule

chiragramani:
This PR introduces support for ios_dylib, which closely follows the implementation of macos_dylib to ensure consistency across platforms.

One of our projectgen use-cases involves combining all Swift libraries into a single dynamic library. This was previously handled outside of Bazel, but with this addition, we now leverage Bazel to support the same use-case in a reproducible and maintainable way.


adincebic:
Wasn't ios_framework producing dynamic library already or am I not understanding the difference here? @chiragramani

apple/internal/ios_rules.bzl
```diff
@@ -2881,6 +3002,44 @@ use only extension-safe APIs.
     ],
 )
 
+ios_dylib = rule_factory.create_apple_rule(
+    doc = "Builds a iOS Dylib binary.",
```
luispadron, line 3006: Could we add more info here, primarily what is a good use case for this rule, etc. This isnt intended to be used in `deps` of other targets right?

PR #2707, opened 2025-05-06
Infer tvOS/watchOS archs from the CPU parameters, similarly to iOS

rockbruno:
iOS currently has some special logic in `transition_support.bzl` that defaults the architecture to the value of `//command_line_option:cpu` if the value is iOS related (or a arm64 mac), but the same was not being done for other platforms. This was causing some confusion on our team as although on one end the regular iOS builds were being automatically resolved to arm64 simulator builds, on the other end everything else was resolving to x86_64 despite no changes in the environment.

This PR adds this same special logic to tvOS/watchOS for make them consistent with iOS. So in my case, building tvOS from a arm64 mac would now result in a tvOS arm64 simulator build.

It's unclear to me if the iOS case was done just to support some legacy use-case (meaning we should probably not do this for the other platforms), so if that's the case, let me know how to best solve this!

### DIFF
--- a/apple/internal/transition_support.bzl
+++ b/apple/internal/transition_support.bzl
@@ -163,6 +163,11 @@ def _cpu_string(*, environment_arch, platform_type, settings = {}):
         tvos_cpus = settings["//command_line_option:tvos_cpus"]
         if tvos_cpus:
             return "tvos_{}".format(tvos_cpus[0])
+        cpu_value = settings["//command_line_option:cpu"]
+        if cpu_value.startswith("tvos_"):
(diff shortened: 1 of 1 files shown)


keith:
I don't feel strongly about this, but you're right that it was because iOS existed before the --cpu type options were no longer recommended, so the google folks didn't implement this for the other platforms so that folks didn't start relying on it. IMO it's so low impact it would be fine to land

rockbruno:
@brentleyjones @luispadron Any thoughts on this?

PR #2708, opened 2025-05-09
Fix permissions when openrsync is used

mjburghard:
Fix file permissions when `openrsync` is used on macOS 14.5+.
### What was changed
Apply write permission using `find` & `chmod` when running on macOS 14.5+. To mitigate any previous `bazel run` invocations that already left the temporary directory without write permissions, an additional check and call to fix the permissions is added before calling `rsync`. This is needed when `rsync` needs to delete or modify any existing files.

### Why was this changed
macOS 15.4.1 replaced the rsync implementation with openrsync. Apple seems to have added some additional flags like `--chmod` to openrsync to match the functionality and interface of rsync. Unfortunately, not there are bugs like `--chmod` being not properly applied when syncing directories.

### How to test
- macOS 15.4+ is required
- Run `bazelisk run //examples/ios/HelloWorld --define=apple.experimental.tree_artifact_outputs=1`. `simctl` should fail with a permission error.  (If this steps succeeds, delete the temporary directory and uninstall HelloWorld from the iOS simulator).
- Switch to this PR and rerun the command. The command should succeed.


apple/internal/templates/apple_device.template.py
```diff
@@ -300,12 +300,22 @@ def extracted_app(
...
+    if (os.path.exists(dst_dir_application)
+        and not os.access(dst_dir_application, os.W_OK)):
+        fix_permissions(dst_dir, file_mode)
     rsync_command = [
         "/usr/bin/rsync",
         "--archive",
         "--delete",
         "--checksum",
-        "--chmod=u+w",
+        "--perms",
```
mjburghard, line 317: Adding `--perms` since the man page says it is required when passing `--chmod`.
Does not have any observable effect at the moment.

PR #2713, opened 2025-05-16
Tests: Allow post_action to determine exit status

kapoorlakshya:
Add new attr `post_action_determines_exit_code` to let the test runner post action determine the exit status. Our use case is allow soft failing of certain test targets.

```starlark
ios_xctestrun_runner(
    name = "ios_sim_runner_with_soft_fail",
    device_type = "iPhone Xs",
    post_action = ":post_action_soft_fail",
    post_action_determines_exit_code = True,
)
```


apple/testing/default_runner/ios_xctestrun_runner.template.sh
```diff
@@ -551,7 +551,7 @@ post_action_binary=%(post_action_binary)s
 TEST_EXIT_CODE=$test_exit_code \
   TEST_LOG_FILE="$testlog" \
   SIMULATOR_UDID="$simulator_id" \
-  "$post_action_binary"
+  "$post_action_binary" || post_action_exit_code=$?
```
brentleyjones, line 554: Please set `post_action_exit_code=0` just before this.

apple/testing/default_runner/ios_xctestrun_runner.template.sh
```diff
@@ -577,6 +577,10 @@ if [[ "${COLLECT_PROFDATA:-0}" == "1" && -f "$profdata" ]]; then
 fi
 
 if [[ "$test_exit_code" -ne 0 ]]; then
+  if [[ "${POST_ACTION_DETERMINES_EXIT_CODE:-0}" == "1" ]]; then
+    exit $post_action_exit_code
+  fi
+
   echo "error: tests exited with '$test_exit_code'" >&2
   exit "$test_exit_code"
 fi
```
brentleyjones, line 586:
```suggestion
if [[ "${POST_ACTION_DETERMINES_EXIT_CODE:-0}" == "1" ]]; then
  if [[ "$post_action_exit_code" -ne 0 ]]; then
    echo "error: post_action exited with '$post_action_exit_code'" >&2
    exit "$post_action_exit_code"
  fi
else
  if [[ "$test_exit_code" -ne 0 ]]; then
    echo "error: tests exited with '$test_exit_code'" >&2
    exit "$test_exit_code"
  fi
fi
```

apple/testing/default_runner/ios_xctestrun_runner.template.sh
```diff
@@ -551,7 +551,7 @@ post_action_binary=%(post_action_binary)s
 TEST_EXIT_CODE=$test_exit_code \
   TEST_LOG_FILE="$testlog" \
   SIMULATOR_UDID="$simulator_id" \
-  "$post_action_binary"
+  "$post_action_binary" || post_action_exit_code=$?
```
brentleyjones, line 555:
```suggestion
  "$post_action_binary" || post_action_exit_code=$?

if [[ "${POST_ACTION_DETERMINES_EXIT_CODE:-0}" != "1" ]]; then
  if [[ "$post_action_exit_code" -ne 0 ]]; then
    echo "error: post_action exited with '$post_action_exit_code'" >&2
    exit "$post_action_exit_code"
  fi
fi

```

brentleyjones (approved):

apple/testing/default_runner/ios_xctestrun_runner.bzl
```diff
@@ -191,6 +191,7 @@ A binary to run prior to test execution. Runs after simulator creation. Sets the
             cfg = "exec",
             doc = """
 A binary to run following test execution. Runs after testing but before test result handling and coverage processing. Sets the `$TEST_EXIT_CODE`, `$TEST_LOG_FILE`, and `$SIMULATOR_UDID` environment variables, in addition to any other variables available to the test runner.
+Add `POST_ACTION_DETERMINES_EXIT_CODE=1` to the test `env` to allow the post action to set the exit code of the test run. This is useful for tools that need to fail the test run based on their own criteria.
```
aaronsky, line 194: can we get this added to the other test runners that support post-actions, too?

aaronsky (changes requested):

kapoorlakshya:
> I like this but could we make this an actual attr of the test runner in addition to an env var? It's more discoverable that way

@brentleyjones and I discussed this earlier and the reason for this approach is to allow toggling this from a `.bazelrc`. Brentley, thoughts?

brentleyjones:
For us I guess it doesn't matter, since we have a macro and can apply it to every instance. The environment variable has the nice effect of (1) easily applying it globally if you want (2) automatically passed into `post_action` where we can check it if we want to.

apple/testing/default_runner/ios_xctestrun_runner.bzl
kapoorlakshya: Updated `ios_test_runner` and `macos_test_runner` as well.

aaronsky (approved):

PR #2714, opened 2025-05-17
wip: Fix compatibility with rules_swift 3.x

aaronsky:


apple/internal/framework_import_support.bzl
```diff
@@ -500,7 +498,7 @@ def _swift_info_from_module_interface(
         target_name = ctx.label.name,
     )
 
-    return swift_common.create_swift_info(
```
brentleyjones, line 503: Do we want to do a `hasattr` check? Or are there other things that will make this PR force a rules_swift minimum version?

Issue #2726, opened 2025-05-30
Add ability to concatenate Swift source files produced with apple_core_data_model


daltonclaybrook:
Today, `apple_core_data_model` produces a directory as its output file rather than individual Swift file outputs. This is because, under the hood, this rule invokes `xcrun momc --action generate`, which produces multiple Swift files from a single input file, and there's no way to know during the analysis phase what Swift files will be created.

Using a directory as an output rather than individual Swift files is problematic. For example, `swift_library` [does not support using a directory](https://github.com/bazelbuild/rules_swift/issues/969) in its `srcs` without a hacky workaround.

I'd like to propose a fix for this specific issue where, instead of only producing a directory of source files, `apple_core_data_model` provides an optional attribute `merge_outputs = True`. When set, the rule merges all Swift output files into a single Swift source file. This file can be known statically during analysis and can be included in the `srcs` list of a `swift_library` without the need for a hacky workaround.

daltonclaybrook:
Another possibly less invasive solution would be to allow the caller of `apple_core_data_model` to provide the static list of expected output files. When provided, the rule would simply return a depset of these files rather than the directory.

PR #2727, opened 2025-05-31
Ability to explicitly specify output files in apple_core_data_model rule

daltonclaybrook:
This PR contains one of the two fixes mentioned in #2726. This change makes it possible to explicitly specify the output files that are expected to be generated from the provided core data models. This causes the rule to return a provider containing these individual Swift files rather than just the containing folder.

### DIFF
--- a/apple/internal/resource_rules/apple_core_data_model.bzl
+++ b/apple/internal/resource_rules/apple_core_data_model.bzl
@@ -85,8 +85,13 @@ def _apple_core_data_model_impl(ctx):
         ["xcdatamodeld"],
         attr = "datamodels",
     )
+    output_file_groups = group_files_by_directory(
+        ctx.outputs.outs,
+        ["coredata.sources"],
+        attr = "sources",
+    )
 
-    output_files = []
+    output_files = list(ctx.outputs.outs)
     for datamodel_path, files in datamodel_groups.items():
         datamodel_name = paths.replace_extension(
             paths.basename(datamodel_path),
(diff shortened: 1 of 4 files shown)


luispadron (approved):

apple/internal/resource_rules/apple_core_data_model.bzl
```diff
@@ -127,6 +136,12 @@ apple_core_data_model = rule(
             "swift_version": attr.string(
                 doc = "Target Swift version for generated classes.",
             ),
+            "outs": attr.output_list(
+                doc = """
+An optional list of expected output files. If not empty, the rule will
+return these files individually rather than returning a directory.
```
luispadron, line 142: Could we add some info on how users can figure out these paths? Could we make them shorter/easier to declare somehow?

daltonclaybrook: Yeah, I can definitely improve the docs. Shorter/easier paths are a bit trickier. It's complicated by the fact that the rule can accept multiple `xcdatamodeld` files, and generated source files are grouped under a different directory for each data model. If the rule accepted only one data model, this would be a lot simpler.

Perhaps we could make something like this work where the `outs` attr is a dict instead of a list:
```
apple_core_data_model(
    name = "explicit_outs_data_model",
    srcs = [
        "//test/starlark_tests/resources:swift_datamodel",
    ],
    outs = {
        "swift_datamodel": [
            "Item+CoreDataProperties.swift",
            "swift_datamodel+CoreDataModel.swift",
        ],
    },
    tags = common.fixture_tags,
)
```

daltonclaybrook: I went ahead and updated the implementation based on this idea. I also updated the docs. Let me know what you think.

luispadron:
Looks good thanks for updating! Ill leave this here for another day or two to see if any other maintainers have thoughts then we can merge it

PR #2732, opened 2025-06-04
Fix issue with codesign not supporting bundles that have spaces

adincebic:
Cherry pick [259159545f16bfcef167b2be2d7b689b230e1c8c](https://github.com/bazelbuild/rules_apple/commit/259159545f16bfcef167b2be2d7b689b230e1c8c)


mattrobmattrob:
Meaning that the targets don't exist without cherry picking another change, correct?

adincebic:
> Meaning that the targets don't exist without cherry picking another change, correct?

Yes. I could manually add those targets though.

PR #2734, opened 2025-06-12
xctrunner - Use apple_support.run

kapoorlakshya:
Use `apple_support.run` to use the correct Xcode dirs when multiple versions are installed.

### DIFF
--- a/apple/internal/xctrunner.bzl
+++ b/apple/internal/xctrunner.bzl
@@ -18,6 +18,10 @@ platform and architectures as the given `tests` bundles.
 """
 
 load(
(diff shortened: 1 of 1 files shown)


apple/internal/xctrunner.bzl
```diff
@@ -143,7 +157,7 @@ def _xctrunner_impl(ctx):
 
 xctrunner = rule(
     implementation = _xctrunner_impl,
-    attrs = {
+    attrs = dicts.add(apple_support.action_required_attrs(), {
```
brentleyjones, line 160: We can/should use `|` instead of `dicts.add` now.

PR #2742, opened 2025-06-26
Collect framework dSYMs in test bundle debug dependencies

sewerynplazuk:
This PR fixes dSYM collection for frameworks specified in the `frameworks`  attribute of `ios_ui_test` targets. 

Currently, while framework binaries are properly made available to test targets, their corresponding dSYM files are not collected in the test bundle's debug symbol directory. This causes missing debug symbols when debugging test failures that involve framework code.

The fix ensures that dSYMs from framework dependencies are included alongside dSYMs from regular dependencies specified in the `deps` attribute.

### DIFF
--- a/test/starlark_tests/ios_ui_test_tests.bzl
+++ b/test/starlark_tests/ios_ui_test_tests.bzl
@@ -137,6 +137,33 @@ def ios_ui_test_test_suite(name):
         tags = [name],
     )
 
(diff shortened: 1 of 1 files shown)


brentleyjones (approved):

aaronsky:
is it possible to add a test for this so it doesn’t regress?

brentleyjones (approved):

PR #2745, opened 2025-07-04
Feat: Add bundle_extension attribute to apple_resource_bundle

SyedaAnshrahGillani:
### Summary
This pull request introduces a new `bundle_extension` attribute to the `apple_resource_bundle` rule, enabling customization of the file extension used for generated resource bundles.

---

#### New Attribute: `bundle_extension`
- **Purpose:**  
  Allows users to specify a custom file extension for the generated resource bundle.

- **Default Value:**  
  `.bundle` (ensures backward compatibility)

- **Motivation:**  
  The bundle extension was previously hardcoded to `.bundle`, limiting flexibility. This update allows:
  - Creating bundles with alternative extensions (e.g., `.framework`, `.xctest`)
  - Supporting specific use cases aligned with Xcode target types and Apple development practices

---

### Expanded Documentation (Per Reviewer Feedback)
The documentation for `bundle_extension` has been expanded to clarify usage and provide concrete examples related to Apple development tools and targets:

```python
"bundle_extension": attr.string(
    default = ".bundle",
    doc = """
The file extension for the resource bundle.

This attribute allows customization of the bundle's file extension. While `.bundle` is the 
default and most common extension for generic resource bundles, other extensions are used for 
specific purposes within Apple development.

For example:
* `.framework`: Used for frameworks, which are bundles containing both code and resources. 
  In Xcode, this corresponds to a "Framework" target.
* `.xctest`: Used for test bundles, which contain compiled test code and associated resources. 
  In Xcode, this corresponds to a "Unit Testing Bundle" or "UI Testing Bundle" target.

Using the correct extension ensures that the bundle is recognized and handled appropriately 
by Xcode and the Apple ecosystem.
""",
),

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# Apple Rules for [Bazel](https://bazel.build)
+# Apple Rules for [Bazel](https://bazel.build) 
 
 This repository contains rules for [Bazel](https://bazel.build) that can be
 used to bundle applications for Apple platforms.
--- a/apple/internal/resource_rules/apple_resource_bundle.bzl
+++ b/apple/internal/resource_rules/apple_resource_bundle.bzl
@@ -46,6 +46,20 @@ The desired name of the bundle (without the `.bundle` extension). If this attrib
 then the `name` of the target will be used instead.
 """,
         ),
+        "bundle_extension": attr.string(
+            default = ".bundle",
+            doc = """
+The file extension for the resource bundle.
+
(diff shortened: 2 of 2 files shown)


keith:
can you expand in the docs on when this might be used?

keith:
thanks for updating the docs, but in what case do you use this? is there a doc in apple's developer docs about when this is required? or what target in xcode does this?

also please sign the CLA

PR #2746, opened 2025-07-04
Fix bazelci on HEAD due to old dependency behavior

aaronsky:
- Bumped stardoc and removed overrides
- Bumped protobuf to a version without a transitive dependency on a version of aspect_bazel_lib that invoked --incompatible_use_toolchain_transition
- Bumped rules_cc to 0.1.2 to fix docs, corrected stardoc dep labels, updated docs

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -1,3 +1,5 @@
+"""rules_apple MODULE.bazel file"""
+
 module(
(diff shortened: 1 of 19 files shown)


aaronsky:
This is ready for review.

mattrobmattrob (approved):

PR #2751, opened 2025-07-11
use stardoc_with_diff_test from bazel-lib, remove stardoc from WORKSPACE

aaronsky:
trialing this to see if it works alright (it does at work, but I'm not sure how we feel about the new dep, even if it's a dev dep). if it works well, i'll try it out in rules_swift next.


aaronsky:
ah you're right, deleted in https://github.com/bazel-contrib/bazel-lib/pull/1188

i'll draft this and return to the drawing board. no need to take on a new dep if there's a better approach

PR #2759, opened 2025-07-31
fix: Relax exec_compatible_with requirement on dtrace rules.

mkosiba:
In #2661 the `exec_compatible_with` was added. This breaks some of our Linux-based tests where we use systemtap dtrace. I was hoping we could move the constraint onto the tool itself (that way we can supply one that's Linux-compatible and things keep working).`target_compatible_with` works for my usecase, but it's not exactly the same as the previous solution. @brentleyjones, could you have a look and let me know if this works for you?

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -36,3 +36,5 @@ use_repo(provisioning_profile_repository, "local_provisioning_profiles")
 
 apple_cc_configure = use_extension("@build_bazel_apple_support//crosstool:setup.bzl", "apple_cc_configure_extension")
 use_repo(apple_cc_configure, "local_config_apple_cc")
+
+register_toolchains("//apple:dtrace_toolchain")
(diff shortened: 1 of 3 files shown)


brentleyjones:
`target_compatible_with` doesn't influence toolchain selection, which is what is needed for `exec_compatible_with`. So some form of `exec_compatible_with` will be needed still.

mkosiba:
@brentleyjones, can you explain a bit more about toolchain selection for `dtrace_compile`? AFAICT the `dtrace` binary path is hardcoded (so not affected by toolchain resolution). Or is it that you have both macos and linux remote workers available and want to make sure this rule only runs on the macos workers?

brentleyjones:
Neat!

> I'm happy to toolchain-ify the other rules that have exec_compatible_with set if you'd like me to (I don't think there is benefit to this other than consistency).

Only a benefit if we are using a path that we want to override.

PR #2766, opened 2025-08-15
Remove clonefile() optimization when copying dSYMs

ma-oli:
dSYMs are currently copied using `cp -cp` on MacOS, which means MacOS will use clonefile(). While this optimizes speed and storage, this will also fail if src and dst are not on the same filesystem. This will typically happen if the sandbox and bazel-out are on different filesystems.


keith:
the manpage for cp says:

```
     -c    copy files using clonefile(2).  Note that if clonefile(2) is not
           supported for the target filesystem, then cp will fallback to using
           copyfile(2) instead to ensure the copy still succeeds.
```

is this not the case? it makes it sound like it shouldn't fail, but will fallback gracefully? i think in general we probably need to keep this optimization but figure out a fallback if it doesn't work. we do that in python like this https://github.com/bazelbuild/rules_apple/blob/bb6da2ccfcde089fff249d7771714f1851ce9e10/tools/bundletool/bundletool_experimental.py#L232-L235

ma-oli:
> is this not the case? it makes it sound like it shouldn't fail, but will fallback gracefully? i think in general we probably need to keep this optimization but figure out a fallback if it doesn't work. we do that in python like this

I see the confusion indeed; but looking at the clonefile() manual, it looks like `underlying filesystem does not support this call` and `src and dot are not on the same filesystem` are two different error codes. I suspect `cp -c` is lenient on the first one and not on the second one.

> i think in general we probably need to keep this optimization but figure out a fallback if it doesn't work
I'll update my patch

brentleyjones:
Here is how rules_xcodeproj does it:
```
if [[ $(stat -f '%d' "$src") == $(stat -f '%d' "$dest") ]]; then
  readonly cp_cmd="cp -c"
else
  readonly cp_cmd="cp"
fi
```

apple/internal/partials/debug_symbols.bzl
```diff
@@ -150,13 +150,16 @@ def _generate_dsym_binaries(
             mnemonic = "DsymDwarf",
             progress_message = "Copy DWARF into dSYM `%s`" % dsym_binary.short_path,
             command = """
-if [[ $OSTYPE == darwin* ]]; then
+st_dev_src=$(stat -f "%d" {src})
+st_dev_dst=$(stat -f "%d" $(dirname {dst}))
+# Only use `-c` if we're on Darwin and src and dst are on the same filesystem
+if [[ $OSTYPE == darwin* && ${{st_dev_src}} == ${{st_dev_dst}} ]]; then
```
brentleyjones, line 156: Perf wise, it would be nice to not call stat if we are on Linux.

PR #2769, opened 2025-08-21
Fix `bazel run` for applications when `devicectl` returns incomplete data

brentleyjones:


aaronsky (approved):
do we need to handle the case of `"unknown"` from these fields at all?

brentleyjones:
> do we need to handle the case of `"unknown"` from these fields at all?

Not sure. AFAICT other logic just ends up not picking this device over others. Maybe we need to do a but more, but this minimally unblocked me.

PR #2775, opened 2025-08-26
Include AppIntentsMetadataProcessor to mnemonic list for optimized remote cache and build execution

thelvis4:
The action key for `AppIntentsMetadataProcessor` is unstable because its `_app_intents_stub_binary` input is non-deterministic. This causes frequent remote cache misses, causing the action to run locally.


brentleyjones (approved):

brentleyjones:
But is it's outputs deterministic? Because this more cares about the outputs. The AC for this being unstable just results in a new AC entry (which is small), but if the outputs are deterministic, it's worth caching them, right?

PR #2786, opened 2025-09-12
Remove obsolete "xcrun simctl" cmds from presubmit

fweikert:
We've upgraded all CI workers from Xcode 16.2 to 16.4.


aaronsky:
thank you for your effort here! we believe we've fixed this in #2802

PR #2802, opened 2025-10-30
Fix CI by pinning to old versions temporarily

keith:
- Also fixes https://github.com/google/xctestrunner/issues/70 but for the other rule in this repo
- Adds xctestrunner patches which are needed for newer Xcode versions


keith:
notably https://github.com/bazelbuild/bazel/commit/013bd8f3834324e9409938082be577f586cfcc66 and https://github.com/bazelbuild/bazel/commit/0522f8e58bf864005207eb8f09162d35ddd1f678 caused breakages

PR #2803, opened 2025-10-30
Execute coverage logic (if enabled) prior to post action binary in xc…

maxwellE:
…testrunner

This allows consumers in post action binary to act upon coverage results (if they are enabled)

### DIFF
--- a/apple/testing/default_runner/ios_xctestrun_runner.template.sh
+++ b/apple/testing/default_runner/ios_xctestrun_runner.template.sh
@@ -553,8 +553,6 @@ else
     2>&1 | tee -i "$testlog" || test_exit_code=$?
 fi
 
-llvm_cov_status=0
-llvm_cov_json_export_status=0
 if [[ "${COVERAGE:-}" -eq 1 || "${APPLE_COVERAGE:-}" -eq 1 ]]; then
   profdata="$test_tmp_dir/$simulator_id/Coverage.profdata"
   if [[ "$should_use_xcodebuild" == false ]]; then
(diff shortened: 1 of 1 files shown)


apple/testing/default_runner/ios_xctestrun_runner.template.sh
```diff
@@ -553,6 +553,85 @@ else
...
+      -format text \
+      "${lcov_args[@]}" \
+      @"$llvm_coverage_manifest" \
+      > "$TEST_UNDECLARED_OUTPUTS_DIR/coverage.json" \
+      2> "$error_file" \
+      || llvm_cov_json_export_status=$?
+    if [[ -s "$error_file" || "$llvm_cov_json_export_status" -ne 0 ]]; then
+      echo "error: while exporting json coverage report" >&2
+      cat "$error_file" >&2
+      exit 1
```
keith: sgtm but note there is a difference here that the post action might not run if the test fails because of coverage collection where it would have run before

keith (approved):

apple/testing/default_runner/ios_xctestrun_runner.template.sh
aaronsky: yeah this is the main change i'm worried about for our purposes, but it is admittedly FUD

apple/testing/default_runner/ios_xctestrun_runner.template.sh
```diff
@@ -641,6 +640,8 @@ if [[ -n "${result_bundle_path:-}" ]]; then
     TEST_LOG_FILE="$testlog" \
     SIMULATOR_UDID="$simulator_id" \
     TEST_XCRESULT_BUNDLE_PATH="$result_bundle_path" \
+    LLVM_COV_STATUS="${llvm_cov_status:-0}" \
```
brentleyjones:
```suggestion
    LLVM_COV_EXIT_CODE="${llvm_cov_status:-0}" \
```

apple/testing/default_runner/ios_xctestrun_runner.template.sh
brentleyjones: I think we have to not `exit 1` here?= as well?

aaronsky (approved):

PR #2820, opened 2025-11-14
Simulator: Add support for preferring persistent sims

rockbruno:
This adds support for a `BAZEL_APPLE_PREFER_PERSISTENT_SIMS` env that allows ignoring the temporary simulator logic. In my experience they can be quite flaky, but my primary reasoning for this is to allow matching Xcode's behavior of always using neatly user-defined simulators. The change to support an explicit sim_identifier technically takes care of this but I think this is also good for those not using an extension to provide those.


erneestoc:
oh thanks! removes the need for my patch, did the exact same thing this weekend

PR #2839, opened 2025-12-07
Implement `apple_archive` rule to stop generating `.ipa` and `.zip` files from `{platform}_application` targets

adincebic:
We want to decouple creation of `.ipa` and `.zip` files from `{platform}_application` rules. This implies enabling tree artifacts for those application rules by default.

* Implemented new apple_archive rule for packaging Apple bundles into .ipa files
* Made sure to preserve existing functionality for packaging Symbols, Swift support files and so on.
* Created new providers (AppleSymbolsFileInfo, AppleSwiftDylibsInfo, AppleWatchosStubInfo, AppleMessagesStubInfo) for propagating archive-specific metadata
* Enabled tree artifact outputs by default for iOS/tvOS/visionOS applications to support .ipa creation
* Removed include_symbols_in_bundle attribute from ios_application (symbols now handled separately via apple_archive)
* Updated tests to reference newly created `apple_archive` targets which now exercise this rule.
* Updated documentation

Remaining things to do:
* Rename all occurrences of `ipa_post_processor` to `post_processor`
* Support producing `.zip` archives for `macos_application`. This is a little tricky since there is framework rules that don't support tree artifacts for macOS.
* Remove all tree artifacts related code from `processor.bzl`


apple/internal/apple_archive.bzl
```diff
@@ -0,0 +1,351 @@
...
+    files from all `{binary: .dSYM, ...}` pairs for the application and its
+    dependencies, then packages them under the `Symbols/` directory in the
+    final .ipa.
+    """,
+        ),
+        "_xplat_tools": attr.label(
+            default = Label("//apple/internal:xplat_tools_toolchain"),
+            providers = [AppleXPlatToolsToolchainInfo],
+            doc = """\
+A reference too xcplat_toolchain.
```
aaronsky, line 325:
```suggestion
A reference to xplat_tools_toolchain.
```

PR #2840, opened 2025-12-11
Propagate output groups from Swift and other dependencies

davbeck:
This PR enables all Apple bundle rules (`ios_framework`, `ios_application`, `macos_application`, etc.) to propagate output groups from their dependencies, particularly Swift-related output groups from `swift_library` targets. This is to support work I did in https://github.com/bazelbuild/rules_swift/pull/1628 to output diagnostic files.

Previously, output groups like `swift_index_store` were only available directly on `swift_library` targets. This made it difficult to collect supplemental compilation outputs (diagnostics, indexstore, AST files, etc.) when Swift code was built as part of a framework or application bundle.

Full disclosure: most of these changes were made using Cursor and only verified on 1 project.

### DIFF
--- a/test/starlark_tests/ios_application_tests.bzl
+++ b/test/starlark_tests/ios_application_tests.bzl
@@ -31,6 +31,10 @@ load(
     "analysis_target_actions_tree_artifacts_outputs_test",
 )
 load(
+    "//test/starlark_tests/rules:output_group_test.bzl",
+    "output_group_test",
+)
+load(
     "//test/starlark_tests/rules:analysis_target_outputs_test.bzl",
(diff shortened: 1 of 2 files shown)


brentleyjones:
I don't think we should do this. Output groups are tied to the targets that produce the output. If you want to collect these outputs transitively the idiomatic way to do that is with an aspect.

davbeck:
> I don't think we should do this. Output groups are tied to the targets that produce the output. If you want to collect these outputs transitively the idiomatic way to do that is with an aspect.

Thanks for the feedback. I'll look into that.

PR #2851, opened 2026-01-08
Remove unsupported includes parameter for apple_metal_library

mjburghard:
`includes` parameter was never properly supported and is not forwarded to the Metal compiler.


adincebic:
@mjburghard is this available upstream?

adincebic (approved):

Issue #2878, opened 2026-02-09
Support multiple module on `app_intents` of applications


kntkymt:
## Background

Xcode 26.0 added support for defining App Intents across multi-module such as those built with SwiftPM package. This makes it possible to separate App Intent definitions across multiple modules, allowing App Intents to be separated according to responsibility. 

In Addition, by splitting App Intents into separated modules, we can precisely control which App Intents are bundled into each application target.
 e.g. 

- `A`: shared AppIntents module
- `B`: iOS specific AppIntents module

then bundle both to iOS app, bundle only `A` to Widget extension. 
(we have a motivation to control it since running process of AppIntents depends on whether it's bundled to which applications. [ref](https://stackoverflow.com/questions/76656439/how-to-add-appintent-to-app-target-and-use-it-in-an-interactive-widget) )

### How it's achieved in xcodebuild

This is achieved via the `--static-metadata-file-list` argument of `appintentmetadataprocessor`. Passing multiple `Metadata.appintents/extract.actionsdata` as input through this argument, merged into a single metadata.

[reference implementation of Swift Build](https://github.com/swiftlang/swift-build/blob/db2fabe4d0c61913f076f6e45a7634fa3a8edc6f/Sources/SWBApplePlatform/AppIntentsMetadataCompiler.swift#L114-L118)

we can see the behavior when we create App support AppIntents with multi module configuration.

<img width="1509" height="312" alt="Image" src="https://github.com/user-attachments/assets/57273cb3-8e1f-47b4-ac2f-f5cf4ef6f946" />

## What to do

Support multiple module on `app_intents` of applications using  `--static-metadata-file-list` of `appintentmetadataprocessor`
currently we only support a single module on it.

## Appendix

`--static-metadata-file-list` also supports solving dependency AppIntens like using AppIntents/AppEntity defined in external module. 
in order to support this behavior, we should recursively call `appintentmetadataprocessor` for each dependency of target passed to `app_intents` of application.

I think we can support this in other issue/PR since its functionally could be big change, as currently `rules_apple` calling `appintentmetadataprocessor` for only the target passed to `app_intents` of application.

PR #2879, opened 2026-02-09
Support multiple modules on `app_intents` for applications

kntkymt:
# Background

resolve https://github.com/bazelbuild/rules_apple/issues/2878

# Approach
## As-is

1. Generate App Intents metadata for a target passed to app_intents (only a single target is supported).
1. Bundle the output from step 1 into the application.

### When using Xcode 26 or later

1. Generate App Intents metadata (Metadata.appintents) for each target passed to app_intents.
1. Generate an additional App Intents metadata file for a dummy intermediate target by passing the metadata files from `1.` to `--static-metadata-file-list`.
    - `appintentmetadataprocessor` merges the metadata files into a single metadata.
1. Bundle the output from step 2 into the application.

Note: `appintentmetadataprocessor` supports `--static-metadata-file-list` starting with the Xcode 26 toolchain.

### Otherwise

Keep the current behavior.

# Sample

sample project

https://github.com/kntkymt/ios-app-intents-sample/tree/main/bazel-multi-app-intents

### DIFF
--- a/apple/internal/apple_xcframework_import.bzl
+++ b/apple/internal/apple_xcframework_import.bzl
@@ -690,7 +690,16 @@ def _apple_static_xcframework_import_impl(ctx):
         linkopts = sdk_linkopts + linkopts,
         swiftinterface_imports = [xcframework_library.swift_module_interface] if xcframework_library.swift_module_interface else [],
         swiftmodule_imports = xcframework_library.swiftmodule,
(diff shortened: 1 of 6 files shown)


kntkymt:
any kind of feedbacks are welcome (ideas, approaches, blockers need to solve first, etc...)

PR #2889, opened 2026-03-07
Fix SDK->simruntime mapping assumptions in test runners

aaronsky:
Redux of #2675. 

Refactors simulator_creator.py to discover a recent simruntime when none is explicitly declared, and removes the SDK version fallback from the test runner rule implementations. This assumption is problematic in recent times whenever Apple pushes a new simruntime OTA for an existing Xcode, or when the simruntime identifier does not align with the simruntime version and consequently the SDK version (such as OS patch releases).


adincebic (approved):
Looks good at first glance. Are identifier and version enough to match the runtime or do we have to take platform into account as well? Perhaps identifier is enough?

luispadron (approved):
Thank you!

adincebic:
@aaronsky I just hit the issue with patch version and proposed a change in https://github.com/bazelbuild/rules_apple/pull/3049 let me know if it makes sense or if there is a better solution.

PR #2911, opened 2026-04-01
feat: Add support for ios_kernel_extension() bazel rule

YungRaj:
This PR is meant to add support for building an iOS kernel extension. There is no practical use case for this other than building kernel extensions that can support Virtualization.framework and Private Cloud Compute iOS VMs for example with https://github.com/wh1te4ever/super-tart-vphone

Still a work in progress need to touch up a few things.


YungRaj:
Sorry I broke the tests should be fixed now.

YungRaj:
Okay so I needed to ensure that the macOS IOKit headers are used when using IOKit to build the kernel extension so that classes like IOService, IORegistryEntry, etc can be used. Just added support for that in recent updates to the commit.

Requires https://github.com/bazelbuild/apple_support/pull/570 for merge

test/starlark_tests/resources/kext_resources/BUILD
```diff
@@ -31,5 +31,8 @@ cc_library(
         "-isystem__BAZEL_XCODE_SDKROOT__/System/Library/Frameworks/Kernel.framework/Headers",
     ],
     tags = common.fixture_tags,
+    deps = [
+        "@macos_iokit_headers//:iokit_headers",
+    ],
```
keith: isn't the `-isystem` above attempting to do this same thing?

ilhanraja:
```
In file included from test/starlark_tests/resources/kext_resources/KextBuilder.cc:15:
test/starlark_tests/resources/kext_resources/KextBuilder.h:15:10: fatal error: 'IOKit/IOService.h' file not found
   15 | #include <IOKit/IOService.h>
      |          ^~~~~~~~~~~~~~~~~~~
test/starlark_tests/resources/kext_resources/KextBuilder.h:15:10: note: did not find header 'IOService.h' in framework 'IOKit' (loaded from '/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneSimulator.platform/Developer/SDKs/iPhoneSimulator26.4.sdk/System/Library/Frameworks')
1 error generated.
Error in child process '/usr/bin/xcrun'. 1
Target //test/starlark_tests/targets_under_test/ios:kext failed to build
Use --verbose_failures to see the command lines of failed build steps.
INFO: Elapsed time: 4.277s, Critical Path: 3.10s
INFO: 11 processes: 41 action cache hit, 7 internal, 4 darwin-sandbox.
ERROR: Build did NOT complete successfully
ilhanraja-mac:rules_apple ilhanraja$
```

Getting this without adding the headers as a dep. Presumably the SDKROOT will point to the iOS SDK which doesn't have the kernel framework headers

keith: hrm i'm skeptical that we actually want the macOS version of the headers when building for iOS tho?

but either way you can do this instead of the repo rule:

```
-isystem__BAZEL_XCODE_DEVELOPER_DIR__/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/System/Library/Frameworks/Kernel.framework/Headers
```

YungRaj: There's no version of the iOS Kernel framework headers in the iOS SDK. The thing is though all the headers do (beyond compiling the source) is dynamically link against the symbols in the XNU kernel that are exported, including the symbols that are needed for starting a kext via IOKit. The symbol names to resolve will not change at all between macOS and iOS. I don't foresee that that this should be a problem when injecting the kext into an iOS kernelcache and loading it during runtime.

But I made the changes you suggested. Done.

keith (approved):

test/starlark_tests/resources/kext_resources/BUILD
keith: Yea that's pretty interesting. Ofc the problem is at any new Xcode version they could add macOS specific stuff to those headers and we'd have to do something else. But I'm happy to wait that out especially for tests

YungRaj: Sounds good.  Do reach out to me if any tests break and I can handle fixing anything that comes up.

YungRaj:
It looks like the release 2 days ago didn't include this PR. The `apple_support` repo release did contain my changes tho. 

When will the next release for this repo take place?

PR #2919, opened 2026-04-19
Provide default Info.plist for *_application rules

### LINKED ISSUE #1858: Feature request: Don't require an Info.plist for applications
New Xcode projects don't have an Info.plist, and instead generate one from default values. I've been using this simple version in all my new projects, and we can probably just default our base Info.plist to match it:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<plist version="1.0">
<dict>
	<key>CFBundleName</key>
	<string>$(PRODUCT_NAME)</string>
	<key>CFBundleVersion</key>
	<string>1.0</string>
	<key>CFBundleShortVersionString</key>
	<string>1.0</string>
	<key>CFBundleIdentifier</key>
	<string>$(PRODUCT_BUNDLE_IDENTIFIER)</string>
	<string>$(PRODUCT_BUNDLE_PACKAGE_TYPE)</string>
</dict>
</plist>
```

ssarad:
Fixes https://github.com/bazelbuild/rules_apple/issues/1858

The issue is quite old but we're currently at a larger monorepo with several apps so I figured this would be nice to get in. In this PR, I focused mostly on` ios_application` but happy to hear some thoughts if we want to expand this towards `*_application`


README.md
```diff
@@ -47,7 +47,7 @@ ios_application(
         "iphone",
         "ipad",
     ],
-    infoplists = [":Info.plist"],
+    infoplists = [":Info.plist"],  # Optional; a default Info.plist will be used if not provided
```
keith, line 50:
```suggestion
```

i think we should toss this from the example. IMO the value of this example is just providing the most minimal info

keith (commented):
is the way that xcode treats this plist just that it's a default and if you have custom stuff? or does it always merge it in to whatever you provide too? if it's the latter i wonder if instead of just treating this as a default if we should just do something more like `default_plist | user_plist` always.

also can this apply to more targets than just ios_application? i assume some others could benefit from this too?

this is very nice to do, i always find this annoying in simple cases

README.md
ssarad: Yeah, makes sense. Also for 

> is the way that xcode treats this plist just that it's a default and if you have custom stuff? or does it always merge it in to whatever you provide too? if it's the latter i wonder if instead of just treating this as a default if we should just do something more like default_plist | user_plist always.

Its actually due to some custom stuff but also we are moving towards a monorepo setup so before all apps join, they'd want to evaluate and play around with simple Bazel setup and often times I'm running myself into creating these info.plist which I realized could be much simpler if we just made a default one like Xcode does

ssarad: > also can this apply to more targets than just ios_application? i assume some others could benefit from this too?

Yeah this was my thinking too - should we expand to all *_applications? I think yes, and I'll do that and make the push, if you feel we want to scope it to just ios_application or macos_application, let me know

apple/internal/rule_attrs.bzl
```diff
@@ -716,6 +723,13 @@ _TEST_HOST_ASPECTS = [framework_provider_aspect]
 # Returns the default root Info.plist required to support a test bundle rule.
 _test_bundle_infoplist = "//apple/testing:DefaultTestBundlePlist"
 
+# Returns the default Info.plist for application rules.
+_ios_application_infoplist = "//apple/internal/templates:default_application_infoplist"
+_macos_application_infoplist = "//apple/internal/templates:default_application_infoplist"
+_tvos_application_infoplist = "//apple/internal/templates:default_application_infoplist"
+_watchos_application_infoplist = "//apple/internal/templates:default_application_infoplist"
+_visionos_application_infoplist = "//apple/internal/templates:default_application_infoplist"
```
keith:
```suggestion
_default_infoplist = "//apple/internal/templates:default_application_infoplist"
```

apple/internal/rule_attrs.bzl
```diff
@@ -749,5 +763,10 @@ rule_attrs = struct(
             watchos = ["watch"],
         ),
         test_bundle_infoplist = _test_bundle_infoplist,
+        ios_application_infoplist = _ios_application_infoplist,
+        macos_application_infoplist = _macos_application_infoplist,
+        tvos_application_infoplist = _tvos_application_infoplist,
+        watchos_application_infoplist = _watchos_application_infoplist,
+        visionos_application_infoplist = _visionos_application_infoplist,
```
keith:
```suggestion
        ios_application_infoplist = _default_infoplist,
        macos_application_infoplist = _default_infoplist,
        tvos_application_infoplist = _default_infoplist,
        watchos_application_infoplist = _default_infoplist,
        visionos_application_infoplist = _default_infoplist,
```

test/starlark_tests/macos_application_tests.bzl
```diff
@@ -521,6 +525,25 @@ def macos_application_test_suite(name):
...
+    infoplist_contents_test(
+        name = "{}_no_infoplist_has_default_values_test".format(name),
+        target_under_test = "//test/starlark_tests/targets_under_test/macos:app_minimal_no_infoplist",
+        expected_values = {
+            "CFBundleIdentifier": "com.google.example",
+            "CFBundleName": "app_minimal_no_infoplist",
+            "CFBundlePackageType": "APPL",
+        },
+        tags = [name],
+    )
```
keith, line 545: can you also add a test specifically for validating that if an app provides a plist with a different value for attributes that are in the default plist, the provided one's values win

ssarad: Added here https://github.com/bazelbuild/rules_apple/pull/2919/changes/830cee1aca71f839d8ceabdb8f53a2c3087f04a8 the provided one should win (2.0 vs 1.0 as the default one)

keith (approved):

aaronsky (approved):
stamping for application, but we should really do this for all of the bundle packaging rules where Xcode supplies defaults.

ssarad:
> stamping for application, but we should really do this for all of the bundle packaging rules where Xcode supplies defaults.

Oooh that’s a good shout. I’ll throw an issue so we can track this. I can likely pick that up when I have some time

PR #2922, opened 2026-04-22
Support module.private.modulemap in framework and xcframework imports

john-flanagan:
I ran into issues compiling https://github.com/mapbox/mapbox-common-ios through rules_swift_package_manager when using RBE. I believe the issue is that the `module.private.modulemap` file wasn't being handled by `apple_dynamic_xcframework_import`

https://github.com/swiftlang/llvm-project/blob/next/clang/docs/Modules.rst#private-module-map-files

Claude wrote most of this code. I'm not totally clear on why `xcframework_library.private_clang_module_map` is being added to `header_imports` when `xcframework_library.clang_module_map` is not, but I tried removing it and my build broke.

Should the private module map info be included in this block of code somehow?
```star lark
        # Create SwiftInteropInfo provider for swift_clang_module_aspect
        swift_interop_info = framework_import_support.swift_interop_info_with_dependencies(
            deps = deps,
            module_name = framework.bundle_name,
            module_map_imports = framework.module_map_imports,
        )
```

### DIFF
--- a/apple/internal/apple_framework_import.bzl
+++ b/apple/internal/apple_framework_import.bzl
@@ -222,10 +222,12 @@ def _apple_dynamic_framework_import_impl(ctx):
         ),
     ))
 
-    # Create CcInfo provider.
+    # Collect header imports
     header_imports = list(framework.header_imports)
     if framework.private_module_map_imports:
         header_imports.extend(framework.private_module_map_imports)
+
+    # Create CcInfo provider.
     cc_info = framework_import_support.cc_info_with_dependencies(
         actions = actions,
         cc_toolchain = cc_toolchain,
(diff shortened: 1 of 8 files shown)


adincebic:
There is a flag to enable private modulemap propagatiom

john-flanagan:
@adincebic I think that flag resolves the need for these blocks of code:
```starlark
    # Collect header imports
    header_imports = list(framework.header_imports)
    if framework.private_module_map_imports:
        header_imports.extend(framework.private_module_map_imports)
```

I believe the change to find `module.private.modulemap` and pass it to `xcframework_processor_tool` is still required.

Should there any updates to the SwiftInteropInfo provider?

john-flanagan:
> I believe the change to find module.private.modulemap and pass it to xcframework_processor_tool is still required.

Actually maybe not... I'll do more testing tomorrow

john-flanagan:
Never mind, `--define=apple.incompatible.objc_framework_propagate_modulemap=true` resolves my issue. Thanks for the pointer @adincebic

PR #2924, opened 2026-04-26
Ensure kernel extension rules don't produce entitlements sections if entitlements is None

YungRaj:
After testing the new `ios_kernel_extension()` rule that I added, I noticed that the final artifacts unnecessarily include the `__TEXT.__entitlements` and `__TEXT.__ents_der` even when entitlements is `None`. This pull request fixes that.


test/starlark_tests/ios_kernel_extension_tests.bzl
```diff
@@ -66,6 +78,28 @@ def ios_kernel_extension_test_suite(name):
         tags = [name],
     )
 
+    _action_arm64e_ios_cpu_test(
```
keith, line 81: can you add a positive version of this to make sure the argv doesn't drift and give us false negatives?

ilhanraja: Done

keith: I mean also with a `-sectcreate` arg

YungRaj: Oh sorry. Just needed clarification. Done.

However, please check the `apple/internal `changes as I needed to make those changes such that when the entitlements are present, it adds the `-sectcreate` flags to the rule properly when targeting a device. 

Also I might be wrong, but action command line test changes are also needed so that the `action_command_line_test()` can parse out comma separated flags properly without needing to type out the full path to e.g. the entitlements file.

keith (approved):

PR #2927, opened 2026-04-30
Manually compile from swiftinterface files by default

keith:
We flipped this off because of breakages in the past, but this is
required for explicit modules. Users can still opt out with
`-apple.import_framework_via_swiftinterface` in the features, but we're
also trying to iron out the issues.


keith:
previous thread https://bazelbuild.slack.com/archives/C04DFUBQGSU/p1692298016393289

keith:
likely depends on https://github.com/bazelbuild/rules_swift/pull/1705

aaronsky (approved):

PR #2933, opened 2026-05-07
Adopt symbolic macros for public-facing API

aaronsky:
This PR migrates legacy macros in our public API to symbolic macros, now that we're on Bazel 8. This will allow downstream consumers to wrap these macros in their own symbolic macros. It does not make any changes to the core_ml and intent library macros in `apple/resources.bzl`. 

Reviewers and tests should assert that this PR:

- Does not introduce breaking changes, except where possible to support symbolic macros
- Does not regress functionality through new attr type annotations or performance characteristics
- Makes sense to do at all


keith (approved):

apple/ios.bzl
```diff
@@ -55,68 +69,129 @@ ios_imessage_extension = _ios_imessage_extension
...
     apple_test_assembler.assemble(
         name = name,
         bundle_rule = _ios_internal_ui_test_bundle,
         test_rule = _ios_ui_test,
         runners = runners,
+        visibility = visibility,
         **kwargs
     )
 
+ios_ui_test_suite = macro_factory.create_apple_test_suite_macro(
```
congt, line 181: @aaronsky We cannot pass `runners` to `ios_ui_test_suite` or `ios_unit_test_suite` anymore. The attr `runner` inherited from ios_ui_test is mandatory. Meanwhile, `apple_test_assembler.assemble` fails if we pass both `runner` and `runners` because of [this check](https://github.com/bazelbuild/rules_apple/blob/2bf52c2458a32821b1748a8a3f9e106aa5d76325/apple/internal/testing/apple_test_assembler.bzl#L91-L92)

aaronsky: I can take a look. In the future, please file an issue rather than commenting on a merged PR like this.

aaronsky: #2983

PR #2935, opened 2026-05-07
Add pipefail to all verifier_scripts

keith:
I locally broke a test and realized this was hiding some errors


aaronsky (approved):

keith:
ok yea so actually tests are broken on main because of the same case i found, i will merge this into https://github.com/bazelbuild/rules_apple/pull/2934

PR #2936, opened 2026-05-08
Remove CI workarounds

keith:
im not sure if this being green means it's fixed or we mutated enough machines already but it saves a bunch of time, we can revert if needed


keith:
im not sure if this being green means it's fixed or we mutated enough machines already

luispadron (approved):

aaronsky (approved):

PR #2955, opened 2026-05-12
Remove objc_library archive from runfiles

keith:
This was always weird and was removed from objc_library in https://github.com/bazelbuild/rules_cc/commit/d305b348ed81f86f305c064f5f756614124ef6ae

The one from apple_static_library still exists, but that makes a bit
more sense since it's meant to be redistributable.


keith:
requires rules_cc > [0.2.18](https://github.com/bazelbuild/rules_cc/releases/tag/0.2.18)

keith:
seems to not care if we don't enumerate all the runfiles, so we can merge this now

aaronsky (approved):

PR #2957, opened 2026-05-13
Skip transitive Swift modules without a swiftmodule File

erneestoc:
## Summary

`_apple_test_info_aspect_impl` iterates `module.swift.swiftmodule` for every transitive Swift module. With the `system_swiftmodule` rule landing in rules_swift (https://github.com/bazelbuild/rules_swift/pull/1745), prebuilt SDK swiftmodules expose their path via `swiftmodule_path` (a placeholder string) and leave the `swiftmodule` `File` slot unset, which crashes this iteration.

Skip entries where `swift.swiftmodule` is `None` so the test bundle aspect tolerates the new provider shape.

## Test plan

- [ ] Build a test bundle that depends on a `system_swiftmodule` target and confirm the aspect no longer crashes
- [ ] Existing test bundles continue to pick up their Swift module inputs

### DIFF
--- a/apple/internal/testing/apple_test_bundle_support.bzl
+++ b/apple/internal/testing/apple_test_bundle_support.bzl
@@ -165,7 +165,7 @@ def _apple_test_info_aspect_impl(target, ctx):
         module_swiftmodules = [
             module.swift.swiftmodule
             for module in all_modules
-            if module.swift
+            if module.swift and module.swift.swiftmodule
         ]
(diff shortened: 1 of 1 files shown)


keith:
is this fixing something that exists today or only with https://github.com/bazelbuild/rules_swift/pull/1745?

if it's fixing an issue right now can you add a test covering that? i kinda expected https://github.com/bazelbuild/rules_apple/pull/2958 to fail given this pr but it does not

PR #2964, opened 2026-05-14
Convert tvos_application_test to starlark

keith:


keith:
depends on the new rule from https://github.com/bazelbuild/rules_apple/pull/2963

aaronsky (approved):

PR #2965, opened 2026-05-14
Convert watchos_application_test to starlark

keith:


keith:
requires some of the other ones to land for new rules

aaronsky (approved):

PR #2967, opened 2026-05-14
Convert ios_coverage_test to starlark

keith:
This isn't perfect but we couldn't run the other one on CI and this way
we don't have to run bazel in bazel


keith:
depends on https://github.com/bazelbuild/rules_apple/pull/2968

aaronsky (approved):

PR #2974, opened 2026-05-19
Add app run modes to Apple application runners

brentleyjones:
Adds a `BAZEL_APPLE_RUN_MODE` control to the device and simulator application runners so callers can install without launching or relaunch without reinstalling. This also clears stale launch info for install-only runs and uses explicit terminate-before-launch behavior for deterministic profiling workflows.


adincebic (approved):
Useful. Perhaps we should document this in doc/?

PR #2977, opened 2026-05-26
Correctly handle module.private.modulemap files

keith:
Modules can contain `module.private.modulemap` alongside
`module.modulemap`. In this case we have to precompile both for explicit
modules, and the private module must be precompiled with the public
module as a dependency. We now handle this by using lists of modulemaps

This change also revealed that we weren't correctly handling frameworks that
were missing modulemaps but had a swiftmodule.


adincebic (approved):
Thanks! I hit this recently.

aaronsky (approved):

PR #2985, opened 2026-06-02
Fix resuse_simulator value always being True

john-flanagan:
`resuse_simulator` was always being set to `True` because of the default value of the `--reuse-simulator` argument.

This PR updates the logic not include a default value for `--reuse-simulator` directly, while still defaulting to `True` when the argument is missing and `SIMULATOR_REUSE_SIMULATOR` is not `0` or `False`.

The previous check for `os.getenv("SIMULATOR_REUSE_SIMULATOR") is not None` wasn't working correctly either because the template always set either `SIMULATOR_REUSE_SIMULATOR=1` or `SIMULATOR_REUSE_SIMULATOR=` which resolve to `"1"` and `""` respectively, both of which returned `True` from `is not None`.

I added a print statement just below the `reuse_simulator` logic to illustrate the problem

### Before
```
$ SIMULATOR_DEVICE_TYPE="iPhone 15" SIMULATOR_REUSE_SIMULATOR= ./simulator_creator.py
Reuse simulator True

$ SIMULATOR_DEVICE_TYPE="iPhone 15" SIMULATOR_REUSE_SIMULATOR=1 ./simulator_creator.py
Reuse simulator True

$ SIMULATOR_DEVICE_TYPE="iPhone 15" SIMULATOR_REUSE_SIMULATOR=True ./simulator_creator.py
Reuse simulator True

$ SIMULATOR_DEVICE_TYPE="iPhone 15" SIMULATOR_REUSE_SIMULATOR=0 ./simulator_creator.py
Reuse simulator True

$ SIMULATOR_DEVICE_TYPE="iPhone 15" SIMULATOR_REUSE_SIMULATOR=False ./simulator_creator.py
Reuse simulator True

$ SIMULATOR_DEVICE_TYPE="iPhone 15" SIMULATOR_REUSE_SIMULATOR=Foo ./simulator_creator.py
Reuse simulator True
```

### After

```
$ SIMULATOR_DEVICE_TYPE="iPhone 15" SIMULATOR_REUSE_SIMULATOR= ./simulator_creator.py
Reuse simulator True

$ SIMULATOR_DEVICE_TYPE="iPhone 15" SIMULATOR_REUSE_SIMULATOR=1 ./simulator_creator.py
Reuse simulator True

$ SIMULATOR_DEVICE_TYPE="iPhone 15" SIMULATOR_REUSE_SIMULATOR=True ./simulator_creator.py
Reuse simulator True

$ SIMULATOR_DEVICE_TYPE="iPhone 15" SIMULATOR_REUSE_SIMULATOR=0 ./simulator_creator.py
Reuse simulator False

$ SIMULATOR_DEVICE_TYPE="iPhone 15" SIMULATOR_REUSE_SIMULATOR=False ./simulator_creator.py
Reuse simulator False

$ SIMULATOR_DEVICE_TYPE="iPhone 15" SIMULATOR_REUSE_SIMULATOR=Foo ./simulator_creator.py
Reuse simulator True
```

I think this logic would be better if we explicitly checked `SIMULATOR_REUSE_SIMULATOR` for `True` or `1` and defaulted everything else to `False`, but I wasn't sure if that would mesh with the help text of "Toggle simulator reuse; default is True"

i.e.
```python
    reuse_simulator_env = os.getenv("SIMULATOR_REUSE_SIMULATOR")
    reuse_simulator: bool = args.reuse_simulator or (
        reuse_simulator_env.lower() in ("1", "true") if reuse_simulator_env else False
    )
```

### DIFF
--- a/apple/testing/default_runner/simulator_creator.py
+++ b/apple/testing/default_runner/simulator_creator.py
@@ -211,7 +211,7 @@ def _build_parser() -> argparse.ArgumentParser:
     parser.add_argument(
         "--reuse-simulator",
         action=argparse.BooleanOptionalAction,
-        default=True,
+        default=None,
         help="Toggle simulator reuse; default is True",
     )
     return parser
@@ -231,7 +231,7 @@ def _main() -> None:
     os_version = args.os_version or os.getenv("SIMULATOR_OS_VERSION")
     sdk_version = args.sdk_version or os.getenv("SIMULATOR_SDK_VERSION")
     reuse_simulator: bool = args.reuse_simulator or (
-        os.getenv("SIMULATOR_REUSE_SIMULATOR") is not None
+        os.getenv("SIMULATOR_REUSE_SIMULATOR", default = "1").lower() in ("1", "true")
     )
 
     selected_runtime = _selected_simulator_runtime(os_version, sdk_version)
--- a/apple/testing/default_runner/simulator_creator.py
+++ b/apple/testing/default_runner/simulator_creator.py
@@ -230,7 +230,7 @@ def _main() -> None:
 
     os_version = args.os_version or os.getenv("SIMULATOR_OS_VERSION")
     sdk_version = args.sdk_version or os.getenv("SIMULATOR_SDK_VERSION")
-    reuse_simulator: bool = args.reuse_simulator or (
+    reuse_simulator: bool = args.reuse_simulator if args.reuse_simulator is not None else (
         os.getenv("SIMULATOR_REUSE_SIMULATOR", default = "1").lower() in ("1", "true")
     )
 


apple/testing/default_runner/simulator_creator.py
```diff
@@ -230,8 +230,10 @@ def _main() -> None:
 
     os_version = args.os_version or os.getenv("SIMULATOR_OS_VERSION")
     sdk_version = args.sdk_version or os.getenv("SIMULATOR_SDK_VERSION")
+
+    reuse_simulator_env = os.getenv("SIMULATOR_REUSE_SIMULATOR")
     reuse_simulator: bool = args.reuse_simulator or (
-        os.getenv("SIMULATOR_REUSE_SIMULATOR") is not None
+        reuse_simulator_env.lower() not in ("0", "false") if reuse_simulator_env else True
```
keith:
```suggestion
        reuse_simulator_env and reuse_simulator_env.lower() in ("1", "true")
```

maybe a bit more clear?

apple/testing/default_runner/ios_xctestrun_runner.template.sh
```diff
@@ -379,7 +379,7 @@ readonly profraw="$test_tmp_dir/coverage.profraw"
 if [[ "%(reuse_simulator)s" == true ]]; then
   reuse_simulator=1
 else
-  reuse_simulator=
+  reuse_simulator=0
```
keith: can you change the callsites of this to not have `:-` so it errors if it is unset now?

apple/testing/default_runner/simulator_creator.py
john-flanagan: With that change as written I got this output:
```
$ SIMULATOR_DEVICE_TYPE="iPhone 15" SIMULATOR_REUSE_SIMULATOR= ./simulator_creator.py
Reuse simulator
```

I had to update it to 
```python
    reuse_simulator_env = os.getenv("SIMULATOR_REUSE_SIMULATOR")
    reuse_simulator: bool = args.reuse_simulator or (
        reuse_simulator_env is not None and reuse_simulator_env.lower() in ("1", "true")
    )
```
to get 
```
$ SIMULATOR_DEVICE_TYPE="iPhone 15" SIMULATOR_REUSE_SIMULATOR= ./simulator_creator.py
Reuse simulator False
```

john-flanagan: What about
```python
    reuse_simulator: bool = args.reuse_simulator or (
        os.getenv("SIMULATOR_REUSE_SIMULATOR", default = "").lower() in ("1", "true")
    )
```

john-flanagan: actually no, that changes the default behavior. It'd have to be:
```python
    reuse_simulator: bool = args.reuse_simulator or (
        os.getenv("SIMULATOR_REUSE_SIMULATOR", default = "1").lower() in ("1", "true")
    )
```
which is maybe a bit weird

apple/testing/default_runner/simulator_creator.py
```diff
@@ -211,7 +211,7 @@ def _build_parser() -> argparse.ArgumentParser:
     parser.add_argument(
         "--reuse-simulator",
         action=argparse.BooleanOptionalAction,
-        default=True,
+        default=None,
```
aaronsky, line 214: iirc a default of None for a BooleanOptionAction isn’t valid. can we make this work with just a bool rather than an optional bool?

apple/testing/default_runner/ios_xctestrun_runner.template.sh
aaronsky: what distinction are we trying to draw between unset and set to 0 here? can we achieve the same result without changing this and just handling the discrepancy correctly in the simulator_creator script?

john-flanagan: I think with an update to something like this we could leave that as is:
```python
    reuse_simulator: bool = args.reuse_simulator or (
        os.getenv("SIMULATOR_REUSE_SIMULATOR", default = "1").lower() in ("1", "true")
    )
```

I wasn't sure how you wanted to resolve the argument comment of "Toggle simulator reuse; default is True" when `SIMULATOR_REUSE_SIMULATOR` is `""`, which is what happens with the current approach. Should that fall into the "default is True" case since it's basically not set, or should we treat `SIMULATOR_REUSE_SIMULATOR` being the empty string as False. I think that was the original intention, but it could be considered in conflict with the doc comment depending on how you look at it.

apple/testing/default_runner/simulator_creator.py
john-flanagan: I haven't run into issues with providing `None` for the default.
It looks like `None` is the default value for default here: https://github.com/python/cpython/blob/402668b2b1a63a2b3cfd7a2ede07f6786f9beb8e/Lib/argparse.py#L1020-L1027

PR #2988, opened 2026-06-02
Fix the attr runners of apple test-suite macros

congt:
* Make runners non-configurable in apple test-suite macros

The attr runners of apple test-suite macros was configurable. Symbolic macros pass configurable attribute values into the implementation function wrapped in a `select()`. The `_ios_*_test_suite_impl` forwards 'runners' to `apple_test_assembler.assemble`, which iterates through runners. Then it fails with the error 'type select is not iterable'.

* Use Label fields to get the runner name in apple_test_assembler.assemble

The attr runners is label_list now, while apple_test_assembler.assemble still treats it as a string list. The fix should work for both legacy callers and symbolic-macro callers


apple/internal/testing/apple_test_assembler.bzl
```diff
@@ -143,7 +143,7 @@ def _assemble(name, bundle_rule, test_rule, runner = None, runners = None, **kwa
     elif runners:
         tests = []
         for runner in runners:
-            test_name = "{}_{}".format(name, runner.rsplit(":", 1)[-1])
+            test_name = "{}_{}".format(name, str(runner).rsplit(":", 1)[-1])
```
aaronsky: this seems like an API regression, shouldn’t we instead use the fields on Label?

congt: Good idea! It's cleaner. I update it to `Label(runner).name` so that it still works for legacy macro callers

aaronsky (approved):

PR #2995, opened 2026-06-09
Support Xcode 27 simulator launch via DeviceHub

iMostfa:
Xcode 27 removed `Simulator.app` from the Xcode bundle. `Xcode.app/Contents/Developer/Applications/` no longer exists; the apps moved to `Xcode.app/Contents/Applications/`, and the simulator GUI is now hosted by `DeviceHub.app` (`com.apple.dt.Devices`).

`boot_simulator` in `apple_simulator.template.py` hardcodes:

```python
simulator_path = os.path.join(developer_path, "Applications/Simulator.app")
subprocess.run(["open", "-a", simulator_path, "--args", "-CurrentDeviceUDID", udid], check=True)
```

On Xcode 27 that path does not exist, so `open` fails and the `check=True` raises, breaking `bazel run` of any simulator `*_application` target with an `NSFilePath=.../Contents/Developer/Applications/Simulator.app` error.

## Change

- Boot the device explicitly via `simctl boot` rather than relying on launching `Simulator.app` with `-CurrentDeviceUDID` as the boot trigger (DeviceHub does not boot a device on launch). This works on every Xcode version; an already-booted device exits non-zero and is ignored.
- Open a GUI window **best-effort**: prefer the legacy `Simulator.app`, fall back to `DeviceHub.app` under `Contents/Applications`, and warn if neither is present. The window is cosmetic for `bazel run`, so a missing/failing launcher no longer fails the run.
- `wait_for_sim_to_boot` (`simctl bootstatus`) still gates success, so behavior is unchanged on older Xcodes.

## Testing

- `python3 -m py_compile` on the changed template.
- Verified the offending path is absent in an Xcode 27 beta bundle (`Contents/Developer/Applications/` removed; `Simulator.app` gone; `DeviceHub.app` present under `Contents/Applications/`).

### DIFF
--- a/apple/internal/templates/apple_simulator.template.py
+++ b/apple/internal/templates/apple_simulator.template.py
@@ -466,10 +466,41 @@ def boot_simulator(*, developer_path: str, simctl_path: str, udid: str) -> None:
 
   # Boot the device explicitly via simctl. Historically the boot was a side
   # effect of launching Simulator.app with `-CurrentDeviceUDID`, but Simulator.app
-  # was removed from the Xcode bundle in favor of DeviceHub.app (Xcode 26+), which
+  # was removed from the Xcode bundle in favor of DeviceHub.app (Xcode 27), which
   # does not boot a device on launch. Booting via simctl works on every Xcode
-  # version; an already-booted device exits non-zero, which we ignore.
-  subprocess.run([simctl_path, "boot", udid], check=False)
+  # version.
+  try:
+    subprocess.run([simctl_path, "boot", udid], check=True)
+  except subprocess.CalledProcessError as e:
+    # `simctl boot` errors out if the device is already booted, exiting 149
+    # with "Unable to boot device in current state: Booted". Since our goal is
+    # simply for the device to be booted, that case is already satisfied and we
+    # can carry on. But 149 is not unique to "already booted" -- other states
+    # like "Shutting Down" report it too -- so, as simulator_creator.py does, we
+    # only ignore it after confirming via `simctl list` that the device really
+    # is booted. Every other failure propagates.
+    already_booted = False
+    if e.returncode == 149:
+      list_result = subprocess.run(
+          [simctl_path, "list", "devices", "-j", udid],
+          encoding="utf-8",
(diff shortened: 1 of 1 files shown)


apple/internal/templates/apple_simulator.template.py
```diff
@@ -463,20 +463,43 @@ def boot_simulator(*, developer_path: str, simctl_path: str, udid: str) -> None:
...
-  simulator_path = os.path.join(developer_path, "Applications/Simulator.app")
-  subprocess.run(
-      ["open", "-a", simulator_path, "--args", "-CurrentDeviceUDID", udid],
-      check=True,
+
+  # Boot the device explicitly via simctl. Historically the boot was a side
+  # effect of launching Simulator.app with `-CurrentDeviceUDID`, but Simulator.app
+  # was removed from the Xcode bundle in favor of DeviceHub.app (Xcode 26+), which
+  # does not boot a device on launch. Booting via simctl works on every Xcode
+  # version; an already-booted device exits non-zero, which we ignore.
```
brentleyjones: > an already-booted device exits non-zero, which we ignore.

I think we should check for that specific error code and only ignore that? Because if we fail for other reasons don't we want to fail?

brentleyjones (commented):
Also, needs Google CLA signed before we can accept this.

apple/internal/templates/apple_simulator.template.py
iMostfa: agreed, also i'm not sure if it's a an error in this DeviceHub in general, or just beta issues.

so now, it's aligned with simulator_creator.py:boot_simulator where we ignore 149 as it mean, simulator is already booted

aaronsky:
why xcode 26? device hub is new in xcode 27 beta

iMostfa:
i trusted claude more than i should @aaronsky
sorry, updated it

iMostfa:
@keith let me know what's missing in this PR, would love to get it moving forward

keith:
https://github.com/bazelbuild/rules_apple/pull/3013

PR #3001, opened 2026-06-11
Default to darwin_arm64 when building macOS targets with a minimum_os_version of 27.0 or higher

adincebic:
Starting from macOS 27 x86_64 is no longer supported.

(cherry picked from commit [a9d699e2abc0330d8b4c28f623d06214e45f957a](https://github.com/bazelbuild/rules_apple/commit/a9d699e2abc0330d8b4c28f623d06214e45f957a)


keith (approved):
we can land but we're also discussing defaulting to this always now. im quite confident we have more devs on arm64 now than x86_64

adincebic:
Landed as is. We can always drop x86_64 if there is no consumers.

Issue #2986, opened 2026-06-02
How to best cleanup simulators after test timeout?


john-flanagan:
We were running into an issue where non-reused simulators would end getting left over when hosted iOS unit tests timed out. We have a patch that combines the trap handler that cleans up `test_tmp_dir` with a call to `xcrun simctl delete` to clean up single-use simulators, but it doesn't seem like a particularly scalable solution.

It'd be nice to formalize calling into `clean_up_simulator_action_binary` when test timeouts happen and how that cleanup script could detect that the test timed out.

mansiverma897993:
Hi @john-flanagan Can you assign this issue to me ? or I can directly raise PR here ?

john-flanagan:
@mansiverma897993 If you have a suggested fix for this issue feel free to reference the issue in your PR. I don't think it needs to be assigned to you for that.

mansiverma897993:
@john-flanagan Look at once I have raised PR and lemme know if any changes  required!!

PR #3005, opened 2026-06-16
Fix iOS simulator leakage on test timeout (#2986)

mansiverma897993:
Clean up simulators and temporary directories when the test runner is aborted due to a timeout (receiving SIGTERM/SIGINT) by trapping these signals and setting a TEST_TIMEOUT_DETECTED environment variable for the cleanup script. Modify the default cleanup script to delete the simulator if it timed out, even if reuse is enabled.


apple/testing/default_runner/simulator_cleanup.sh
```diff
@@ -7,7 +7,7 @@ if [ -z "${SIMULATOR_UDID:-}" ]; then
   exit 1
 fi
 
-if [ -z "${SIMULATOR_REUSE_SIMULATOR:-}" ]; then
+if [ -z "${SIMULATOR_REUSE_SIMULATOR:-}" ] || [ -n "${TEST_TIMEOUT_DETECTED:-}" ]; then
```
john-flanagan, line 10: If the simulator is being shared between tests won't deleting it when a single test times out cause a bunch of unrelated failures?

PR #3010, opened 2026-06-18
Delete apple_split_cpu from transition

keith:
This comment says this is needed for the objc_library transition which
is long gone.

This transition was deleted first in bazel 8.x https://github.com/bazelbuild/bazel/commit/d553273303dde1555cfd69691fac3b705a3bffe7


keith:
failure to debug:

```
bazel test //test/starlark_tests:watchos_application_device_ios_cpus_platform_test \
    --config=ci \
    --xcode_version=26.0.0 \
    --test_output=errors \
    --verbose_failures
```

keith:
ok so this thing is still load bearing for watchos transitioning built through an iOS transition, it's fine to keep, nbd

PR #3016, opened 2026-07-06
Add support for newer [Skip|Only]TestingIdentifiers xctestplan format to test runner template

john-flanagan:
Xcode 26.4+ doesn't honor the `SkipTestIdentifiers` array for Swift Testing tests.

This PR adds the newer `SkipTestingIdentifiers` format which supports both XCTest and Swift Testing tests on Xcode 26.4+ and supersedes `SkipTestIdentifiers` if both are present.

I assume we probably need to add the same update to the macOS runner. I'd be happy to include that in this PR once folks are happy with the format of this change.

### DIFF
--- a/apple/testing/default_runner/ios_xctestrun_runner.template.sh
+++ b/apple/testing/default_runner/ios_xctestrun_runner.template.sh
@@ -335,6 +335,105 @@ if [[ -n "$main_thread_checker_dyld_env" ]]; then
   fi
 fi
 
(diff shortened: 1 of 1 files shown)


keith (approved):
i think it's fine, we need to rewrite this soon in a higher level language. I would wait on adding this to any other runners as @aaronsky is working on just merging them all https://github.com/bazelbuild/rules_apple/pull/3008

adincebic (approved):

PR #3019, opened 2026-07-09
Remove unused bazel_features dep

keith:


keith:
should be possible after https://github.com/bazelbuild/rules_swift/pull/1838

keith:
merged elsewhere

PR #3015, opened 2026-07-01
Implement developer frameworks repo

adincebic:
Frameworks like `XcodeKit` are bundled with Xcode so we need a way to expose them to users.

PR #3021, opened 2026-07-16
Add screen_capture_format attribute to ios_xctestrun_runner

erneestoc:
## Summary

Adds a new `screen_capture_format` string attribute to `ios_xctestrun_runner` that threads Xcode's `PreferredScreenCaptureFormat` xctestrun key (added in Xcode 15) into the generated file. Accepted values:

- `""` (default): key is not written, so Xcode's platform default is used — no behavior change for existing users.
- `"screenshots"`: capture screenshots only.
- `"screenRecording"`: capture a full screen recording of each UI test run, attached to the resulting `.xcresult` bundle.

The same value can be set ad hoc for a single run without editing the runner target via `--test_arg=--xctestrun_screen_capture_format=<value>`.

Setting a non-empty value implies:
1. Use of `xcodebuild test-without-building` (the direct `xctest` path does not read the xctestrun file).
2. `create_xcresult_bundle = True` (the capture is only observable in the `.xcresult`).

Both are documented on the attribute and echoed as `note:` lines when triggered, matching the existing `attachment_lifetime`/`xcodebuild_args` conventions.

## Interaction with `attachment_lifetime`

Screen captures are *system* attachments, so `attachment_lifetime` must be `"keepAlways"` or `"deleteOnSuccess"` for the capture to survive into the `.xcresult` — the default `"keepNever"` discards it. This is called out in the attribute doc, and the runner emits a `warning:` at runtime when a capture format is requested while the lifetime is `"keepNever"`.

## Test plan

- [x] `//test:ios_xctestrun_runner_ui_test` — 2 new template-emission cases:
  - `test_ios_ui_test_default_screen_capture_format_arg` — asserts `PreferredScreenCaptureFormat` is **not** emitted when the flag is unset.
  - `test_ios_ui_test_screen_capture_format_arg` — asserts the key + `screenRecording` value + `SystemAttachmentLifetime=keepAlways` are all written when both flags are set.
- [x] `//doc:check_ios` passes with regenerated `doc/rules-ios.md`.
- [x] `buildifier --lint=warn` passes clean.
- [x] Manual end-to-end verification against a UI test on iOS 18.6 simulator: `.xcresult` produced and opens in Xcode with the `.mov` video attachment visible in the Attachments panel.
- [x] Verified locally on Xcode 26.2 / iOS 26.2 simulator: both new cases plus `test_ios_ui_test_pass` and `test_ios_ui_test_attachment_lifetime_arg` pass.


apple/testing/default_runner/ios_xctestrun_runner.template.sh
```diff
@@ -494,7 +498,31 @@ if [[ "$build_for_device" == false ]]; then
...
+      echo "note: waiting for another test to finish creating a simulator"
+      /usr/bin/lockf -s /dev/fd/201
+    fi
+  fi
+
+  simulator_id="$(SIMULATOR_DEVICE_TYPE="%(device_type)s" SIMULATOR_OS_VERSION="%(os_version)s" SIMULATOR_REUSE_SIMULATOR="${reuse_simulator:-}" SIMULATOR_SDK_VERSION="%(sdk_version)s" XCTESTRUN_RUNNER_PID="${BASHPID:-$$}" "%(create_simulator_action_binary)s" 201>&-)"
+
+  if [[ -n "$simulator_creation_lock" ]]; then
+    exec 201>&-
+  fi
```
keith: can you split this out into a separate change? afaiui this isn't part of the change here. we should probably do this in the creator itself

apple/testing/default_runner/ios_xctestrun_runner.template.sh
```diff
@@ -538,12 +566,43 @@ if (( ${#custom_xcodebuild_args[@]} )); then
   echo "note: Using 'xcodebuild' because '--xcodebuild_args' was provided"
   should_use_xcodebuild=true
 fi
+if [[ -n "$screen_capture_format" ]]; then
+  echo "note: Using 'xcodebuild' because a screen capture format was requested"
+  should_use_xcodebuild=true
+  # The capture is only observable in the XCResult bundle, so force one.
+  create_xcresult_bundle=true
+  if [[ "$attachment_lifetime" == "keepNever" ]]; then
+    echo "warning: 'screen_capture_format' is set but 'attachment_lifetime' is 'keepNever'; the capture will be discarded before it reaches the .xcresult bundle" >&2
```
keith: should we error in this case?

apple/testing/default_runner/ios_xctestrun_runner.template.sh
```diff
@@ -538,12 +566,43 @@ if (( ${#custom_xcodebuild_args[@]} )); then
...
+simulator_lock=""
+if [[ "$build_for_device" == false && -x /usr/bin/lockf ]]; then
+  simulator_lock="$(getconf DARWIN_USER_TEMP_DIR 2>/dev/null || echo /tmp)"
+  simulator_lock="${simulator_lock%/}/rules_apple_simulator_$simulator_id.lock"
+  exec 200>"$simulator_lock"
+  if ! /usr/bin/lockf -s -t 0 /dev/fd/200; then
+    echo "note: waiting for another test to finish using simulator '$simulator_id'"
+    /usr/bin/lockf -s /dev/fd/200
+  fi
+fi
```
keith: same as above, separate pr

apple/testing/default_runner/ios_xctestrun_runner.template.sh
erneestoc: Split out into #3026 — and per the follow-up discussion it grew into a small simulator pool: rather than serializing concurrent tests on one shared simulator, each concurrent test action claims its own (slot 0 keeps today's simulator name/reuse). The create+boot serialization lives in the creator itself as suggested. Note this PR's simulator tests will flake on CI until #3026 lands, since touching the runner makes them actually execute concurrently against one shared simulator.

apple/testing/default_runner/ios_xctestrun_runner.template.sh
erneestoc: Agreed — changed to a hard error, with a test covering the failure.

apple/testing/default_runner/ios_xctestrun_runner.template.sh
erneestoc: Removed here; superseded by the pool in #3026 — with per-test simulator ownership there's no session to serialize.

keith (approved):

keith:
lgtm, can you rebase

erneestoc:
Rebased onto main — the conflict was with `test_environment` (#3045), which landed in the same attribute block; `doc/rules-ios.md` regenerated with `bazel run //doc:update`.

CI note, correcting my earlier read of this: `8-dot-0-1-on-darwin-macos-arm64` and `9-dot-x-on-darwin-macos-arm64` failed with Bazel exit status 3 (build fine, tests failed or timed out), while `8-dot-x` and `last-green-bazel-on-darwin-macos-arm64` passed on the same commit. So this is not an analysis or build problem with the change.

I first attributed this to the simulator-concurrency flakiness described in #3026, but that is at most part of the story and I overstated it. #3026 itself fails `9-dot-x` with the same exit status, so it plainly does not make that job green on its own. `9-dot-x` is also currently red on several other open PRs (#3008, #3015, #3028) and green on others.

What the red ones have in common is that they invalidate the cached results of the simulator tests, so those tests actually execute instead of returning as cache hits — which is the condition #3026 addresses. But that is a correlation across PR contents, not a diagnosis: I have not been able to read the Buildkite logs to confirm which test actually fails, so I would not want the concurrency explanation taken as established.

Happy to re-run, or to narrow this down if someone with Buildkite access can share the failing output.

Issue #1953, opened 2023-05-02
Unable to strip frameworks in ipa_post_processor (invalid signature)


andre-alves:
I'm trying to strip frameworks to decrease app size (as suggested by [EmergeTools](https://docs.emergetools.com/docs/strip-binary-symbols)). Here is the `ipa_post_processor`:

```sh
#!/bin/bash

set -eu

WORKDIR="$1"
APPDIR="$WORKDIR/Payload"

for app in $(find "$APPDIR" -type d -maxdepth 1 -mindepth 1); do
  # Strip symbols from frameworks, as long as the framework is NOT signed by Apple
  find "$app/Frameworks" -type f -perm +111 -maxdepth 2 -mindepth 2 -exec bash -c "/usr/bin/codesign -v -R='anchor apple' {} &> /dev/null || (echo {} && /usr/bin/strip -rSTx {})" \;
done
```

The script works fine but the app fails to install afterwards due to an invalid signature:

```
codesign --verify --verbose MyApp.app/Frameworks/SendbirdChatSDK.framework
MyApp.app/Frameworks/SendbirdChatSDK.framework: invalid signature (code or signature have been modified)
In architecture: arm64
```

I would expect it to work given that codesign is executed after ipa_post_processor.

Maybe something has changed since this [issue](https://github.com/bazelbuild/rules_apple/issues/327#issuecomment-590994343)?

Environment:
- rules_apple 2.2.0
- Xcode 14.1

qyang-nj:
We ran into this as well, with the following crash log. This is launching in simulator. (I used to think that simulator doesn't verify code signature.)
```
Exception Type:  EXC_BAD_ACCESS (SIGKILL (Code Signature Invalid))
Exception Subtype: UNKNOWN_0x32 at 0x000000011d4c4000
Exception Codes: 0x0000000000000032, 0x000000011d4c4000
VM Region Info: 0x11d4c4000 is in 0x11d4c4000-0x11d520000;  bytes after start: 0  bytes before end: 376831
      REGION TYPE                    START - END         [ VSIZE] PRT/MAX SHRMOD  REGION DETAIL
      mapped file                 11d498000-11d4c4000    [  176K] r--/rwx SM=COW  ...t_id=2d3d2fa5
--->  mapped file                 11d4c4000-11d520000    [  368K] r-x/rwx SM=COW  ...t_id=30240fa5
[...]

PR #3023, opened 2026-07-19
Verify framework signatures before skipping re-sign (#1953)

vakhidbetrakhmadov:
The `ipa_post_processor` runs before `rules_apple`'s signing step. 
We strip binaries (`strip -rSTx`) in `ipa_post_processor` to reduce app size, which invalidates the
framework signatures applied during each framework's own build target.
`rules_apple`'s signing step then skips those frameworks because they are listed in `signed_frameworks` - a depset
computed at analysis time, unaware that the `ipa_post_processor` modified them. 

See: https://github.com/bazelbuild/rules_apple/issues/1953

### DIFF
--- a/test/starlark_tests/targets_under_test/apple/ipa_post_processor_remove_signature.sh
+++ b/test/starlark_tests/targets_under_test/apple/ipa_post_processor_remove_signature.sh
@@ -1,0 +1,48 @@
+#!/bin/bash
+
+# Copyright 2020 The Bazel Authors. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -eu
+
+WORKDIR="$1"
+case "$APPLE_SDK_PLATFORM" in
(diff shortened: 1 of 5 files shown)


keith:
can you add a test for this case? i think it could be something like a fixture app w/ frameworks + postprocessor just does `codesign --remove` and then you can write an assertion like https://github.com/bazelbuild/rules_apple/blob/321e9e30759015ab02f40752c17d14064fe534ed/test/starlark_tests/tvos_application_tests.bzl#L147-L162 that it is signed afterwards

vakhidbetrakhmadov:
added a test

keith:
it looks to me like that test passes even w/o your python change?

vakhidbetrakhmadov:
thanks for catching this, i fixed the test and it's now green with the changes in `tools/codesigningtool/codesigningtool.py` and red without

test/starlark_tests/targets_under_test/apple/ipa_post_processor_remove_signature.sh
```diff
@@ -0,0 +1,48 @@
...
+  if [ "$APPLE_SDK_PLATFORM" != "MacOSX" ]; then
+    FRAMEWORK_DIR="$app/Frameworks"
+  else
+    FRAMEWORK_DIR="$app/Contents/Frameworks"
+  fi
+
+  if [[ -d "$FRAMEWORK_DIR" ]]; then
+    for fmwk in \
+        $(find "$FRAMEWORK_DIR" -type d -maxdepth 1 -mindepth 1); do
+      /usr/bin/codesign --remove-signature "$fmwk"
```
keith, line 46:
```suggestion
      /usr/bin/codesign --verify "$fmwk"
      /usr/bin/codesign --remove-signature "$fmwk"
```

this catches if we accidentally don't sign them and therefore invalidate the test in the future

keith (approved):

PR #3026, opened 2026-07-22
Give each concurrent test action its own simulator via a per-(device, OS) pool

erneestoc:
## Summary

Split out from #3021 per review feedback, redesigned as a simulator pool after further discussion: instead of serializing concurrent tests on one shared simulator, each concurrent test action gets its own.

## Problem

A simulator can only service one test session at a time. Two failure modes follow whenever Bazel actually executes several simulator tests concurrently (CI runs `--local_test_jobs=2`):

1. **Session collisions**: concurrent `xcodebuild test-without-building` / `xctest` sessions against the same reused simulator disrupt each other — 300s hangs, or `Failed to initialize for UI testing: … kAXErrorCannotComplete`.
2. **Creation races**: concurrent tests racing through the create action each fail to find an existing simulator and create same-named duplicates. CI logs from #3021 show two `BAZEL_TEST_iPhone 16_26.0` devices created 3 seconds apart, with every simulator test on that machine hanging from then on.

These are the dominant source of iOS test flakiness on this repo's CI. They stay invisible on main because the affected tests are almost always remote-cache hits; any PR that touches the runner invalidates those caches and forces real concurrent executions (#3021 hit this for 6 consecutive builds across 8 different agents until the interim locks landed on its branch).

## Design

- **Pool slot claim (runner template)**: before invoking the simulator creator, the runner claims an exclusive slot in a machine-wide pool by atomically creating a `shlock(1)` pid lockfile in `$TMPDIR`. Slots are *not* keyed on device type or OS version, so a single invocation running tests across several simulator types still hands every concurrent test its own device.
- **Simulator naming**: slot 0 keeps the historical simulator name, so existing simulators keep being reused; higher slots only exist while tests actually run concurrently and get their own suffixed simulator (`…_1`, `…_2`).
- **Where the locks live**: `${SIMULATOR_POOL_LOCK_DIR:-${TMPDIR:-/tmp}}`. Bazel points `TMPDIR` at the per-user temp dir (`/var/folders/…/T/`), which is shared across sandboxed test actions and writable there — unlike `TEST_TMPDIR`, which is per-action. Verified under `darwin-sandbox` with two concurrent test actions.
- **Remote execution escape hatch**: `TMPDIR` is *not* part of the action's declared environment (`bazel aquery` shows only `PATH`), so under remote execution the worker decides what it is, and a worker handing each action its own `TMPDIR` would silently put every test back on slot 0. `--test_env=SIMULATOR_POOL_LOCK_DIR=…` points the pool at a directory the execution environment actually shares. Set-but-unusable warns and falls back rather than failing the test.
- **Robustness**: a claim is valid only while the claiming test process is alive. A test killed for any reason — including SIGKILL from a Bazel test timeout — never leaves a slot permanently claimed: the next prober validates the recorded pid and `shlock` atomically reclaims dead slots. (First attempt used a kernel flock held on an fd via `lockf(1)`; abandoned because macOS `lockf`'s fd mode misbehaves under try-lock probing — after one failed probe, every subsequent probe in that process fails even on free lock files.)
- **Session-wedge recovery (runner)**: a CI attempt log showed the dominant hang: the runner finds its pooled simulator booted, the readiness probe passes, and `xcodebuild test-without-building` sits silent until the Bazel timeout — the device is carrying a dead test session from a previously terminated test, and every new session against it hangs. Two complementary defenses: (1) the runner traps TERM/INT after claiming its simulator and shuts the device down (Bazel delivers SIGTERM with a grace period before SIGKILL), so the next attempt cold-boots clean; (2) for deaths the trap cannot catch — SIGKILL, grace overruns — the runner keeps a session marker holding its pid (same pid-liveness pattern as the pool locks), removed on any controlled exit, and the creator reboots any booted device whose marker names a dead pid before reusing it. Normal completions leave the device booted for warm reuse. This also self-heals agents whose devices are already wedged: the first attempt still times out, but the device is recycled and the retry passes.
- **Bounded boot (creator)**: `simctl bootstatus -b` has no deadline of its own, so a device wedged mid-boot consumed the whole test timeout and was SIGKILLed — leaving it wedged for the next attempt. The boot is now bounded against the test's own `TEST_TIMEOUT` (minus a 30s cleanup reserve, 30s floor — the bound exists to fail cleanly before SIGKILL, not to ration boot time, so slow first-boot data migrations still get nearly the whole budget), and a timed-out boot is deliberately left in progress: the boot is owned by CoreSimulatorService and continues in the background, so a flaky-test retry re-enters `bootstatus` and resumes where the previous attempt left off — a slow first-boot data migration completes across attempts instead of restarting from zero each time.
- **Reused-simulator health check (creator)**: a simulator can report state `booted` while being unusable — if the test that started its initial boot was killed partway through, the device stays `booted` but never becomes able to run tests, and every later test reusing it hangs. Reused `booted` devices are now probed for readiness (SpringBoard in `simctl spawn <udid> launchctl list`, bounded at 30s) and rebooted if not responding.
- **Unaffected**: `ios_test_runner` (doesn't set `SIMULATOR_POOL_SLOT` → slot 0 → today's behavior), `reuse_simulator = False`, device runs, and machines without `shlock(1)` (degrades to today's single-simulator behavior).

## Review feedback addressed

- Pool is no longer device/OS-version specific — it's machine-wide.
- Reads `$TMPDIR` instead of shelling out to `getconf DARWIN_USER_TEMP_DIR`.
- The machine-wide boot gate is **removed** entirely rather than having its timeout shortened: the creation race it partly covered is already fixed by per-test slots, and `--local_test_jobs` already bounds overlapping boots.
- `SIMULATOR_POOL_SLOT` now parses as a hard error instead of silently falling back to slot 0.
- Added `SIMULATOR_POOL_LOCK_DIR` so the lock directory is configurable for execution environments that do not share `$TMPDIR`.

## Test plan

- [x] Concurrent-claim simulation: concurrent runners get distinct slots; a freed slot is reclaimed by the next test; a SIGKILLed holder's slot is atomically reclaimed.
- [x] Locking verified inside Bazel's `darwin-sandbox` (shared `$TMPDIR` is writable and lockable there, and identical across concurrent test actions, under both `darwin-sandbox` and `--spawn_strategy=local`).
- [x] `SIMULATOR_POOL_LOCK_DIR`: three concurrent claimers with *differing* `$TMPDIR`s still get distinct slots when pointed at a shared dir; unwritable dir, missing dir and unset `$TMPDIR` all fall back to slot 0 immediately instead of hanging; propagation through `--test_env` confirmed.
- [x] `//test:ios_xctestrun_runner_ui_test` behaves identically to main; single-test runs claim slot 0 and reuse the historically-named simulator.


apple/testing/default_runner/ios_xctestrun_runner.template.sh
```diff
@@ -494,7 +494,31 @@ if [[ "$build_for_device" == false ]]; then
...
+  # while this process is alive, so a test killed for any reason — including
+  # SIGKILL from a test timeout — never leaves a slot permanently claimed;
+  # the next prober validates the recorded pid and atomically reclaims dead
+  # slots. The lock files live in the per-user temp dir, which is shared
+  # across Bazel's sandboxed test actions. Skipped when shlock(1) is
+  # unavailable, preserving today's single-simulator behavior.
+  simulator_pool_slot=0
+  if [[ "${reuse_simulator:-}" == 1 && -x /usr/bin/shlock ]]; then
+    pool_lock_root="$(getconf DARWIN_USER_TEMP_DIR 2>/dev/null || echo /tmp)"
+    pool_lock_prefix="${pool_lock_root%/}/rules_apple_simulator_pool_$(printf '%s' "%(device_type)s_%(os_version)s" | /usr/bin/tr -C '[:alnum:]._-' '_')"
```
keith: I don't think this should be device or OS version specific, since the issue is wider than that even if 1 invocation launches multiple simulator types in tests

apple/testing/default_runner/ios_xctestrun_runner.template.sh
```diff
@@ -494,7 +494,31 @@ if [[ "$build_for_device" == false ]]; then
...
+  # simulator. A claim is an atomic shlock(1) pid lockfile: it is valid only
+  # while this process is alive, so a test killed for any reason — including
+  # SIGKILL from a test timeout — never leaves a slot permanently claimed;
+  # the next prober validates the recorded pid and atomically reclaims dead
+  # slots. The lock files live in the per-user temp dir, which is shared
+  # across Bazel's sandboxed test actions. Skipped when shlock(1) is
+  # unavailable, preserving today's single-simulator behavior.
+  simulator_pool_slot=0
+  if [[ "${reuse_simulator:-}" == 1 && -x /usr/bin/shlock ]]; then
+    pool_lock_root="$(getconf DARWIN_USER_TEMP_DIR 2>/dev/null || echo /tmp)"
```
keith: we should read the tempdir env var that bazel sets instead of computing this ourselves

apple/testing/default_runner/simulator_creator.py
```diff
@@ -22,13 +24,62 @@
...
+from typing import Iterator, List, Optional
 
 
 def _simctl(extra_args: List[str]) -> str:
     return subprocess.check_output(["xcrun", "simctl", *extra_args], text=True)
 
 
+def _shared_lock_dir() -> str:
+    # The per-user temp dir is shared across Bazel's sandboxed test actions
+    # (each action gets a private $TMPDIR), so locks placed there coordinate
```
keith: I don't think this is true, but might be true on linux only

apple/testing/default_runner/simulator_creator.py
```diff
@@ -22,13 +24,62 @@
...
+    # concurrent test actions on the same machine.
+    try:
+        path = subprocess.check_output(
+            ["/usr/bin/getconf", "DARWIN_USER_TEMP_DIR"], text=True
```
keith: same as above

apple/testing/default_runner/simulator_creator.py
```diff
@@ -22,13 +24,62 @@
...
+            fcntl.flock(lock_file, fcntl.LOCK_EX | fcntl.LOCK_NB)
+        except BlockingIOError:
+            print(
+                "note: waiting for another test to finish booting a simulator",
+                file=sys.stderr,
+            )
+            # Bound the wait: a hung boot elsewhere must not consume this
+            # test's whole timeout. Booting ungated is a better outcome than
+            # timing out in the queue.
+            deadline = time.time() + 120
```
keith: definitely can't do this long of a timeout, likely many tests default to 60s timeout. im not convinced doing multiple boots at once is an issue. back when I was managing a simulator pool on intel mac minis even I was booting 4 at once

apple/testing/default_runner/simulator_creator.py
```diff
@@ -234,9 +326,13 @@ def _main() -> None:
         os.getenv("SIMULATOR_REUSE_SIMULATOR") is not None
     )
 
+    pool_slot = os.getenv("SIMULATOR_POOL_SLOT", "0")
+    if not pool_slot.isdigit():
+        pool_slot = "0"
```
keith:
```suggestion
    pool_slot = int(os.getenv("SIMULATOR_POOL_SLOT", "0"))
```

should be a hard error

apple/testing/default_runner/ios_xctestrun_runner.template.sh
erneestoc: Done — the pool is now machine-wide, not keyed on (device type, OS version). Slot N is claimed globally, so an invocation launching several simulator types still hands every concurrent test its own device. Slot 0 keeps the historical name per device type, so existing simulators are still reused.

apple/testing/default_runner/ios_xctestrun_runner.template.sh
erneestoc: Done — now uses `$TMPDIR`. You were right: I checked what a test action actually sees on macOS and Bazel sets `TMPDIR` to the shared per-user temp dir (`/var/folders/.../T/`, identical to `getconf DARWIN_USER_TEMP_DIR`), while `TEST_TMPDIR` is the per-action one. Two concurrent sandboxed test actions see the same `TMPDIR` and it is writable there, so the `getconf` call is gone.

apple/testing/default_runner/simulator_creator.py
erneestoc: You were right, and this whole helper is gone now — see the boot-gate thread below. For the record: on macOS `TMPDIR` in a test action is the shared per-user temp dir, not a private one; the private per-action dir is `TEST_TMPDIR`. The comment was wrong.

apple/testing/default_runner/simulator_creator.py
erneestoc: Same — helper removed.

apple/testing/default_runner/simulator_creator.py
erneestoc: Removed the boot gate entirely rather than shortening the timeout. The creation race it was partly covering (concurrent tests each missing the existing simulator and creating same-named duplicates) is already fixed by the pool giving each concurrent test its own slot and name, and `--local_test_jobs` already bounds how many boots can overlap — so with your experience of booting 4 at once there is no reason to serialize. This also drops the `fcntl`/`contextlib` imports and the shared-lock-dir helper above.

apple/testing/default_runner/simulator_creator.py
erneestoc: Applied.

apple/testing/default_runner/ios_xctestrun_runner.template.sh
erneestoc: Following up with what I measured, since the lock dir choice has a subtle scope implication.

Within a single Bazel invocation `$TMPDIR` is rock solid — I probed two concurrent `sh_test` actions with `--local_test_jobs=2` and both saw the identical `/var/folders/…/T/`, under `darwin-sandbox` and under `--spawn_strategy=local`, and it is writable from inside the sandbox. So for the case this PR targets (concurrent test actions of one invocation) `$TMPDIR` is exactly right.

The one gap: `$TMPDIR` is inherited from the client environment, so it is not invariant the way `getconf DARWIN_USER_TEMP_DIR` is. `TMPDIR=/tmp/foo bazel test …` propagates straight through to the test action (notably `--test_env=TMPDIR=…` does *not* override it). Two concurrent Bazel invocations on one machine that disagree on `TMPDIR` — say a CI agent that sets a per-job scratch dir, or one invocation from a context where it is unset — would then keep separate pools while sharing the same per-user CoreSimulator device set, and each would independently claim slot 0.

That degrades to exactly today's shared-simulator behavior rather than to anything worse, so I have kept `$TMPDIR` as you asked and documented the caveat in the comment. Flagging it in case you would rather have the lock scope keyed to the user (which is the actual scope of the simulator set) than to the environment — happy to switch it back if so.

erneestoc: Done — uses `$TMPDIR`. Also added an optional `SIMULATOR_POOL_LOCK_DIR` override: `TMPDIR` is not in the action's declared env, so under remote execution a worker giving each action its own tmpdir would silently disable the pool; the override lets such setups point the locks at a genuinely shared dir. While testing that path I also fixed a hang — `shlock` reports "can't create lock file" the same as "slot taken", so an unusable lock dir spun the claim loop until test timeout; it's now bounded with a slot-0 fallback.

apple/testing/default_runner/__pycache__/simulator_creator.cpython-314.pyc
aaronsky, line 1: delete this please

apple/testing/default_runner/ios_xctestrun_runner.template.sh
```diff
@@ -498,9 +520,108 @@ if [[ "$build_for_device" == false ]]; then
...
+    done
+    if (( simulator_pool_slot == 64 )); then
+      echo "note: could not claim a simulator pool slot in '$pool_lock_dir'; falling back to the shared simulator"
+      simulator_pool_slot=0
+    else
+      simulator_pool_slot_claimed=true
+    fi
+  fi
+
+  simulator_id="$(SIMULATOR_DEVICE_TYPE="%(device_type)s" SIMULATOR_OS_VERSION="%(os_version)s" SIMULATOR_POOL_SLOT="$simulator_pool_slot" SIMULATOR_REUSE_SIMULATOR="${reuse_simulator:-}" SIMULATOR_SDK_VERSION="%(sdk_version)s" XCTESTRUN_RUNNER_PID="${BASHPID:-$$}" "%(create_simulator_action_binary)s")"
```
aaronsky, line 571: seems like `SIMULATOR_POOL_SLOT` should be documented in `ios_xctestrun_runner.bzl`

apple/testing/default_runner/simulator_creator.py
```diff
@@ -151,13 +291,36 @@ def _create_and_boot_simulator(
...
         name = existing_device["name"]
-        # If the device is already booted assume that it was created with this
-        # script and bootstatus has already waited for it to be in a good state
-        # once
+        # If the device is already booted it was created with this script and
+        # bootstatus has already waited for it to be in a good state once, so
+        # only its readiness needs re-checking below.
         state = existing_device["state"].lower()
         print(f"Existing simulator '{name}' ({simulator_id}) state is: {state}", file=sys.stderr)
+        marker_state = _session_marker_state(simulator_id)
```
aaronsky, line 299: idle thought, would `simctl clone` be more appropriate here

apple/testing/default_runner/__pycache__/simulator_creator.cpython-314.pyc
erneestoc: Done — removed.

apple/testing/default_runner/ios_xctestrun_runner.template.sh
erneestoc: Done — documented alongside the other `SIMULATOR_*` variables in the `create_simulator_action` contract, with `doc/rules-ios.md` regenerated.

apple/testing/default_runner/simulator_creator.py
erneestoc: Looked into it — `clone` does not fit this particular moment, for a mechanical reason: `simctl clone` requires the source device to be shut down (`Unable to clone device in current state: Booted`, verified on Xcode 26.2), and slot 0 is booted precisely when a concurrent test is creating slot N. Clone also copies the source data container and the clone still pays its own boot, which is the expensive part. Where it could shine is a different optimization: pre-seeding a pool from a shut-down, fully-migrated template device so first boots skip data migration — happy to explore that as a follow-up if there is interest.

aaronsky (approved):

Issue #338, opened 2019-02-24
State of code coverage


keith:
I've been looking through bazel/rules_apple/apple_support/rules_swift and it appears that there's a decent amount of integration for supporting code coverage that could work for iOS. But it seems like some critical pieces are not hooked up, like [empty filegroups](https://github.com/bazelbuild/apple_support/blob/77c9d5371949352c00bc74d4755c93e00199172b/tools/BUILD#L3-L9) [stub binaries](https://github.com/bazelbuild/bazel/blob/34bf56f25eb1bed7fbd5c4cf7bae15bf461cf904/tools/objc/gcov_stub) and variables that [wouldn't be passed through to iOS tests](https://github.com/bazelbuild/bazel/blob/34bf56f25eb1bed7fbd5c4cf7bae15bf461cf904/tools/test/collect_coverage.sh#L79).

What's the plan with supporting code coverage for iOS? Is this something that's already supported inside of Google but not in the public for some reason?

thomasvl:
@allevato likely knows the most about this.

allevato:
Indeed, the coverage support in the rules is currently for 🤫 internal use 🤫 at this time.

I'm not too up-to-date on the current state of Bazel coverage. Bazel parses `lcov` tracefiles to generate its coverage reports, and I recently added support to `llvm-cov` to export `lcov` formatted files (https://github.com/llvm/llvm-project/commit/b2091c930b765a1f1b75c9159f49329af59604f9) so that we could use that within Google, but externally you'd need to wait for an Xcode release that includes that patch or build your own copy for it to be useful.

keith:
Is that the only thing missing from making this work externally? (It seems like the rules and test runner might also need some changes?)

allevato:
It looks like the version of `llvm-cov` distributed with Xcode 11 beta includes the `--format=lcov` flag that I added, so once that's in place, it would be up to the coverage collection scripts in Bazel to invoke it and transform the raw coverage data into the format it uses.

tinder-maxwellelliott:
Any update on this ticket? Looking to start trying to surface some coverage on my Bazel project. I would believe this would be supported if the last ticket was from the xcode 11 beta

keith:
We have this working on our project internally, but we have a custom test runner to do so. But I can outline the gist of how it works here:

1. Set `--experimental_use_llvm_covmap` globally
[...]

PR #3037, opened 2026-07-29
Collect normal-mode profiles from XCUITest target apps

karim-alweheshy:
## 🤖 AI Disclaimer

I used Codex to help investigate, implement, and test this change. I reviewed
the resulting behavior against LLVM's profile format and exercised the
integration test locally.

## Problem

The `ios_xctestrun_runner` can collect coverage from the XCTest runner, but it
cannot currently put `LLVM_PROFILE_FILE` in the XCUITest target application's
environment or retrieve that application's raw profiles.

This matters for instrumentation PGO. The existing
`ClangProfileDataDirectoryPath` path selects compiler-rt continuous mode.
Continuous mode preserves counters, but its writer skips LLVM value profile
data, including indirect-call targets. A counter-complete profile can therefore
still be unusable for indirect-call promotion and profile-guided
devirtualization.

## Change

Passing:

```text
--test_env=LLVM_PROFILE_FILE_FOR_TARGET_APP=%t/<name>-%p.profraw
```

now makes the runner:

- set the target app's `LLVM_PROFILE_FILE` through
  `UITargetAppEnvironmentVariables`;
- remove `ClangProfileDataDirectoryPath` and its paired coverage metadata so
  Xcode does not select continuous mode;
- collect only raw files created during the current test run from the target
  app's simulator data container;
- fail when no fresh target-app profile was written;
- preserve those raws under `target-app-profraw` in undeclared outputs; and
- merge them into `Coverage.profdata`, failing if the profile is invalid.

The opt-in is restricted to simulator XCUITests with a host app. The pattern
must be a file directly under `%t`, contain `%p`, and use no other substitutions
or glob characters.

Compiler-rt registers an automatic writer for normal process exit, so calling
`__llvm_profile_write_file()` is not mandatory when that writer produces a
fresh profile. An explicit target-app call remains useful to guarantee that the
profile is committed at an exact journey boundary, before XCTest terminates the
app or before a later crash or forced termination. The rule documentation now
contains Objective-C and Swift examples.

This deliberately does not swizzle `XCUIApplication.terminate()`. A swizzle
runs in the XCTest runner and cannot directly call the profiling runtime in
another process. The runner instead enforces the observable result: normal exit
or an explicit target-app flush must produce a fresh, mergeable profile.

Related to #338.

## Verification

Tested locally with Bazel 9.2.0 and Xcode 26.5:

- `bazel build //apple/testing/default_runner:ios_default_runner`
- focused `test_ios_ui_test_collects_target_app_profiles` integration case
  through `//test:ios_xctestrun_runner_ui_test`
- `bazel test //doc:check_ios`
- `shellcheck test/ios_xctestrun_runner_ui_test.sh`
- buildifier and buildifier-lint 8.5.1 through pre-commit
- a reduced Swift 6.3.2 compile/run/merge check for the documented
  `@_silgen_name` declaration

The committed integration case launches an IR-instrumented target app twice,
explicitly flushes in the target process, collects two fresh raw profiles,
merges them, and verifies with `llvm-profdata show --ic-targets` that the merged
profile contains an indirect-call target record.

A local control app with the same instrumentation but no explicit flush also
produced two fresh profiles when XCUITest replaced and ended its processes on
the Xcode 26.5 simulator. That confirms an explicit call is not universally
required when compiler-rt's normal-exit writer runs; the explicit integration
call remains the deterministic journey-boundary example.

### DIFF
--- a/apple/testing/default_runner/ios_xctestrun_runner.bzl
+++ b/apple/testing/default_runner/ios_xctestrun_runner.bzl
@@ -325,6 +325,60 @@ If you would like this test runner to generate xcresult bundles for your tests,
 pass `--test_env=CREATE_XCRESULT_BUNDLE=1`. It is preferable to use the
 `create_xcresult_bundle` on the test runner itself instead of this parameter.
 
+For an instrumented UI test host app, pass
+`--test_env=LLVM_PROFILE_FILE_FOR_TARGET_APP=%t/<name>-%p.profraw` to collect
+normal-mode profiles from the target app. `%t` resolves to the target app's
+temporary directory and `%p` gives each process a unique file. The app must
+exit normally or explicitly flush its profile. The profile runtime registers
+an automatic writer for normal process exit, so an explicit call is not
+mandatory when that writer runs and produces a fresh profile. To guarantee that
+the profile is committed at the end of a journey, before XCTest terminates the
+app or before a later crash or forced termination, call
+`__llvm_profile_write_file()` from the instrumented target app.
+
+Objective-C can use Apple Clang's instrumentation define:
+
+```objective-c
+#if __LLVM_INSTR_PROFILE_GENERATE
+extern int __llvm_profile_write_file(void);
+#endif
+
+BOOL WriteLLVMProfile(void) {
+#if __LLVM_INSTR_PROFILE_GENERATE
+  return __llvm_profile_write_file() == 0;
+#else
+  return NO;
+#endif
+}
+```
+
+Swift can import the Objective-C helper above, or declare the runtime entry
+point directly. The latter uses the underscored `@_silgen_name` attribute;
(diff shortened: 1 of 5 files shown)


karim-alweheshy:
CI follow-up: a no-source-change retry is fully green across Bazel 8.0.1, Bazel 8.x, Bazel 9.x, last-green Bazel, buildifier, and documentation tests. The prior failure had the same unrelated simulator and coverage targets time out in both Bazel 8 lanes; this PR’s new profiling integration target did not fail, and an earlier same-patch lane also passed. I therefore treated the failure as shared simulator infrastructure noise and did not add a source workaround.

karim-alweheshy:
@aaronsky @keith, could one of you review this XCUITest runner change when convenient? It preserves the existing test-process profile collection and adds normal-mode collection from the launched target app, with focused integration coverage. The no-source-change CI retry is green across every supported Bazel lane.

PR #3038, opened 2026-07-31
Test no transition for build_test rules

keith:
This makes it so you can use target_comaptible_with to scope these to
macOS to avoid building them on Linux

Reverts 7f75280edbb786b964bf31bfede4de2c5bb0c357


keith:
https://github.com/bazelbuild/rules_apple/pull/3040

PR #3040, opened 2026-08-05
Remove top level transitions from build_test rules

keith:
This makes it so you can use target_comaptible_with to scope these to
macOS to avoid building them on Linux

This is a partial revert of 7f75280edbb786b964bf31bfede4de2c5bb0c357 but
works the same for 9.x


keith:
this works but there isn't much of a win if we can't also do the same on ios_application and such

PR #3043, opened 2026-08-11
Support cc_toolchains with rules_cc rule-based configs in triplet

mauriciogg:
`rules_cc` rule-based toolchain configuration (cc/toolchains/...) generates a `CcToolchainConfigInfo` with an empty target_system_name. Importing a framework/xcframework under such a toolchain crashes` _get_apple_clang_triplet`

When `target_gnu_system_name `carries no triple, recover the Apple platform from the toolchain's generated config label instead (rules_cc names it _<toolchain name>_config), matching the toolchain name against the conventional Apple cpu names, and fail with a descriptive message when that is not possible.

### DIFF
--- a/apple/internal/cc_toolchain_info_support.bzl
+++ b/apple/internal/cc_toolchain_info_support.bzl
@@ -14,6 +14,59 @@
 
 """Support methods for handling CcToolchainInfo providers."""
 
+# Apple cpu name -> (architecture, normalized os, environment). Used to
+# recover the target platform for cc_toolchains built with rules_cc's
+# rule-based toolchain configuration, whose generated CcToolchainConfigInfo
+# hardcodes an empty target_system_name (see rules_cc
+# cc/toolchains/impl/toolchain_config.bzl).
+_APPLE_CPU_TRIPLETS = {
+    "darwin_arm64": ("arm64", "macos", "device"),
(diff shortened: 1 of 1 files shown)


apple/internal/cc_toolchain_info_support.bzl
```diff
@@ -14,6 +14,59 @@
 
 """Support methods for handling CcToolchainInfo providers."""
 
+# Apple cpu name -> (architecture, normalized os, environment). Used to
+# recover the target platform for cc_toolchains built with rules_cc's
+# rule-based toolchain configuration, whose generated CcToolchainConfigInfo
+# hardcodes an empty target_system_name (see rules_cc
```
keith, line 20: what toolchain is this true for? i added support for this in https://github.com/bazelbuild/rules_cc/commit/bf92fe4236f1600636ab42888f14905a5517551d and made sure to always set this in apple_support https://github.com/bazelbuild/apple_support/blob/8a5de128485e553bd35b8cd881b43480708f43ec/toolchain/cc_toolchain.bzl#L190

i believe without this other things break?

mauriciogg: you're right- we are on an older version of `rules_cc`. Closing this PR

PR #3044, opened 2026-08-11
Constrain macOS-only tool rules to macOS execution platforms

mauriciogg:
Constrain macOS-only tool rules to macOS execution platforms

Several standalone tool rules run macOS-only tools but carry no execution-platform constraint which causes failures where there are multiple execution platforms available. E.g. RBE with a linux toolchains (see [rules_applecross](https://github.com/apple-cross-toolchain/rules_applecross/))

Also switch `apple_mac_tools_toolchain`  `alticonstool to cfg = "target"` like the toolchain's other tool attributes: the toolchain target is already resolved for a mac execution platform, so "exec" would re-resolve against the first registered execution platform.

### DIFF
--- a/apple/BUILD
+++ b/apple/BUILD
@@ -285,6 +285,9 @@ bzl_library(
 
 cc_toolchain_forwarder(
     name = "default_cc_toolchain_forwarder",
+    exec_compatible_with = [
+        "@platforms//os:macos",
+    ],
 )
 
 # Consumed by bazel tests.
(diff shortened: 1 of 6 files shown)


keith:
shouldn't most of these platform selections be done correctly by the toolchain registered in the exec groups? the toolchain itself should be the one forcing actions to run on the right platform.

rules that only fetch the "xplat" toolchain but then are being modified here seem like we must be doing something wrong since those should also be up to the toolchain to decide the platform

keith:
fixed by https://github.com/bazelbuild/rules_apple/pull/3066, please reopen any other cases you find after pulling that in!

PR #3045, opened 2026-08-12
Add test_environment to ios_xctestrun_runner

john-flanagan:
Add the `test_environment` attribute that's on `ios_test_runner` to `ios_xctestrun_runner`


aaronsky:
~~This is redundant against `env` and `--test_env`, I’m pretty sure.~~

oh I see this is actually for the runner. hmm, could you explain your use-case for this?

john-flanagan:
@aaronsky It's porting this PR over to the xctestrunner version: https://github.com/bazelbuild/rules_apple/pull/682

We'd like to use it because some of our snapshot tests are dependent on the device timezone. We could define the timezone on every individual unit test target, but I'd like to just do it on the test runner.

keith (approved):

PR #3050, opened 2026-08-21
Contain xcodebuild's derived data in the test's temp dir

erneestoc:
Every `xcodebuild test-without-building` invocation creates a `~/Library/Developer/Xcode/DerivedData/temporary-<hash>/` directory that nothing ever deletes. When the runner does not pass `-resultBundlePath` — i.e. any xcodebuild-path run without `create_xcresult_bundle` — that directory receives a **full `.xcresult` of the run**; and passing `-resultBundlePath` alone does not fix it, a bookkeeping directory still leaks per run.

One development machine had accumulated **267 of these directories totaling 26GB**. Every CI agent running simulator tests through the xcodebuild path leaks one per test run, silently.

Fix: pass `-derivedDataPath` pointed into the runner's `test_tmp_dir`, which the existing EXIT trap already deletes. An explicit `-resultBundlePath` (from `create_xcresult_bundle` or `xcodebuild_args`) still wins for the result bundle itself.

## Measurements

New DerivedData `temporary-*` directories created per invocation (Xcode 26.2, iOS 26.2 simulator, hosted unit test):

| xcodebuild flags | new dirs per run | contents |
|---|---|---|
| (none — today's runner without `create_xcresult_bundle`) | **1** | 152K, including a complete orphan `.xcresult` |
| `-resultBundlePath` only | **1** | 24K bookkeeping |
| `-derivedDataPath` (this PR) | **0** | — |
| both | **0** | — |

End-to-end through the runner after this change: default mode, dir count unchanged across runs (was +1 per run); with `create_xcresult_bundle = True`, the bundle still lands in `TEST_UNDECLARED_OUTPUTS_DIR` (verified present in `outputs.zip`) and DerivedData stays untouched.

## Steps to reproduce the leak

```sh
ls -d ~/Library/Developer/Xcode/DerivedData/temporary-* | wc -l
# run any hosted test through ios_xctestrun_runner without create_xcresult_bundle,
# e.g. a unit test with a test_host (forces the xcodebuild path):
bazel test //your:HostedTest --runner=@rules_apple//apple/testing/default_runner:ios_xctestrun_ordered_runner --nocache_test_results
ls -d ~/Library/Developer/Xcode/DerivedData/temporary-* | wc -l   # +1, forever
```

## Test plan

- [x] Manual `xcodebuild test-without-building` flag matrix above (each variant run against the same xctestrun).
- [x] Runner end-to-end: default mode creates no DerivedData directories; `CREATE_XCRESULT_BUNDLE=1` still produces `tests.xcresult` in undeclared outputs with DerivedData untouched.
- [x] Rendered template passes `bash -n`.

### DIFF
--- a/apple/testing/default_runner/ios_xctestrun_runner.template.sh
+++ b/apple/testing/default_runner/ios_xctestrun_runner.template.sh
@@ -615,6 +615,16 @@ if [[ "$should_use_xcodebuild" == true ]]; then
     -xctestrun "$xctestrun_file" \
   )
 
+  # Contain everything xcodebuild writes to disk. Without -derivedDataPath,
+  # every invocation creates a ~/Library/Developer/Xcode/DerivedData/
+  # temporary-<hash>/ directory that nothing ever deletes: when no result
+  # bundle path is passed it receives a full .xcresult of the run, and even
+  # with one it still leaks a bookkeeping directory per run. Pointing derived
+  # data into the test's temp dir routes all of it somewhere the runner
+  # already deletes on exit. An explicit -resultBundlePath (below, or via
+  # xcodebuild_args) still wins for the result bundle itself.
+  args+=(-derivedDataPath "$test_tmp_dir/derived_data")
+
   if [[ -n "$destination_timeout" ]]; then
     args+=(-destination-timeout "$destination_timeout")
   fi


keith (approved):

erneestoc:
The CI failures here are the pre-existing wedged-simulator flake, not this change:

- The failing set (`HelloWorldSwiftUITests`, `ButtonsUITests`, `ios_coverage_ui_test_coverage_new_runner`) is the same session-sensitive group that has been failing on other PRs whose runner-test caches get invalidated (#3026 has the full diagnosis: a previously-terminated test leaves the reused, still-booted simulator with a dead device-side session, and every later `xcodebuild test-without-building` against it hangs or fails to initialize).
- The attempt log shows the signature exactly: reused booted `BAZEL_TEST_iPhone 16_26.0`, a well-formed invocation with `-derivedDataPath`, then silence until the timeout SIGTERM.
- 1163 tests passed on the same run — including the hosted unit tests, which exercise `-derivedDataPath` on every invocation.
- All three failing targets pass locally on this branch (9.8s / 11.6s / 5.3s).

#3026 adds the recovery machinery for this (its own CI is now fully green, including the previously-failing 9.x job, because its first run self-heals a wedged agent). Suggest retrying this build after #3026 lands — no change needed here.

adincebic (approved):
Good.

PR #3063, opened 2026-09-14
Add back default iOS test device_type

keith:
This was removed in d6fd5b55bc07c70226716a7661672a9d2877dc22 which
breaks users who don't pass an explicit simulator type on the command
line.


keith:
we don't remember why this was removed, but it does break people who don't pass that command line flag and use the default runners, which we don't really want to do

aaronsky (approved):

PR #3065, opened 2026-09-14
Require macOS execution for SDK discovery and iOS runners

### LINKED ISSUE #3064: Missing macOS execution constraints in environment_plist and iOS runners break mixed Linux/macOS execution
## Problem

When Linux is the first registered execution platform and macOS is also available, some Xcode-dependent helper rules select Linux despite the surrounding iOS compilation and bundling rules correctly selecting macOS.

Two concrete failures were reproduced:

1. `environment_plist` runs an `xcrun`-dependent action on Linux.
2. `ios_xctestrun_runner` configures `simulator_creator` and related exec dependencies for Linux, then the macOS test runner tries to execute the Linux Python interpreter.

The desired graph deliberately permits Linux execution: a hermetic LLVM toolchain can build a **macOS** executable such as `protoc` on Linux, and the resulting tool can be used on macOS during the iOS build. Restricting every macOS target, or the entire iOS invocation, to macOS execution would disable this valid cross-compilation use case.

## Environment and reproduction

- Bazel 9.2.0
- rules_apple 5.0.0; rules_swift 4.0.1
- apple_support 2.8.2 with `--repo_env=APPLE_SUPPORT_RULES_BASED_TOOLCHAIN=1`
- rules_python 2.3.3, Python 3.14
- Hermetic LLVM module `llvm` 0.8.19
- Apple Silicon macOS host; Xcode 26.6.0 (17F113)
- Linux x86_64 and macOS arm64 execution platforms, in that order

Repository: [DoDoENT/bazel-playground](https://github.com/DoDoENT/bazel-playground/tree/7d97b198fffc61b9d32cee76a1bc0acc0faa2d28).

```sh
bazel test //:ios --config=ios_simulator \
  --extra_execution_platforms=//platforms/linux:haswell_x86_64,//platforms/macos:arm64
```

The explicit platform-list flag overrides the repository's existing iOS workaround. This reproduction uses its configured remote execution backend for Linux actions and a local Mac for simulator tests.

**Prerequisite:** apple_support's unconstrained `force_exec` helper causes an earlier analysis failure. The minimal fix in https://github.com/bazelbuild/apple_support/issues/636 was applied through a temporary module override before reproducing the failures below. Without it, the earlier failure can mask these rules_apple issues.

## Failure 1: environment_plist selects Linux for an Xcode action

After fixing the earlier apple_support failure, the build reached:

```text
Action external/rules_apple+/apple/internal/environment_plist_ios.plist failed
bazel-out/haswell_x86_64-opt-exec/bin/external/rules_apple+/tools/environment_plist/environment_plist --platform iphonesimulator26.5 --output .../environment_plist_ios.plist
# Execution platform: //platforms/linux:haswell_x86_64
environment_plist: SDK not located. This may indicate that the xcode and SDK version pair is not available.
.../environment_plist: line 58: /usr/bin/xcrun: No such file or directory
```

[`environment_plist`](https://github.com/bazelbuild/rules_apple/blob/master/apple/internal/environment_plist.bzl) obtains a Mac tool through `apple_toolchain_utils.get_mac_toolchain(ctx)` and registers the action using `apple_support.run`, but its rule definition does not require macOS execution. The parent bundle rule's macOS constraint does not constrain this dependency's execution platform.

Adding `exec_compatible_with = ["@platforms//os:macos"]` to this rule fixed the failure.

## Failure 2: simulator tools are configured for Linux but used on macOS

After the build completed, all five simulator tests initially failed locally with:

```text
.../rules_apple+/apple/testing/default_runner/simulator_creator
  os.execv(python_program, argv)
OSError: [Errno 8] Exec format error: '.../rules_apple+/apple/testing/default_runner/_simulator_creator.venv/bin/python3'
```

The `ios_xctestrun_runner` rule has `cfg = "exec"` dependencies for simulator creation, cleanup, and other actions, but no macOS execution constraint. Its default execution platform can therefore be Linux, yielding Linux-configured helper runfiles. Requiring macOS for the final test execution group does not retroactively change the runner rule's exec dependencies.

Setting `exec_compatible_with = ["@platforms//os:macos"]` on the downstream `ios_xctestrun_runner` target fixed the runtime failure. The `ios_test_runner` rule has the same unconstrained exec-dependency pattern and should be covered by the upstream fix too; that second runner variant was inspected but not independently exercised in this simulator run.

## Suggested patch

The following adds the requirement at the rule definitions, where all consumers benefit. It leaves the cross-platform Apple execution group and ordinary C++/LLVM rules unrestricted:

```diff
--- a/apple/internal/environment_plist.bzl
+++ b/apple/internal/environment_plist.bzl
@@ -106,4 +106,5 @@
     fragments = ["apple"],
     outputs = {"plist": "%{name}.plist"},
     implementation = _environment_plist_impl,
+    exec_compatible_with = ["@platforms//os:macos"],
 )
--- a/apple/testing/default_runner/ios_test_runner.bzl
+++ b/apple/testing/default_runner/ios_test_runner.bzl
@@ -122,6 +122,7 @@
 
 ios_test_runner = rule(
     _ios_test_runner_impl,
+    exec_compatible_with = ["@platforms//os:macos"],
     attrs = {
         "create_simulator_action": attr.label(
             cfg = "exec",
--- a/apple/testing/default_runner/ios_xctestrun_runner.bzl
+++ b/apple/testing/default_runner/ios_xctestrun_runner.bzl
@@ -148,6 +148,7 @@
 
 ios_xctestrun_runner = rule(
     _ios_xctestrun_runner_impl,
+    exec_compatible_with = ["@platforms//os:macos"],
     attrs = {
         "attachment_lifetime": attr.string(
             default = "keepNever",
```

The current upstream definitions were checked and still lack these constraints. For the runner portion, the exercised workaround was the equivalent target-level `exec_compatible_with`; the rule-definition changes above are the proposed upstream placement.

## Validation and related downstream detail

With the apple_support fix, the environment_plist fix, and macOS constraints on the downstream runner targets, all five iOS simulator tests actually executed and passed while Linux remained the first execution platform.

One additional downstream adjustment was necessary: the repository put `no-remote-exec` on `ios_unit_test`, and the test assembler propagated it to the internal bundle target. Its cross-platform bundle tool was correctly configured for Linux but was then forced to execute locally on macOS. Removing that tag allowed portable bundling on Linux; the existing `exclusive` test tag retained local, sequential simulator execution. This is separate from the missing constraints reported here; the suggested patch does not force cross-platform bundling onto macOS.

A useful regression test would register Linux before macOS, build an iOS test with the default simulator runner, and assert that SDK discovery and simulator helper dependencies select macOS. It should also preserve a macOS tool cross-compiled on Linux to ensure the fix remains limited to helpers that require the Apple execution environment.

DoDoENT:
With Linux registered before macOS, `environment_plist` can run `xcrun` on Linux, and iOS runner exec dependencies can be configured for Linux even though their generated scripts run on macOS. The latter fails when `simulator_creator` tries to execute its Linux Python interpreter locally on a Mac.

Require macOS execution on `environment_plist`, `ios_test_runner`, and `ios_xctestrun_runner`. This places the constraints on the Apple-dependent helpers while keeping cross-platform bundling and Linux-hosted LLVM cross-compilation of ordinary macOS tools available.

Fixes #3064. The earlier apple_support compiler-wrapper analysis failure is tracked in bazelbuild/apple_support#636 and fixed by bazelbuild/apple_support#637.

Validation:
- `buildifier -mode=check -lint=warn` on all three modified files and `git diff --check` passed.
- All five downstream iOS simulator tests executed and passed with Linux first and `--nocache_test_results`, using this exact checkout together with the companion apple_support fix. This exercises `environment_plist` and `ios_xctestrun_runner`; the analogous `ios_test_runner` variant was not independently exercised.
- The downstream `no-remote-exec` tag was temporarily removed from the test macro because it propagated to the Linux-configured bundle actions. The existing `exclusive` tag kept simulator tests local and sequential.


DoDoENT:
Superseded by the maintainer fixes in #3066 and #3067. I verified that all five downstream iOS simulator tests pass with Linux registered first using #3067 (which includes #3066), together with the independent apple_support fix in bazelbuild/apple_support#637. I am preparing a separate regression-tests-only PR from my fork against the #3067 branch.

PR #3069, opened 2026-09-14
Add CI job for basic extra_execution_platforms

keith:
This runs a single test with linux first in extra_execution_platforms to
make sure we correctly setup things to force run on macOS when required.

PR #3070, opened 2026-09-14
Extend Linux-first CI coverage to custom iOS runner actions

DoDoENT:
Extend #3069's Linux-first iOS test job to cover the legacy runner and custom execution dependencies in both runner implementations.

The existing test covers the default xctestrun runner and its toolchain-provided simulator scripts. Two additional tests reuse the same XCTest source and exercise custom creation, pre-action, post-action, and cleanup dependencies. The fixtures declare `target_compatible_with = ["@platforms//os:macos"]`, so changing any of these attributes back to the default `cfg = "exec"` makes the explicitly requested test incompatible when Linux is first. A platform-independent shell script therefore cannot silently hide the regression.

This PR targets `ks/add-ci-job-for-basic-extra_execution_platforms` so it can be merged into #3069. It contains only the extra fixtures and CI target entries; the production fixes remain in #3066 and #3067 (issue #3064).

Validation:
- Bazel 9.2.0: all three fixture tests executed and passed with the #3066/#3067 fixes, Linux first, `sim_arm64`, and the same iPhone 16 setting as upstream CI.
- Independently reverted each of seven custom-action `config.exec(exec_group = APPLE_MAC_EXEC_GROUP)` transitions to `cfg = "exec"`: all seven caused the corresponding test to fail analysis with an incompatible macOS-only dependency. Restored every transition afterward.
- Buildifier, shell syntax validation, and `git diff --check` passed.

Validation used the downstream bazel-playground workspace's Linux C++ toolchain and execution setup, with this test directory added to the #3067 checkout through a module override. In a standalone rules_apple checkout, the original unmodified #3069 `unit_test` also fails analysis because the Linux platform has no matching C++ toolchain for `@bazel_tools//src/tools/launcher:launcher_maker`. That setup issue predates these extra cases; this PR leaves the existing platform setup unchanged.


keith:
i think https://github.com/bazelbuild/rules_apple/pull/3069 is probably enough to cover us, thanks

PR #3078, opened 2026-09-21
Avoid unused test bundle codesigning dossiers

brentleyjones:
Test bundles currently generate codesigning dossiers even when their runners do not consume them. Skip those actions for iOS, macOS, tvOS, and watchOS while retaining dossiers for visionOS and when `disable_legacy_signing` is enabled. Required dossiers remain available in test runfiles.


keith:
what's special about visionOS?

brentleyjones:
@Keith Removed the exception.

aaronsky (approved):

apple/internal/testing/apple_test_bundle_support.bzl
```diff
@@ -128,8 +128,14 @@ def _test_host_bundle_id(test_host):
     test_host_bundle_info = test_host[AppleBundleInfo]
     return test_host_bundle_info.bundle_id
 
-def _apple_test_bundle_impl(*, ctx, product_type):
-    """Implementation for bundling XCTest bundles."""
+def _apple_test_bundle_impl(*, ctx, product_type, requires_dossiers):
```
keith, line 131: `requires_dossiers` is always false now i think?

keith (approved):

